PR #976, opened 2014-09-14
Changed Aluminum to Zinc, AlBrass to Brass, and Alumite to Zarcon

lakotajames:
No idea if this is something you'd want to do now or later, but I wanted it because my modpack had both zinc and aluminum, and they were each only used by one mod, so I figured I'd go ahead and make them the same.


Adaptivity:
@bonii-xx are you still planning to implement this in 1.8.9? Just interesting.

bonii-xx:
Right, I guess this is obsolete now.

PR #1284, opened 2015-01-01
Allow arrows to be made with any OreDictionary'd stickWood

squeek502:


bonii-xx:
Already works by default, no need to change anything :3

PR #1298, opened 2015-01-02
Russian localization for books

Adaptivity:
Translations might not be 100% correct. This is the first version and might contain some mistakes.
Also, fixed some Tinker's Logs being broken in Diary.
Credits:
Me
@fewizz


bonii-xx:
Wow. Good work, thanks! :)

PR #1327, opened 2015-01-11
Create ja_JP.lang

libertyCrown:
It is still incomplete.


bonii-xx:
Just post a comment here when it's complete so i can pull it in

libertyCrown:
I'm sorry, I English is not fluent.
So, there is no confidence to complete than the previous lang file.

bonii-xx:
Ok, maybe somebody will do the rest sometime later. :)

PR #1344, opened 2015-01-14
Fixes world-crash involving some tools in tubestuff block breakers

Shados:
Crash was due to tools with AOE block-breaking capabilties performing an
unsafe cast from EntityPlayer to EntityPlayerMP, which is invalid for
the fake player class used by immibis' tubestuff block breakers (and
potentially other fake players)

### DIFF
--- a/src/main/java/tconstruct/library/tools/HarvestTool.java
+++ b/src/main/java/tconstruct/library/tools/HarvestTool.java
@@ -277,7 +277,7 @@ public abstract class HarvestTool extends ToolCore
                 block.onBlockDestroyedByPlayer(world, x, y, z, meta);
 
             // send update to client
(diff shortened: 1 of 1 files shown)


bonii-xx:
FakePlayer extends EntityPlayerMP. Crashlog?

Shados:
[Here you go](http://dpaste.com/1FRVVK0).

bonii-xx:
Mhm, so he doesn't extend the proper player it seems. Anyway, issue should be fixed as a side effect of some restructuring done with this change: https://github.com/SlimeKnights/TinkersConstruct/commit/824ec2dc55edebed7ee5c6cad2a9458c663bab0d#diff-dcd2eee73a3b53567824cec5e3e242b8R262

PR #1397, opened 2015-02-12
New system for preventing certain modifiers from being combined

DemoXinMC:
ItemModifiers now use a new, blacklist-based, system for checking incompatibilities.
Registering an incompatibility between two modifiers can now be done by calling ItemModifier.addIncompatibility(<key>) on either end where <key> is the ItemModifier.key from the modifier it should be incompatible with.
Added the RegisterModifierEvent, which fires on MinecraftForge.EVENT_BUS whenever a new modifier is registered to the ModifyBuilder.

### DIFF
--- a/src/main/java/tconstruct/library/crafting/ModifyBuilder.java
+++ b/src/main/java/tconstruct/library/crafting/ModifyBuilder.java
@@ -4,6 +4,8 @@ import java.util.*;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import tconstruct.library.modifier.*;
+import net.minecraftforge.common.MinecraftForge;
+import cpw.mods.fml.common.eventhandler.Event;
 
 public class ModifyBuilder
 {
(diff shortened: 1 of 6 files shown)


bonii-xx:
Is there any special reason you need this? I can see and understand the idea behind it, but considering the delicate state of modifiers, why is it necessary?

DemoXinMC:
This is actually designed to address an issue that Drullkus arrived at when attempting to add new modifiers in his TiCon addon mod.  Much like vanilla enchantments, up until a PR I created for Forge itself, checking for conflicts between modifiers is one-way, only the modifier being added can currently check whether there are exclusive modifiers already on the tool, with the ones already present having no say.

The exact situation currently in question is a new modifier called "Void Touch" that is designed to cause blocks to NOT drop items when broken by a Void Touch tool.  This should obviously not be compatible with anything else that affects drops.  However, under the current system, there's no way for Void Touch to prevent Silk Touch, Fortune, or Auto-Smelt being added after Void Touch, it can only stop itself from being placed if those modifiers are already there.

bonii-xx:
Mhm. I think a simpler solution would be to simply add an event when a modifier is applied, that is cancelable. This way you can disallow ANY combination even under special circumstances easily without much change to the code.

bonii-xx:
d1f86adb5840601873539a1b4b2182787fb56a9e

PR #1401, opened 2015-02-15
Updated the description in mcmod.info.

KamranMackey:
The description in the file was a bit, old. Decided to update it. :D


bonii-xx:
Uhm, no? Why would we want such a generic description.

KamranMackey:
Yeah, guess so. Maybe you could update the description then, lol

bonii-xx:
Why would we want to update it? It still describes the mod accurately.

PR #1406, opened 2015-02-17
Fix bug with opening/closing GUI speeding smeltery up...

OndraSter:
... and other possible hidden and related bugs.

markDirty called updateEntity that called heatItems which called markDirty ---

markDirty should definitely not update the entity to begin with.

### DIFF
--- a/src/main/java/tconstruct/smeltery/logic/SmelteryLogic.java
+++ b/src/main/java/tconstruct/smeltery/logic/SmelteryLogic.java
@@ -701,9 +701,6 @@ public class SmelteryLogic extends InventoryLogic implements IActiveLogic, IFaci
     @Override
     public void markDirty ()
     {
-        updateTemperatures();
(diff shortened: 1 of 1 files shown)


bonii-xx:
That's not an issue because it's only clientside and it's needed in several cases. It'd actually introduce many bugs.

BluSunrize:
... markDirty() is called serverside. This is a legitimate bug.

bonii-xx:
Not when opening the GUI.

BluSunrize:
But you don't differentiate ._.
You just call the update in markDirty without checking which side you are on.

OndraSter:
OK, the melting bug is not coming from here - but I am pretty sure the melting is not happening only clientside as you can let the fluid out with a drain/pipe when it melts after quickly opening and closing GUI. Tested with two smelteries next to each other, one being cheesed, one not.

PR #1412, opened 2015-02-20
Update pt_BR.lang

MrKunji:
The original file is too derpy even to exist. (I'm looking at words like 'Enxó', 'Jabá' and such non-commom words)
Thermal Expansion metals names are fixed, as Shiny means Platinum and Mana Infused is Mithril.
Modifiers now have the proper vanilla enchantment name.


MrKunji:
My bad.

PR #1413, opened 2015-02-20
Update pt_BR.lang

MrKunji:
The original file is too derpy even to exist. (I'm looking at words like 'Enxó', 'Jabá' and such non-commom words)
Thermal Expansion metals names are fixed, as Shiny means Platinum and Mana Infused is Mithril.
Modifiers now have the proper vanilla enchantment name. Just fixed things that I let pass early :)


al-myr:
Nice! :)
Thanks for the feedback! :D
[]s

MrKunji:
Y' know, I'm not saying you are doing wrong, I know that as it takes times, things can happen :)

al-myr:
kkk... :D Tranquilo! 
No problem! :D Valeu pelas dicas... ;-)
[]s

PR #1423, opened 2015-02-23
Make tool parts added via addPartCastingMaterial smeltable.

### LINKED ISSUE #1422: Tool parts added by other mods cannot be smelted
Casting recipes added for things like TE's Electrum are added via the addPartCastingMaterial IMC.

This IMC adds the casting recipe for the new part, but does not add a melting recipe to the smeltery so the parts cannot be melted back down into molten metals. The addPartCastingMaterial IMC ought to add the melting recipe too.

codewarrior0:
Allows tool parts added by TE4 (and others) to be smelted back into molten metal.

Fixes #1422

### DIFF
--- a/src/main/java/tconstruct/util/IMCHandler.java
+++ b/src/main/java/tconstruct/util/IMCHandler.java
@@ -12,6 +12,7 @@ import org.apache.logging.log4j.Level;
 import tconstruct.TConstruct;
 import tconstruct.library.TConstructRegistry;
 import tconstruct.library.crafting.CastingRecipe;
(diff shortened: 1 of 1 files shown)


bonii-xx:
That would crash since in your PR the liquids don't have a FluidType. However I recently changed that and can adapt the PR manually.

codewarrior0:
That's right. I had just tried adding a casting from another mod and it crashed because it couldn't get a block to show in the smeltery when melting the parts back down. I ended up adding a Block and Temperature field to addPartCastingMaterial and registering the fluid type with those.

PR #1425, opened 2015-02-23
Some improvements

Adaptivity:
and a bit better mcmod.info


Wulfson:
I would say the author's list was better as it was...

Adaptivity:
This change changes nothing for actual displaying in-game. It's better because it's just one line vs 8...

bonii-xx:
Which makes it harder to read, yes. I don't like it :P

PR #1456, opened 2015-03-09
Update ko_KR.lang

nark31:
* Change
Add ranged weapon information
Add Moltn Metal
Fix renamed modify tag information
Rename Chisel '정' to '끌'

* Yet Translate
Crossbow ability : Reduced reload time with harder materials
- Can not understand the information on the attribute.
- Does that mean a reload of the steel is faster than wood?


bonii-xx:
No, it means a steel crossbow would be faster in comparison to a steel bow, compared to their wooden counterparts. ...it's complicated

nark31:
Still confused. If literal translation, it is possible to be as follows.

'Reduced reload time with harder materials' => '재료가 단단할수록 재장전 속도가 감소됩니다.'

bonii-xx:
Yes, that's a pretty good fit imo. :)
Otherwise it'd take a whole paragraph to explain how it works.

PR #1506, opened 2015-03-27
Fix tab position when effects are being displayed.

micdoodle8:


radfast:
Fixes this issue: https://imgur.com/hH6DhqJ
(play without NEI installed, or with NEI hidden, to see the issue)

We've confirmed the issue occurs in:
* Tinker's Construct alone
* Galacticraft alone
* Tinker's Construct + Galacticraft together

PR #1528, opened 2015-04-09
No more furious item swapping when adding modifiers

kokolihapihvi:
Made it so you can just click on the result slot on tool station/forge to apply modifiers or build a tool.
I personally hate how you have to move the tool back to the slot on the right every time you apply a modifier.
Other fixes too, apparently.

### DIFF
--- a/src/main/java/tconstruct/smeltery/blocks/SearedBlock.java
+++ b/src/main/java/tconstruct/smeltery/blocks/SearedBlock.java
@@ -2,6 +2,9 @@ package tconstruct.smeltery.blocks;
 
 import cpw.mods.fml.relauncher.*;
 import java.util.List;
+
+import com.sun.istack.internal.logging.Logger;
+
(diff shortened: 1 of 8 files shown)


bonii-xx:
Thanks for the effort and while I appreciate what you tried to do there are a few problems with it:
1. Don't put multiple features into a single pull request
2. Don't comment things out, this makes the actual changes harder to see. Delete them, we got a version history.
3. Faucets can't be turned off anymore
4. The change to the tool crafting does not conform to the expected behaviour of a crafting UI. It'd need a different kind of UI or an extra button.
5. You're changing state in a getter/check that should not have any side effects (canTakeStack) which can lead to a lot of unexpected behaviour. Don't do that. Ever.
6. Sadly I cannot take any more complex changes and/or changes where there are many possible unforseen side-effects, since that would require a lot of testing, and this version of tinkers will probably not get updated anymore.

PR #1573, opened 2015-04-25
Books translated

TheVikingWarrior:
I translated all the books, however i'm not sure it works since when i tried to put the file in the folder it didn't work


Adaptivity:
Strange, it should work just fine. You should change game language first and then restart the game, so changes will apply.

TheVikingWarrior:
I'll try again

TheVikingWarrior:
The two volumes of "Materials and you" are in english, the other 3 are fine a work

Adaptivity:
Just added your .xml files and they works for me :smile:
But there a small crash problem in materials and you volume II. If you are want to translate titles of materials ([like this](https://github.com/TheVikingWarrior/TinkersConstruct/blob/patch-9/resources/assets/tinker/manuals/it_IT/materials.xml#L491)) you should add [this](https://github.com/TheVikingWarrior/TinkersConstruct/blob/patch-9/resources/assets/tinker/manuals/ru_RU/materials.xml#L503) (same for stone, iron etc, check ru_RU for these strings) after [this line](https://github.com/TheVikingWarrior/TinkersConstruct/blob/patch-9/resources/assets/tinker/manuals/it_IT/materials.xml#L494). Also, you shoudn't translate [this](https://github.com/TheVikingWarrior/TinkersConstruct/blob/patch-9/resources/assets/tinker/manuals/it_IT/materials.xml#L504) as I know.

TheVikingWarrior:
Thank's i'll try, i don't know what happened in the end, probably a bit too much Ctrl H :)

TheVikingWarrior:
I don't know, the two books won't translate, even with the fix

Adaptivity:
Fix only prevent crash. I dunno why books not working for you.

bonii-xx:
Do all the books work now?

TheVikingWarrior:
Nope, firstday.xml end materials.xml don't work now, tried with TConstruct 1.8.4a

Adaptivity:
@TheVikingWarrior are you ok if I'll add your translations in my next PR? Because 7 commits is too much :)

TheVikingWarrior:
Yeah, that's fine

TheVikingWarrior:
Added in https://github.com/SlimeKnights/TinkersConstruct/pull/1613

PR #1577, opened 2015-04-28
Right Click Placement Fix

SamRaven2:
A re-write of the logic for placing things with tools on right click. Fixes the inaccurate logic when a player is on the edge of the block as well as allows items such as torches to be placed when the player is in the same space as them.

Allows for placing blocks underneath player more consistently (noob-towers)
Allows torches/webs/saplings/etc to be placed even if player is in same block space
Removes minor chance of being inside block when placed causing suffocation damage

### DIFF
--- a/src/main/java/tconstruct/library/tools/HarvestTool.java
+++ b/src/main/java/tconstruct/library/tools/HarvestTool.java
@@ -5,6 +5,7 @@ import cpw.mods.fml.relauncher.Side;
 import cpw.mods.fml.relauncher.SideOnly;
 import net.minecraft.block.Block;
 import net.minecraft.block.material.Material;
+import net.minecraft.block.material.MaterialLogic;
 import net.minecraft.client.Minecraft;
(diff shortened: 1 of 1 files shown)


SamRaven2:
Okay, I think that's all the commits for now. Mostly seems to be fixed.

bonii-xx:
Looks ok. Please squash the commits though.

SamRaven2:
Not super familiar with git yet, is there an easy way to do that or will I need to create a new pull?

Wulfson:
No need to create a new PR.
http://davidwalsh.name/squash-commits-git
The push -f at the end will update this PR.

SamRaven2:
So as much as I would like to say that helped... it unfortunately did not. In the process of trying to follow it, I'm afraid that I might have messed up my working directory.

SamRaven2:
oh, there we go
All good now

bonii-xx:
Squashing can be weird the first time, yes. ;)
Thanks

Issue #1434, opened 2015-02-27
Travelers Belt not switching as expected...


Gunner76th:
So the issue is, when I try to access my Traveler's Belt, I push the V key (which is what I have it currently bound to due to keybind conflicts). When I do this, most of the time nothing happens, unless I sit there and rapidly push the V key several times in a row. Usually it takes about 5-8 pushes before the inventory in the hot bar changes over. I know earlier versions of the mod did not have this issue, so this seems to be something recent. 

Mods in pack: http://pastebin.com/34fmfwr8

Version of forge in use: Server 1291 Client 1291
Version of MC: 1.7.10
Version of Java: Server 1.7.0_u72 Client 1.7.0_u71
OS: Server Ubuntu 14.4 Client Windows 7 x64

thr33p10:
I'll second this issue, but reproduced and "fixed" perhaps differently.
When I first log into the server my belt key "B" does not switch the hotbar.. I only tried once or twice, never persisting until it worked.
All I do to get around the issue is go to inventory, click the tab that allows access to where the belt is equipped, then exit inventory.  Now I can access the belt. In addition, after accessing that part of the inventory, My character also shows his gloves, et al. from that equipment screen.

bonii-xx:
Try reproducing with only tinkers

Wulfson:
I haven't reproduced with just Tinker's, but I'll note that this happens for me in the Direwolf20 pack, versions 1.0.3 and 1.0.4. Interestingly, when I've got the belt (and only the belt) keybound to B, I have the issue originally described, but if I hit Control+B it works every time (despite not having used Control when I did the binding).

laz2727:
Keybind collision?

2015-03-04 21:07 GMT+06:00 Josh Hancock <notifications@github.com>:

Wulfson:
In my case, I specifically removed all other bindings. Note: "I've got the belt (and only the belt) keybound to B"

Gunner76th:
So ever since Wulfson posted about using the ctrl key (on a windows machine) I have been using this, and the belt switches as expected every time. If I do not use ctrl along with the assigned key (foe me it is V) I have to push it several times before it switches. 

With that being said, I can confirm that in the version this thread was posted for, it seems that the keybind for the belt is ctrl+ whatever key you select, even though in none of the documentation anywhere states this.

Issue #1552, opened 2015-04-18
Knapsack, Traveller's Belt, Haste Gloves not working until TiC tab opened. (No sync on load)


TherminatorX:
#148 should be actually fixed. Was brought to my attention by a user on my wiki, and confirmed to still be a problem. I've often had this issue before, and had no idea what it was. Very annoying.

PR #1585, opened 2015-04-30
Fix for #1552

covers1624:
The Server was not syncing the players TPlayerStats to the client until
the Armor tab was opened.
Added Packet.
Send Packet on EntityJoinEvent.


covers1624:
This may also fix #1434, as that was the same sync issue, have tested and it seems to be fixed.

bonii-xx:
Looks ok. I know it's not called currently, but change the logging to debug to prevent issues down the road.

PR #1475, opened 2015-03-15
Move TE strings out

Adaptivity:
They handled by TE now. Can be merged when https://github.com/SlimeKnights/IguanaTweaksTConstruct/issues/115 gets fixed.

PR #1613, opened 2015-05-09
Remove TF material strings

Adaptivity:
Plus some other localization fixes. Also, added Italian localization for manuals by @TheVikingWarrior. Closes #1573
Reworked version of #1475


bonii-xx:
Nice. Do all the books work now? The materials added by TF have their own localization in TF right?

Adaptivity:
1) The books(Italian books) working for me. I dunno why they not working for @TheVikingWarrior might be an issue on his side. 2) [Yes, they have](https://github.com/CoFH/ThermalFoundation/blob/master/src/main/resources/assets/thermalfoundation/lang/en_US.lang#L296)

bonii-xx:
Well, he said only one or two didn't work but the remaining ones did.

Adaptivity:
All of them works for me... anyway, merge this, there is nothing what I can fix.

bonii-xx:
Okay, just wanted to make sure. Thanks Adaptivity :)

PR #1627, opened 2015-05-19
Aluminum Brass Fluid name

mDiyo:
Aluminum Brass' fluid name is now named "aluminumbrass" instead of "aluminiumbrass"


bonii-xx:
Actually it's "aluminumbrass.molten", like before: https://github.com/SlimeKnights/TinkersConstruct/blob/a2226a4ee8fae61299d9b2f0fe0b7094c34674d4/src/main/java/tconstruct/smeltery/TinkerSmeltery.java#L328
All your PR would do is break molten alubrass buckets.

PR #1661, opened 2015-06-19
Fixed Lumber Axe Exploit

auxiliary-character:
This exploit: http://youtu.be/4KW0AwXb1zk

### DIFF
--- a/src/main/java/tconstruct/items/tools/LumberAxe.java
+++ b/src/main/java/tconstruct/items/tools/LumberAxe.java
@@ -98,7 +98,7 @@ public class LumberAxe extends AOEHarvestTool
             if(detectTree(world, x,y,z, wood)) {
                 NBTTagCompound tags = stack.getTagCompound().getCompoundTag("InfiTool");
                 int meta = world.getBlockMetadata(x, y, z);
-                breakTree(world, x, y, z, stack, tags, wood, meta, player);
+                breakTree(world, x, y, z, x, y, z, stack, tags, wood, meta, player);
                 // custom block breaking code, don't call vanilla code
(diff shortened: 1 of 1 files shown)


bonii-xx:
I don't see any bug?

Syer10:
I thought that was how it was suppose to work.

auxiliary-character:
It's supposed to keep going until it StackOverflows?

bonii-xx:
Show me a tree that's high enough to cause a StackOverflow ;)

Syer10:
Sacred rubber tree?

auxiliary-character:
Yeah, a sacred rubber tree would probably do it. I just MCEdited in a huge cube of wood for the video, though. I imagine you could get a problem really fast if someone wanted to make a dimension out of wood or something.

auxiliary-character:
The fix is that it doesn't continue outside of an ellipsoid defined [here](https://github.com/auxiliary-character/TinkersConstruct/blob/df0259793e451be88b09fdc754bc6a5bd5e5a785/src/main/java/tconstruct/items/tools/LumberAxe.java#L180).

bonii-xx:
Okay, I see what you're getting at. The ellipsoid is a nice touch. Bonus points if you show me the effect on said huge cube of wood. ;)

Thank you for your contribution. I'll pull it in so the fix is in a dev version, however there are no further releases planned. If enough fixes come together I might do one final update.

MyDeloreanWontStart:
sacred rubber cathedrals

AgentK9:
0.0

bonii-xx:
I like the look of it :D

PR #1674, opened 2015-06-26
Added OreDictionary for a couple recipes

ganymedes01:
I'm mostly looking for the "chestWood" support as my mod adds new chests and users have been complaining that they can't use them to craft the Pattern Chest (despite them being able to turn my chests into the vanilla chest quite easily).

The "dustRedstone" I added just because I stumbled upon it when looking for where the recipe for the Pattern Chest was added and I thought "meh why not?" lol


bonii-xx:
Yeah, oredict integration is severely lacking. Looks ok.

PR #1713, opened 2015-08-02

CandiceJoy:
This makes the DiamondApple's getUnlocalizedName() method more robust by disallowing any possibility of invalid array indices.


bonii-xx:
* Use 0 as the default value for bot hcases, It's invalid in both cases and should have the same outcome.
* Your code is pretty verbose for a simple range check. Compact to 1 if and 1 variable.

CandiceJoy:
I'll open another PR.

PR #1714, opened 2015-08-02

CandiceJoy:
This makes the DiamondApple's getUnlocalizedName() method more robust by disallowing any possibility of invalid array indices.  I have also added the tweaks boni wanted.


squeek502:
You could compress it even further:

```java
// clamp index between 0 and itemNames.length - 1
int nameIndex = Math.max(0, Math.min(itemNames.length - 1, itemstack.getItemDamage());
return (new StringBuilder()).append("item.food.").append(itemNames[nameIndex]).toString();
```

PR #1715, opened 2015-08-02

CandiceJoy:
Fixed ArrayIndexOutOfBoundsException on DiamondApple getUnlocalizedName(); compressed to the max


bonii-xx:
I'd have actually preferred the previous one for readability, but that's ok too.

PR #1747, opened 2015-09-08
Update ArmorControls.java

TheTemportalist:
Updates for NeK 2.0.1, and in the process fixing https://github.com/mod-warriors/NotEnoughKeys/issues/34
Requires testing (could not get TiCon to setup workspace on local machine, changes were made via github.com).

### DIFF
--- a/src/api/java/modwarriors/notenoughkeys/api/Api.java
+++ b/src/api/java/modwarriors/notenoughkeys/api/Api.java
@@ -25,13 +25,14 @@ public class Api {
 	 * Registers a mod's keys with NEK
 	 *
 	 * @param modname        The NAME of the mod registering the key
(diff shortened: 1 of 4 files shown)


bonii-xx:
I'm not pulling any untested changes. What's furthermore is that it's caused by an API change on your side. That shouldn't break anything.

TheTemportalist:
Api changes are meant to be changed on the side of those implementing it. APIs are never completely finished. That's why we use SemVer versioning.

Adaptivity:
@bonii-xx are you going to pull this?

bonii-xx:
I thought I already head. Probably wanted to wait in case more fixes for it were coming.

Adaptivity:
Thanks!

PR #1751, opened 2015-09-12
Hide SlotAccessory id 0

cblair91:
Hide the SlotAccessory with id 0, as users get confused wondering what goes there, and nothing as of yet is implemented to go into that slot.


bonii-xx:
Master moved to 1.8

PR #1771, opened 2015-10-03
Hide technical stuff from NEI

Adaptivity:
These:
javaw 2015-10-03 13-42-14-11
Tested and it's working.


Adaptivity:
ping @bonii-xx

bonii-xx:
The better course of action would be to not add those items/blocks to the creative menu, instead of hiding them in NEI. The boneana already does this.

Adaptivity:
These items aren't available in creative tabs. However, they are available in NEI.

Adaptivity:
So, what do you think?

bonii-xx:
See what I said above.

PR #1790, opened 2015-10-29
Update fr_FR.lang

Mazdallier:
to continued at toolstation...


resources/assets/tinker/lang/fr_FR.lang
```diff
@@ -538,402 +538,402 @@ tile.fluid.molten.enderium.name=Enderium fondu
...
-gui.toolstation18=Modifiers remaining: 
-gui.toolstation19=Damage Reduction: 
-gui.toolstation20=Protection: 
-gui.toolstation21=Ammo: 
-gui.toolstation22=Break Chance: 
-gui.toolstation23=Max. Attack: 
-
-gui.mining1=Stone
-gui.mining2=Iron
+grout.tooltip=J'enduit !
```
OSM-Cait: I believe this part should be changed back.  The grout part refers to Groot from Guardians from the Galaxy who constantly says "I am Groot!"  However, this doesn't work since grout translates to enduit in French.

Mazdallier:
OK i take it back...

PR #1792, opened 2015-10-29
Updated source compat to Forge 1.7.10-10.13.4.1448 and snapshot 20140925

techbrew-mc:
I did a somewhat lengthy update to work with newer MCP mappings (20140925), but only to get it compiling.  I recommend you take a hard look at the changes to make sure the behavior is consistent.


bonii-xx:
You crazy man. :D

bonii-xx:
Going to close this for master since we moved it to 1.8

PR #1828, opened 2015-12-06
Update cs_CZ.lang

Cronosus:
update


Cronosus:
bad req.. something missed

PR #1841, opened 2015-12-22
Refactor pattern costs to allow being overridden

UndeadZeratul:
By default the values will still remain the same, but if someone else wants to adjust these values, they can.

### DIFF
--- a/src/main/java/tconstruct/tools/items/Pattern.java
+++ b/src/main/java/tconstruct/tools/items/Pattern.java
@@ -1,7 +1,11 @@
 package tconstruct.tools.items;
 
 import cpw.mods.fml.relauncher.*;
+
+import java.util.HashMap;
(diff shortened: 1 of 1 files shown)


bonii-xx:
You got the right idea, but there's no point in adding it to 1.7.10 and it's already done in 1.8.
There wont be a new 1.7.10 release, but I can pull it if you need it for some internal things.

UndeadZeratul:
Yeah, I know it doesn't exactly fix any bugs, but for something we're doing for our modpack using our own tweaks mod, we were looking to tweak every tool part cost, and since that method is hardcoded, that's pretty difficult to do.  If you don't mind pulling it in that'd be great, but if not, we can always just build and distribute TiC manually if that's ok with you.

UndeadZeratul:
Hey, thanks!

PR #1880, opened 2016-01-15
Fix table recipes in JEI

mezz:


mezz:
Closed because of https://github.com/SlimeKnights/TinkersConstruct/commit/a858ce4e8c8e9a4c7a964da768780f1117b59a30

PR #1884, opened 2016-01-17
Fix JEI recipe transferring in the Crafting Station

shadowfacts:
Previously, (if a side chest was present) only the chest slots would be used, and the player inventory slots would be ignored.

Fixes mezz/JustEnoughItems#128

### DIFF
--- a/src/main/java/slimeknights/tconstruct/plugin/jei/CraftingStationRecipeTransferInfo.java
+++ b/src/main/java/slimeknights/tconstruct/plugin/jei/CraftingStationRecipeTransferInfo.java
@@ -1,0 +1,44 @@
+package slimeknights.tconstruct.plugin.jei;
+
+import mezz.jei.api.recipe.VanillaRecipeCategoryUid;
+import mezz.jei.api.recipe.transfer.IRecipeTransferInfo;
+import net.minecraft.inventory.Container;
+import net.minecraft.inventory.Slot;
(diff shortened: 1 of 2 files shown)


bonii-xx:
Now it doesn't work at all.

shadowfacts:
I tested and it works fine, both with a side chest present and without.

What would you like me to do instead?

bonii-xx:
Looked at again, and it looks like only some recipes work. Gold Ingots -> Gold Block doesn't show up, but Gold Block -> Ingots does.

For the Arraylist either use something that doesn't create a worstcase runtime situation, or a more functional approach like Arrays.copyOfRange.

shadowfacts:
`getInventorySlots` now creates an empty ArrayList and adds all the slots after 10 to the list, instead of adding all the slots and remove the first 10.

It could be done using Arrays.copyOfRange like so:
```
return Arrays.asList(Arrays.copyOfRange(container.inventorySlots.toArray(new Slot[0]), 10, container.inventorySlots.size()));
```
but, IMO, the for loop solution is much cleaner.

bonii-xx:
Perfect, it's working great now. Can you squash the commit?
List stuff is fine, I guess one could preinit the list with the needed size, but it shouldn't matter.

shadowfacts:
Squashed

bonii-xx:
Thanks a lot

PR #1925, opened 2016-02-22
Updating The Cutlass, Rapier and Scythe textures as per Boni's request.

Firedingo:
Old Tcon Textures for the Cutlass, Rapier and Scythe have been copied, renamed and the texture adjusted as per Boni's request.


laz2727:
No customs?

bonii-xx:
Merged onto master

PR #2001, opened 2016-04-06
config option for invulnerability of dropped tools

alexbobp:
Two new config options, which both default to the current behavior, for:
- dropped tools being immune to lava and having a much longer despawn timer
- global item despawn timer for all items, not just tcon tools (only affects tcon tools if the previous option is turned off)

On my server I like to make tinker's construct tools not special, but instead give all items a 15 minute despawn.

### DIFF
--- a/src/main/java/tconstruct/TConstruct.java
+++ b/src/main/java/tconstruct/TConstruct.java
@@ -7,6 +7,7 @@ import cpw.mods.fml.common.Mod.EventHandler;
 import cpw.mods.fml.common.Mod.Instance;
 import cpw.mods.fml.common.SidedProxy;
 import cpw.mods.fml.common.event.*;
+import cpw.mods.fml.common.eventhandler.SubscribeEvent;
 import cpw.mods.fml.common.network.NetworkCheckHandler;
 import cpw.mods.fml.common.network.NetworkRegistry;
 import cpw.mods.fml.common.registry.GameRegistry;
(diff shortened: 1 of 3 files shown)


alexbobp:
By the way if you decide that the option for global item despawn is out-of-scope, I'd be happy to modify this PR to only include the other change.  I put it in tinker's construct only because it serves as a middle-ground alternative to the existing invulnerability.

bonii-xx:
Tools are going to stay indestructible, I don't plan on having a config option for that. If people don't want them they can always one of the numerous trashcans in several mods.

Why do you want to add global item despawn timer configs to TiC? I don't really see how it makes sense to have it in TiC, maybe as a separate mod or part of a tweak-mod?

alexbobp:
This isn't about letting people throw away tools.  It's a balance change to bring tinker's construct more in line with other available sets of tools.  Before this change, all of my players used only tinker's construct tools, and ignored several other mods with a variety of options, because they could die in lava without losing them, and I consider that a balance issue.  What's wrong with a config option for that?

And like I said, I just threw in the global despawn tweak to offset the nerf on my personal server.  I will gladly take that part out of the pull request if desired.

alexbobp:
I just think that with features that make major changes to game balance assumptions (ie, if you die in lava, your stuff is gone), there should be a config option if it's easy to add (which it was).

alexbobp:
Ok, I had the chance to talk with boni on irc.  It seems that we have a fundamental difference of opinion about this config option.  I will close the PR and start maintaining a tcon fork with minimal balance changes.

PR #2018, opened 2016-04-16
Added item racks and drying racks

KnightMiner:
Item and drying racks

Drying racks since 1.7.10 have the new ability to be placed on the ceiling, better handling of hoppers, better placement logic, and automatically add variants for all wooden slabs found in the ore dictionary. Item racks are a non-ticking form of drying racks

Recipes include jerky in all meats, fish, and Tinkers slime types, dead bushes from saplings, dry sponge from wet sponges, and dried clay (a brown variant of clay in normal and brick forms). All recipes can be viewed using JEI.


resources/assets/tconstruct/blockstates/rack.json
```diff
@@ -0,0 +1,22 @@
...
+            "north":  { "transform": { "translation": [ 0, 0, -0.375 ] } },
+            "south":  { "transform": { "translation": [ 0, 0,  0.375 ] } },
+            "east":   { "transform": { "rotation": { "y": 270 }, "translation": [  0.375, 0, 0 ] } },
+            "west":   { "transform": { "rotation": { "y": 270 }, "translation": [ -0.375, 0, 0 ] } }
+        },
+        "drying": {
+            "true": { "model": "tconstruct:drying_rack" },
+            "false": {}
+        }
+    }
```
bonii-xx: Add [inventory variant](https://github.com/SlimeKnights/TinkersConstruct/blob/master/resources/assets/tconstruct/blockstates/faucet.json#L14), can be defined like any other blockstate too

resources/assets/tconstruct/models/block/drying_rack.json
```diff
@@ -0,0 +1,6 @@
+{
```
bonii-xx: File can be removed, since it's redundant

resources/assets/tconstruct/models/block/item_rack.json
```diff
@@ -0,0 +1,6 @@
+{
```
bonii-xx: File can be removed, since it's redundant

resources/assets/tconstruct/models/block/rack.json
```diff
@@ -0,0 +1,28 @@
...
+    "parent": "block/block",
+    "display": {
+        "gui": {
+            "rotation": [ 30, 225, 0 ],
+            "translation": [ 0, -3, 0 ],
+            "scale": [ 0.625, 0.625, 0.625 ]
+        }
+    },
+    "textures": {
+        "particle": "#bottom"
```
bonii-xx: Rename the #bottom texture to #leg, for better logical coupling with the table model. Also define the bottom/leg texture with a default here, or put a default texture into the blockstate, to remove the block models.

resources/assets/tconstruct/models/item/drying_rack.json
```diff
@@ -0,0 +1,3 @@
+{
```
bonii-xx: Remove file, see inventory blockstate variant

resources/assets/tconstruct/models/item/item_rack.json
```diff
@@ -0,0 +1,3 @@
+{
```
bonii-xx: Remove file, see inventory blockstate variant

bonii-xx:
Left you some comments on things that can be simplified.

Regarding the transform situation... it is currently impossible to create the table models based on the blockstate variants. This leaves us with 2 choices: adding the things needed to the baked table model logic, or use the original approach you had, with 2 different block models for on-the-wall and on-the-ceiling, since those only require rotation around the Y axis, which is handled by the baked table model already.

resources/assets/tconstruct/models/block/drying_rack.json
KnightMiner: The reason I had this file was for the sake of resource packs that wanted different models for item/drying racks, as under the current system the model has to point to hardcoded path.

bonii-xx: Incorrect, they can always change it in the blockstate json.

KnightMiner:
I've added the changes suggested and managed to get the offset working as well.

It does seem that some other code in Tinkers Construct relies on some code that is not in the latest Mantle Jenkins build, so the automatic build is into passing.

PR #2028, opened 2016-04-21
Fix table rendering, fixes #2025

### LINKED ISSUE #2025: Items are sometimes flipped in the smeltery table
I need to test this to see if it depends on direction, though.

EDIT: It seems that they are all pointing in 1 direction...

Should be an easy-ish bug to fix, I think...

KnightMiner:
Tables legs now have proper UV (instead of displaying the full texture on the bottom).

Additionally, items no longer display upside down, meaning they properly line up with the cooling animation rendered by the casting table (instead of flipping when it finishes)

**Before:**
Before

**After:**
After


bonii-xx:
I originally had the bottom of the feet with the full texture to give it a more direct touch of the block. But I guess it does look a bit weird in the overall context, since it's a much higher resolution then.

If you want the change, please also adjust the tool forge.

KnightMiner:
Well, you always have [the main bottom texture](http://i.imgur.com/SqhwyW3.png), which contains most of the original texture (and the remainder is shifted onto the feet)

I did not notice the tool forge used a separate model as well, fixed.

Issue #2024, opened 2016-04-19
Glowball glowing blocks need a placement check


bonii-xx:
Glowballs can be attached anywhere, they need to remove themselves if the block they're connected to is not solid on their side.

They also need to check when a neighbor is broken, to see if they have to remove themselves. Think torches/ladders.

KnightMiner:
I looked into the vanilla code and it looks fairly straight forward, so I can have something ready later today. I think I'll have it check if the placement is valid both when the entity places the block and on block update.

bonii-xx:
Yes, the block can just check it itself on placement. The torches etc. do that too iirc. Please do, I already wanted to forward it to you actually. ;)

PR #2037, opened 2016-04-24
Add facing checks (#2024) and pick block to glow

KnightMiner:
Glow is now dependant on a block, and upon placing if the face it is
placed on is not solid it will automatically find a solid face (or break
if none is found)

Pick block on a glow now returns a glowball if available, otherwise does
nothing

### DIFF
--- a/src/main/java/slimeknights/tconstruct/shared/block/BlockGlow.java
+++ b/src/main/java/slimeknights/tconstruct/shared/block/BlockGlow.java
@@ -1,5 +1,8 @@
 package slimeknights.tconstruct.shared.block;
 
+import java.util.List;
+import java.util.Random;
+
 import com.google.common.collect.ImmutableMap;
 
 import net.minecraft.block.Block;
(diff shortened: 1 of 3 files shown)


src/main/java/slimeknights/tconstruct/shared/block/BlockGlow.java
```diff
@@ -51,10 +57,65 @@ public int getMetaFromState(IBlockState state) {
   public IBlockState getStateFromMeta(int meta) {
     return getDefaultState().withProperty(FACING, EnumFacing.getFront(meta));
   }
+  
+  @Override
+  public ItemStack getPickBlock(IBlockState state, RayTraceResult target, World world, BlockPos pos, EntityPlayer player) {
+    // use the glowball for the pickblock, as it is more useful than the technical block
+    if(gadgetsLoaded) {
```
bonii-xx: I'd reduce this to a simple `if(TinkerGadgets.throwball != null)`. That way there is no hard dependency on the pulse.

KnightMiner: Done in latest commit

PR #2046, opened 2016-04-28
Added attack speed boost to haste and lightweight

KnightMiner:
Each level of haste gives an additive 20% speed boost, and lightweight gives a 10% boost.

In addition, the attack speeds of the frying pan and battle signs were changed to be better balanced with the other weapons and the ToolCore attack speed method was made abstract


src/main/java/slimeknights/tconstruct/library/tools/ToolNBT.java
```diff
@@ -12,13 +12,15 @@
...
   public float speed; // mining speed
+  public float attackSpeed;
   public int modifiers; // free modifiers
 
   public ToolNBT() {
     durability = 0;
     harvestLevel = 0;
     attack = 0;
     speed = 0;
+    attackSpeed = 0;
```
bonii-xx: Initialize with 1 here, remove assignment of 1 in all the tool data building. As long as there is no other value set by something else it should always be 1. Also rename it to attackSpeedMultiplier, to not confuse it with the actual attackspeed stat of the tool (which is the product of the value + the tools AS)

src/main/java/slimeknights/tconstruct/tools/modifiers/ModHaste.java
```diff
@@ -17,38 +19,61 @@ public ModHaste(int max) {
...
+      speed += level * 0.5f;
+  
+      // save it to the tool
+      speed -= toolData.speed;
+      speed += tag.getFloat(Tags.MININGSPEED);
+      tag.setFloat(Tags.MININGSPEED, speed);
     }
+    
+    // attack speed: each level adds 0.2 to the multiplier, meaning level 1 is 1.2 times the default speed, and 5 is 2.0 times it
+    float attackSpeed = 1.0f + (level * 0.2f);
```
bonii-xx: Change it to apply a speed buff per redstone applied, not per level. Make it linear: attackspeed = 1 + (0.2 * current / max)

src/main/java/slimeknights/tconstruct/tools/modifiers/ModHaste.java
```diff
@@ -17,38 +19,61 @@ public ModHaste(int max) {
...
-
-    // save it to the tool
-    NBTTagCompound tag = TagUtil.getToolTag(rootCompound);
-    speed -= toolData.speed;
-    speed += tag.getFloat(Tags.MININGSPEED);
-    tag.setFloat(Tags.MININGSPEED, speed);
+    // save it to the tool, adding the original boost from lightweight (if it exists)
+    attackSpeed -= toolData.attackSpeed;
+    attackSpeed += tag.getFloat(Tags.ATTACKSPEED);
+    tag.setFloat(Tags.ATTACKSPEED, attackSpeed);
```
bonii-xx: Oh my, I notice that this really was one of the first modifiers I wrote. Change it to properly use the ToolNBT object wherever possible and assign the stats like this: https://github.com/SlimeKnights/TinkersConstruct/blob/master/src/main/java/slimeknights/tconstruct/tools/modifiers/ModDiamond.java#L30

Prevent manual tag assignment wherever possible.

bonii-xx:
Tested behaviour and stuff locally, looks solid. Once updated we can merge it in :)

KnightMiner:
Updated to include the suggested changes.

PR #2048, opened 2016-04-29
Add vanilla shovel rightclick ability to Tinkers shovels + excavators

KnightMiner:
Allows shovels to do the vanilla shovel feature of making grass paths. It will use AOE.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tools/item/Shovel.java
+++ b/src/main/java/slimeknights/tconstruct/tools/item/Shovel.java
@@ -1,10 +1,23 @@
 package slimeknights.tconstruct.tools.item;
 
+import com.google.common.collect.ImmutableList;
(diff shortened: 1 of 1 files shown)


bonii-xx:
Still needs a check if the tool is broken, also looking at the mattock again, this should be easier: 
https://github.com/SlimeKnights/TinkersConstruct/blob/master/src/main/java/slimeknights/tconstruct/tools/item/Mattock.java#L108-L127

However this does not allow you to rightclick on path to extend it. Hardcode that part in the onItemUse directly. This way we stay more flexible and compatible.

KnightMiner:
Fixed. I was able to use almost the same code as the mattocks actually, I did not think about them doing AOE (the expanders are a little new). Also squashed.

PR #2106, opened 2016-05-23
Fix player heads dropping with the wrong metadata

ganymedes01:
Also, can we get something so I can hook [Headcrumbs](http://minecraft.curseforge.com/projects/headcrumbs) into this withoug much hassle? <3

I saw the TODO on the top of the class and I must say that a Class->ItemStack registry would not suffice at all. Some entities have more than one head (like the vanilla Skeletons). It would be better if there was an actual fired Event, or some sort of registry for head drop handlers of some sort (a simple class with a ```ItemStack getHead(Entity entity)``` method).

If I get an "official" answer on what would be preferred I'd be more than willing to make the PR myself

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tools/modifiers/ModBeheading.java
+++ b/src/main/java/slimeknights/tconstruct/tools/modifiers/ModBeheading.java
@@ -105,7 +105,7 @@ public class ModBeheading extends ToolModifier {
     }
     // meta 3: player
     else if(entity instanceof EntityPlayer) {
-      ItemStack head = new ItemStack(Items.SKULL, 1, 4);
+      ItemStack head = new ItemStack(Items.SKULL, 1, 3);
       NBTTagCompound nametag = new NBTTagCompound();
       nametag.setString("SkullOwner", entity.getDisplayName().getFormattedText());
       head.setTagCompound(nametag);


bonii-xx:
Good thing it's only a todo. Yeah the todo is there so we can get some cross mod compatibility going between all the mods that add heads, I'm completely open for suggestions. You know better what is needed.

Also thanks for the metadata fix.

ganymedes01:
What about something like this:
```
public class IHeadDropHandler {

		ItemStack getHead(Entity entity);
}
```

Mods would register their instances of the interface above to some registry somewhere. Tinker's would then iterate through the list of handlers and if one of them returns a non-null value that would be the dropped ```ItemStack```.

I'm not very familiar with the Tinker's Construct code, so I'm afraid I'd end up making a mess if I tried PRing something like this XD If someone could tell me where I could shove the registry and interface I'd appreciate it :3

bonii-xx:
That still requires all mods to know the interface class. I'm considering using the function-IMC for it, keeping a simply registry with all Function<> handlers. Either general handlers or a entityclass->handler registry.

ganymedes01:
I didn't know about the IMC Function handlers... they certainly seem like a good option.
A ```Function<Entity, ItemStack>``` would certainly work for me

Issue #2127, opened 2016-06-01
Faucet has no texture [1.9.4]


eriico39hi:
Exactly as title says. Faucet has no texture in inventory. Still has texture in block form once placed.

JCMais:
To add to that.

The gold ingot cast has no texture when placed in the casting table. I didn't tested any other casting, so idk if it's for every one of them.

KnightMiner:
I cannot reproduce the issue with the cast being invisible, though that would be a different issue as the cause of the faucets being invisible does not apply to casts.

JCMais:
@KnightMiner I will create a new issue when I get the time to test it again.
Could have been optifine/shaders.

PR #2131, opened 2016-06-02
Fix faucet item and fancy frames appearing invisible

KnightMiner:
It seems something in 1.9.4 broke Forge item blockstate display tags, causing the faucets (as items) and fancy frames to appear invisible (#2127). Switching to a vanilla display tag fixed the issue, and there really is no need to use the Forge blockstate version in this case anyways.

### DIFF
--- a/resources/assets/tconstruct/blockstates/fancy_frame.json
+++ b/resources/assets/tconstruct/blockstates/fancy_frame.json
@@ -1,29 +1,14 @@
 {
     "forge_marker": 1,
     "defaults": {
-        "model": "item_frame",
+        "model": "tconstruct:fancy_frame",
         "textures": {
             "particle": "#wood",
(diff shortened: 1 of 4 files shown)


bonii-xx:
This only fixes the syptoms, not the cause, which seems to be in Forge somehow.

KnightMiner:
It's not so much fixing the symptoms but rather switching to a system that is known to work, as only the item display tags from the BlockStairs are causing issues, and Tinkers is not doing anything that is made simpler by using the BlockStairs version

bonii-xx:
I guess, well it can't hurt. Fix it on the fancy item frames too please.

KnightMiner:
Done. (since when did frames start using a block model? must have been added in 1.9)

I also took the chance to fix the fancy frame's display on the ground and in frames, since they appeared behind the frame and huge as items.

PR #2155, opened 2016-06-08
Add clear glass to Tinkers Construct

KnightMiner:
Picture

Includes regular and all 16 stained variants, plus molten glass.

Also adds a new seared glass block, which is essentially seared windows only with connected textures rather than fluid tank abilities (designed to look good as a large smeltery window)


src/main/java/slimeknights/tconstruct/common/TinkerOredict.java
```diff
@@ -70,6 +72,24 @@
 public class TinkerOredict {
 
   public static final String PulseId = "TinkerOredict";
+  public static final String[] dyes = { // makes oredict to int a bit easier in a couple other places
```
bonii-xx: Doesn't the OreDictionary have an array like that?

src/main/java/slimeknights/tconstruct/common/TinkerOredict.java
```diff
@@ -128,6 +148,14 @@ private static void registerCommon() {
     // Ores
     oredict(oreCobalt, "oreCobalt");
     oredict(oreArdite, "oreArdite");
+    
+    // glass
+    oredict(blockClearGlass, "blockGlass"); // no blockGlassColorless as then it is assumed as available for staining
+                                            // which blocks our own staining recipes
+    oredict(blockClearStainedGlass, "blockGlass");
+    for(int i = 0; i > 16; i++) {
+      oredict(blockClearStainedGlass, i, "blockGlass" + dyes[i]);
```
bonii-xx: What you're looking for here is an Oredict with OreDictionary.WILDCARD_VALUE

bonii-xx:
Nice, I like the seared glass.
What naming does vanilla use for its stained glass? "stainedGlass" or "glassStained"? My gut reflex says ours should be called "clear_glass_stained" rather than "clear_stained_glass", but we should match.

src/main/java/slimeknights/tconstruct/common/TinkerOredict.java
KnightMiner: Yes, but it is a local variable, [see here]https://github.com/MinecraftForge/MinecraftForge/blob/master/src/main/java/net/minecraftforge/oredict/OreDictionary.java#L252)

src/main/java/slimeknights/tconstruct/common/TinkerOredict.java
KnightMiner: Each one actually uses a different name, for example "blockGlassRed". I do wildcard it above for the generic glass oredict

KnightMiner:
Vanilla is just `stained_glass`. Vanilla normally tries to make it like it would be spoken rather than how one would normally name it internally

src/main/java/slimeknights/tconstruct/common/TinkerOredict.java
bonii-xx: Oh yeah, I was reading it wrong.

PR #2161, opened 2016-06-10
Add the ability for comparators to check the fluid level in tanks, drying racks, and casting tables/basins

KnightMiner:
Tanks return a number from 0 to 15 based on how full they are

Drying racks, casting tables, and casting basins return 0 if there is no output, and 15 if one exists


bonii-xx:
I"

KnightMiner:
Yeah, that is the indentation thing I talked about, I have disabled it since then but forgot to fix what it changed from there.

I won't be up e to update it for at least an hour at this point, so if you want to push the formatting now, I can update my PR from that (it already conflicts at this point anyways)

bonii-xx:
Don't worry. We're not in a hurry, and I got loads of stuff to fix tomorrow too >_>

KnightMiner:
Rebased and removed almost all of the automatic formatting.

PR #2210, opened 2016-06-30
Add slime channels

KnightMiner:
Picture

Add slime channels from 1.7, though with quite a few changes as stated below:

* Slime channels can only move in 4 directions instead of 8
* To make up for that, they can now be placed on walls, even allowing a slime channel elevator
* If placed directly next to another channel, it may connect
* The recipe now involves smelting congealed slime
* They are available in all slime colors

This commit also moves slime blocks and congealed slime from commons into world, to make it easier for the other modules to use them (as they are directly relevant to anyone wanting slime based stuff, for example slime channels use them).


KnightMiner:
To continue the discussion from IRC, the channel in this PR checks the powered state on collision with an entity (like was done in 1.7). This is in theory less efficient, though overall it is basically the same system used on clear glass except checking quite a few more blocks

Alternatively, the powered state could be stored in the metadata which would likely be more efficient (over checking redstone values for all the sides), though it requires either increasing the used IDs from 10 to 15 (losing the three spare colors we had before), or adding a tile entity to store the extra data (I would still go with just storing color there and keeping the 5 block sides, as otherwise the blockstate models would be a pain)

I am currently leaning in favor of the tile entity storing the colors and adding a powered bit to the metadata as I would prefer not to do that many block constantly in the channel.

----

A side note is the system I'm using does not really know which way the block is facing half the time (instead checking that all relatively), so it would be trivial to add a top block channel if desired.

----

Oh, and building it locally works fine, so I have no idea why the PR test build failed...

jriwanek:
:deobfMcMCP FAILED

FAILURE: Build failed with an exception.

* What went wrong:
Execution failed for task ':deobfMcMCP'.
> java.io.FileNotFoundException: /var/lib/jenkins/workspace/TCON_PRTESTS/build/depAts/mantle_at.cfg_Mantle-1.9.4-0.10.2.jenkins149_at.cfg (No such file or directory)

bonii-xx:
Okay, this is gonna turn into too big a mess. We're gonna outsource whatever into a data TE, that should be fast. Redstone will be cached of course. We can use the metadata for color or direction, not sure which is better at this point. Maybe we can also get the facing into the same block now.

KnightMiner:
Color would definitely be easiest, as channels don't much care about color meanwhile they consistently check the direction check for other channels (for connections and culling)

For the side, it would be nice to be able to store it all in one block, and I assume that is possible using an extended block state, though the main issue is Forge blockstates don't support adding two different rotations to a model (as in two properties each adding one), so I am not sure of the easiest way to do that part

KnightMiner:
I have updated the PR with the changes I mentioned. Basically the data for side and facing are not stored in a TE and pushed into the blockstate on world load (or on placement), and both color and powered are stored in the metadata. As a result, it now uses just one block/item ID, though I kept the multiple blockstates to make things a bit easier on the blockstate for rotations.

src/main/java/slimeknights/tconstruct/common/TinkerOredict.java
```diff
@@ -1,19 +1,5 @@
 package slimeknights.tconstruct.common;
 
-import com.google.common.eventbus.Subscribe;
```
bonii-xx: Your import settings are still broken ;P

src/main/java/slimeknights/tconstruct/gadgets/GadgetClientProxy.java
```diff
@@ -1,33 +1,78 @@
...
 import slimeknights.tconstruct.gadgets.entity.EntityFancyItemFrame;
 import slimeknights.tconstruct.gadgets.entity.EntityThrowball;
 import slimeknights.tconstruct.gadgets.item.ItemThrowball;
 import slimeknights.tconstruct.library.Util;
+import slimeknights.tconstruct.library.client.model.PropertyStateMapper;
+import slimeknights.tconstruct.shared.block.BlockSlime;
 
 public class GadgetClientProxy extends ClientProxy {
 
+  public static Minecraft minecraft = Minecraft.getMinecraft();
```
bonii-xx: Move to init

PR #2217, opened 2016-07-02
Fix brownstone interaction with water

KnightMiner:
Basically, the speed boost from brownstone was canceling out the deceleration from the water, leading to some strange behavior including excessive max speeds

The issue is quite a bit easier to produce using slime channels as they are not pushing you off the brownstone, thus why I did not notice it until now

### DIFF
--- a/src/main/java/slimeknights/tconstruct/gadgets/block/BlockBrownstone.java
+++ b/src/main/java/slimeknights/tconstruct/gadgets/block/BlockBrownstone.java
@@ -27,8 +27,14 @@ public class BlockBrownstone extends EnumBlock<BlockBrownstone.BrownstoneType> {
 
   @Override
   public void onEntityWalk(World worldIn, BlockPos pos, Entity entity) {
-    entity.motionX *= 1.25;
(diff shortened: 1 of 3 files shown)


bonii-xx:
I totally want to try that now. Waterslide ahoy

bonii-xx:
Hm. I kinda like the interaction. But so far it seems the multiplicator makes you speed up a ton, meaning you accelerate a lot more. Maybe we can get it into a place where, in water, you're still fast as long as you walk on it, but remove the positive feedback loop on acceleration? (Or we just put a max speed on it in water)

KnightMiner:
I know just adding a cap won't cause walking to ever slow down as it's less than the cap, but I could try just reducing the boost when in water and try to find a balance between fun and impractical

KnightMiner:
Alright, I've managed to get it to the point where if you stop moving the momentum stops, but other than that the behavior is mostly the same. The acceleration is still a bit greater than on land, but not as excessive as before.

bonii-xx:
Perfect

PR #2225, opened 2016-07-04
Fix slime channels not syncing facing data when on servers

KnightMiner:
Fixes the packet being missing and uses a more conventional approach for pulling data into the blockstate from the TE.

Also adds the first tweak to placement logic, which is the consistent sides. The rest I don't want to add until I add the additional directions.


bonii-xx:
Very nice, proper constant usage for tag ids.

PR #2228, opened 2016-07-06
Adds four diagonal slime channels and better placement logic

KnightMiner:
The diagonals are exactly as before, except slightly different in texture since I am using vanilla models rather than the tesselator.

New placement logic basically divides a block face into 9 locations, 8 on the edges all set a constant face, and clicking the center takes the player's facing (old behavior)

This commit features some backwards compatibility for those using the previous jenkins builds with the 4 direction slime channels (converts from 0-3 to 0-7). The compatibility can be removed some time later before the release


bonii-xx:
If you feel really motivated:

KnightMiner:
Yeah, it's a little hard to connect on a diagonal as it would not know which part to shrink/grow without expanding the connection logic to 5 values.

Might be something to keep on the ideas list for later, but not something I really want to do now.

PR #2234, opened 2016-07-08
Add modifier textures for blasting and mending moss

KnightMiner:
Picture

Blasting adds a tiny TNT to the tool, mending moss adds a small patch of glowing moss


bonii-xx:
Moss is mostly ok, for hammer and shovel it should probably be located in the center where the toolrod goes in.
How does it look if we make the yellow xp tone a bit greener so it's not that huge contrast-wise?

I toyed around with the blasting a bit since it didn't really match up as I expected it to. I redid the blasting textures based on your graphics, they're in the master branch now.

In the meanwhile I added a new modifier: glowing
We had unused textures for a glowstone modifier in the old resources: https://github.com/SlimeKnights/TinkersConstruct/blob/1.7.10/resources/assets/tinker/textures/items/pickaxe/glowstone_pickaxe_effect.png
Maybe you can work your magic and adjust them to current status. Make them more subtle and fix the color to match a gentle glow (aka very bright and soft yellow)

KnightMiner:
I remember those, it should be Rest enough to get something that looks good with the new tools.

I'll try your suggestions for moss out and see if I can make it look a bit better there. I'm free later today so the PR shouldn't be too far frpm updating

KnightMiner:
Here's what I have for glowing as a direct port of the texture, mainly for the sake of color. I am not currently sure of the best way to make the glow more subtle without it looking odd, as my first plan of transparency does not seem supported. I might go for something like I did with moss and just add it to a part of the tool (in this case, something like a round background glow around the handle rather than the middle of the head), or otherwise try a different effect entirely.

bonii-xx:
You still have the blasting files in the PR, glowing should be fine I think.

KnightMiner:
I haven't updated the PR yet, as I wanted a second opinion on glowimy first so I can finish a few more in that style.

bonii-xx:
Ah. Yeah, looks ok to me. It's a glow. ;D
On a more serious note, it should be fine, since it's the only thing actually going out and around the tool instead of covering it up.

laz2727:
Would look weird with other modifiers though.

KnightMiner:
If it's over another one it might, but under one it looks fine. I'll mess around with it just to make sure

KnightMiner:
I've updated the PR with tweaked mending moss and glowing. For mending moss, I moved the moss like you suggested on shovels and hammers, and made the yellow a bit greener.

For glowing, I found having a full border did not work well on tools that use the full 16x16 area (no padding), so I decided to just do the tool rod as that looked pretty good (and still looks fine on tool rods against the edge.

(My alternate idea for glowing was making the entire tool rod actually become a glowing rod (rather than just the edge, kinda like the blaze modifier), though of course we would lose the material color then, or maybe just replacing part of the rods texture (inside or outside), though the latter looked a bit odd.)

bonii-xx:
No, the glowing outline is better since it's still the same tool.. but glowing. I like it.

PR #2252, opened 2016-07-18
Add seared furnaces

KnightMiner:
Seared furnaces are basically furnaces that run off a smeltery tank, giving them convenience of use (as you don't need to consume a full solid fuel to cook a single item) and allow other smeltery fuels from the integration to work. To use, simply place a seared tank on any side of the seared furnace.

I do have a few stylistic questions that I am unsure of which way to go, as stated below:

* First off, the furnace currently shows its burning texture when the structure is valid, basically the same as the smeltery. My alternate idea here is to set it instead to having fuel, which while it makes more sense logically, you would lose the indication that the furnace structure "works" (outside of the GUI just not opening) and it would be a bit more complicated to implement with rendering block updates.
* Secondly, the furnace is currently balanced to be exactly the same as a vanilla bucket of lava, basically 1000mb does 100 items smelted at the same speed. Should this be tweaked to maybe be faster or in some other way better than a vanilla furnace? (as we are limiting the fuel choices)
* Third, the way I've implemented temperature does not affect fuel consumption, so a higher temperature liquid will consume less fluid as it takes less time. This seems to be the same as the smeltery, but it could still be concidered imbalanced as it uses overall less fuel then
* Lastly, should the seared furnace reset the smelting progress when it runs out of fuel? Smelteries don't, but vanilla furnaces do (I could also go with a winding down instead)


laz2727:
>Secondly
A furnace that works on lava losslessly is good enough, actually.
>Lastly
It probably should, but if you really feel the furnace is too weak, you can just remove it.

LovelyFlare:
Please make it smelt fast or else it would seem to be a waste of seared bricks ;~;

KnightMiner:
Picture

PR updated with the new version of the furnace. GUI shown [here](http://i.imgur.com/kaDG7Ym.png)

Basically, its now a multiblock made of seared bricks (though tanks can be in the frame). I still need to in game documentation for it and double check the cooking heat balance, but its complete other than that

laz2727:
You should just make it a cooking smeltery. It looks too complicated rn.

KnightMiner:
It is basically. Just cooks items in the slots, up to a full stack at a time (though larger stacks are slower)

Slot count is based on size

laz2727:
Oh, okay.

Looks hard to automate, maybe a searing hatch or something? Iirc controller
already gives the whole inventory though, so it's not top priority.

bonii-xx:
It's not meant to be automated.

KnightMiner:
Yeah, additionally it does not make sure an item is fully smelted before allowing it to be extracted. If you want to automate it, you either need filtered pipes, or you can just use a vanilla furnace.

LovelyFlare:
I still think a window would look better

KnightMiner:
The final design was based off of a medieval furnace, where the only access to the inside is a small window in the front (controller), so adding additional windows would definatelly look odd with that design in mind.

LovelyFlare:
Medieval furnaces don't have lava tanks :P

bonii-xx:
Currently building a furnace feels quite.. meh. It's a big block for little gain, totally not worth it. Bump the inventory slots per internal block up to 3, so a minimal size furnace has 3 slots. Maybe we should give a fixed 9 slot inventory + 3 per internal space. A minimal size smeltery needs 5 blocks. A minimal size furnace needs 26. It has to give more bang for its bug, for how clunky it is.

bonii-xx:
On a note of clunkyness and astethics, it would also be cool if we could use slabs and stairs for the ceiling:

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -0,0 +1,562 @@
...
+        currentTank = pos;
+        return;
+      }
+    }
+
+    currentTank = null;
+  }
+
+  // checks if the given location has a fluid tank that contains fuel
+  private boolean hasFuel(BlockPos pos, FluidStack preference) {
```
bonii-xx: Extract to heatingstructure and rename appropriately

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -0,0 +1,562 @@
...
+  }
+  
+  @Override
+  public void validate() {
+    super.validate();
+    // on validation we set active to false so the furnace checks anew if it's formed
+    active = false;
+  }
+  
+  public static final String TAG_FUELQUALITY = "fuelQuality";
```
bonii-xx: :+1: very good. Move to top of class or rather to the intermediate layer mentioned above since all that stuff is needed for everything anyway. Also `smelted` and `tankSide` are unused

bonii-xx:
I'm a bit worried about the cooking speed. Did you do the math to calculate the actual melting times? Currently a single stack is quite slow(er than in a vanilla furnace), but if I see this correctly a full stack might be quicker than an in a furnace?

Also food stuff should be cooked faster than the rest.

src/main/java/slimeknights/tconstruct/smeltery/block/BlockSearedFurnaceController.java
```diff
@@ -0,0 +1,175 @@
...
+    this.setCreativeTab(TinkerRegistry.tabSmeltery);
+    this.setHardness(3F);
+    this.setResistance(20F);
+    this.setSoundType(SoundType.METAL);
+
+    this.setDefaultState(this.blockState.getBaseState().withProperty(FACING, EnumFacing.NORTH).withProperty(ACTIVE, false));
+  }
+
+  @Override
+  public TileEntity createNewTileEntity(World worldIn, int meta) {
```
bonii-xx: You're missing all the NonNull annotations. Can eclipse generate that stuff? I can do it after the PR otherwise.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
KnightMiner: Whoops, both `smelted` and `tankSide` are leftover from older versions of the furnace code.

src/main/java/slimeknights/tconstruct/smeltery/block/BlockSearedFurnaceController.java
KnightMiner: Hmm, not sure. I'll have to look into that.

I fix those up when I add a few of the other fixed.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
KnightMiner: Okay, can do that. Is there an issue with renaming the fuel packet to match that?

KnightMiner:
For slot count, that should be easy enough to implement given how the slots are calculated. I can see how they would be a little low considering the minimum size would fit 9 regular furnaces. (so a starting total of 12 and adding 3 per inside space should be fine)

In the topic of heat, cooking temps may still need a bit more balance, but the idea is larger stack sizes cook the food faster as a whole, and like the smeltery you can divide items over multiple slots to save fuel. I can throw in a check to reduce the time it it's a good (or inversely increase it for non foods)

For slabs and stairs, that should be mostly simple, though I may need to tweak the structue checking code so frames can know if they are on the ceiling/floor. I'll either add a simple Boolean for the function stating as such, or create a couple of new block checking functions for frame ceiling and frame floor, whichever feels cleaner.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
bonii-xx: Why do you want to rename the packet? The function above checks if there if the fluid tank currently selected still has a fuel liquid in it, not wether the smeltery still has fuel, hence rename. The packet sends the actual smeltery fuel state.

KnightMiner: The packet renaming is in reference to the one from `consumeFuel`, as now the packet refers to anything using TileTankFuelHeatingStructure instead of just smelteries.

bonii-xx: Ah, yes. Yeah sure, just keep it as clean as possible.

KnightMiner:
I've updated the PR with the specified changes. Specifically, TileHeatingStructureFuelTank now exists for common code between TileSearedFurance and TileSmeltery, slot count was greatly boosted, and slabs are supported on the ceiling.

----

I have three points that I specifically would like you to double check with an opinion. The first is is `BlockEnumSmeltery` where I replaced the check for `TileSmeltery` with `IMasterLogic`, which may need to be moved to an interface that extends `IMasterLogic` if we don't want seared blocks rebuilding non-seared structures (assuming any are planned)

The second is in the structure checking code, which I added an Enum to `isFrameBlock` stating whether it is on the wall, ceiling, or floor. It seemed like the cleanest way, as the alternative was 3 new functions that would hardly be used.

----

I have not quite finished the heat code, as right before bed I noticed a couple major flaws with the new one, namely it causes the stack size to max to 16 for smelting (or 20 for foods), and at a certain point more items is less temp, so ignore that for now...

KnightMiner:
And the heat code is finished. The resulting time is a tiny bit less than a vanilla furnace for normal items, and is reduced for foods (about 2 seconds less). Increasing the stack size increases the temp in linear fassion, causing a significant boost in speed at a full stack compared to a vanilla furnace.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileHeatingStructureFuelTank.java
```diff
@@ -0,0 +1,275 @@
...
+    TileEntity te = worldObj.getTileEntity(pos);
+    if(te instanceof TileTank) {
+      return ((TileTank) te).getInternalTank();
+    }
+
+    return null;
+  }
+
+  /* GUI */
+  public float getMeltingProgress(int index) {
```
bonii-xx: getHeatingProgress

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileHeatingStructureFuelTank.java
```diff
@@ -0,0 +1,275 @@
...
+        currentTank = pos;
+        return;
+      }
+    }
+
+    currentTank = null;
+  }
+
+  // checks if the given location has a fluid tank that contains fuel
+  private boolean hasFuel(BlockPos pos, FluidStack preference) {
```
bonii-xx: hasTankWithFuel

src/main/java/slimeknights/tconstruct/smeltery/client/GuiSearedFurnace.java
```diff
@@ -0,0 +1,130 @@
...
+
+    // furnace size changed
+    if(furnace.getSizeInventory() != sideinventory.inventorySlots.inventorySlots.size()) {
+      // close screen
+      this.mc.thePlayer.closeScreen();
+    }
+  }
+
+  @Override
+  protected void drawGuiContainerForegroundLayer(int mouseX, int mouseY) {
```
bonii-xx: Lots of duplicate code in here again, see if you can pull stuff in a common class for smeltery and furnace

bonii-xx:
Good good, this is looking really good. Only a few things left now, but otherwise i love that we got the smeltery down to 500 LoC. The furnace also feels much better now. The orange "done" bar is a nice touch btw, I like it.

One thing that bugs me is the large furnace fire graphics in the GUI. When I look at it I see a SNES emulator with one of those terrible filters applied to it. [http://bin.smwcentral.net/u/18802/graphic%2Bsnes.PNG]

Maybe we can come up with something more creative, hm. But otherwise it should be fine for now if we fix we make the color transitions.

KnightMiner:
I've made the renames requested and did what I could to remove duplicate code in the GUIs. The two main things I moved were the fuel tank code (which ended up in a superclass), and the tooltip code (which I placed in RenderUtil as I want it in the Tinkers tank GUI). I also did a couple of bug fixes, including stopping the item tooltip from appearing with the melting progress tooltip.

As for in game documentation, I think I might write up a chapter in in the book on multiblocks, which can include the smeltery, the seared furnace, and any future ones. (most likely would be easier to have a separate PR for that though)

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java.rej
```diff
@@ -0,0 +1,16 @@
+diff a/src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java b/src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java	(rejected hunks)
```
bonii-xx: Whoopsie? ;)

bonii-xx:
Looking good, all the refactorings. Gonna do a small formatting and annotation cleanup, whatever matches, and remove the diff file.

bonii-xx:
Just noticed: your IDE doesn't automatically strip empty lines of whitespaces

KnightMiner:
Well, it could if I told it to. I just had it disabled at first since quite a few files I was just modifying had extra whitespace.

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java.rej
KnightMiner: Yeah, something was a little odd which unintended formatting, so I just rejected a couple hunks from one file. Apparently it left a note behind saying what I did...

PR #2261, opened 2016-07-23
Fix alubrass casting

laz2727:


bonii-xx:
So that's why, huh. Curious.

PR #2303, opened 2016-08-06
Allow anything in the top and bottom borders of a seared furnace, decrease stack size and increase smelting time

KnightMiner:
Smelting time now more properly matches vanilla (though is about 4/5 of the speed just due to TileHeatingStructure logic), and stack size was to prevent temperature limits from being reached.

This does remove the stack size boost (meaning 16 items take 16 times the amount of time), though that can easily be added back if requested.


bonii-xx:
Looks good. Linear scaling like vanilla is perfectly fine. The main advantage is not needing 20 furnaces, but only 1, after all. Parallelization! ;)

PR #2338, opened 2016-08-21
Added Tinker Tanks, basically a multiblock like the smeltery that stores more liquid

KnightMiner:
Picture

Capacity is 4 buckets per block in the tank, more details on the structure are in the manual. This PR refactors quite a few of the smeltery things to use an interface (ISmelteryTankHandler), and some of the liquid rendering was tweaked as I discovered a bug with large amounts of liquid displaying when not on the bottom.

Worth noting I don't currently have any particles for the controller as flames did not make sense. If we wanted we could add bubbles I guess?


laz2727:
Tinkers blast furnace when


Full support tho

resources/assets/tconstruct/book/en_US/smeltery/structure.json
```diff
@@ -2,10 +2,10 @@
   "title":"",
   "text": [
     {
-      "text": "The smeltery can be any size from 3x3 to 11x11, with larger structures increasing both the number of items that can smelt and the liquid capacity. The structure requires a floor, but does not require a ceiling or a frame."
```
laz2727: Last time i checked, 3x3x2 smelteries were valid.

KnightMiner: Whoops, typo from copy paste. I added that line to the seared furnace and tinker tank but forgot to make it 2 for the smeltery...

KnightMiner:
Four buckets per block total, not per internal block. A 3x3x3 tank (minimum) does 108 buckets

laz2727:
That's still a bit too little space and too much effort. 8 would be much
better and still not op.

KnightMiner:
Well, you can hold as many different types of liquids as you want in that space, and all are a connected for access, but yeah, balance may not be final.

At its current point, it's the same as a 3x3's cube of tanks in the area

bookerthegeek:
Just from a balance point of view I do believe the base tank standard is 16 per block. 

Some  count the internal space only and some count the wall. Some just say screw it like the mek's tanks.

Personally I would suggest half ofull Railcraft iron tank capacity per block.

laz2727:
Which is 8 buckets.

asl97:
An 11*11 tank could hold a total of 5324 bucket even if it's just four buckets per block total.

Considering that the block require to build are the same as the smeltery which are pretty cheap and it can also hold as many different types of liquids as you want, it's already pretty op.

laz2727:
A sky-high bc tank would hold 4096 and it's even cheaper.

asl97:
Afaik, a sky-high bc tank can only hold one liquid.

laz2727:
Well, it can be divided. What I mean is, space generally isn't a problem,
and the whole concept of fluid storage being op is laughable. Generally,
fluid storage is actually underpowered. Especially multiblock tanks.
Remember last time you used rc tanks? Me neither.

bookerthegeek:
... currently because I like how they look I use rc tanks in my base for Steam and mob essence. But otherwise it is most buckets in smallest space.

How would this interact with a Comparator?

bonii-xx:
@KnightMiner Regarding yesterday, I realized to day that the liquid display stuff on the controller should just be a flat "draw one liquid quad" + an separate overlay. that way texturepacks can also change the shape of it.

bonii-xx:
- Controller is allowed on the bottommost row (good), but not on the top row?
- Potentially allow controller on the corners of the frame too?
- Allow slabs and stairs for the top row, like the smeltery
- Would it make sense to force seared glass for the walls? (drains and controller would be allowed too ofc)
- Have capacity use the automagic text code that turns mb into buckets/kb/... since we have at least 108b capacity anyway

KnightMiner:
I'm pretty sure the controller locations is a limitation of the multiblock cube code, I can look into allowing it but it might not be easy.

Forcing seared glass in the walls could work, it does look a lot better and helps set it apart from the furnace that forces bricks. My only concern is how much the drains stand out when placed in the middle of glass, but that can be worked with

laz2727:
If you want to force seared glass, you pretty much have to buff them.
Seared glass is much harder to make.

bonii-xx:
How about: no.

laz2727:
Well, it's your mod. Do your thing so people can ignore it.

bonii-xx:
@KnightMiner If it's a technical limitation because we never wanted it.. then it's ok. Controller in the middle makes more sense anyway, but it would've been a nice touch. I personally always make the outer frame "pillars" from opaque materials. ;)

Regarding glass.. yeah.. well people can always place a row of drains or place them in the bottom/top layer. I guess if we'll just allow it all and tell people to use seared glass.. but that they can use other blocks too. Less limits to creativity is good there, and usually you want to see what's inside anyway.

KnightMiner:
tinker_tank

I've updated the PR with the stated improvements and a few others specifically:

* Slabs and stairs now work in the ceiling of the tank
* The bottom most liquid is shown in the window rather than water, in addition the texture is tweaked
* Capacity display in the GUI now shows in buckets and kilobuckets in addition to mb
* Seared glass can now be made in the casting basin with seared stone on glass. Just to make it easier to craft for the structure
* Fixed an odd issue with the smeltery disabling of item culling bleeding to other TESRs. This does cause the liquids in the smeltery to no longer show their backfaces (as the bleeding caused them to show it in the first place) so if that was intended I can add a couple lines to bring it back.

laz2727:
This is fine.

>

bonii-xx:
>Fixed an odd issue with the smeltery disabling of item culling bleeding to other TESRs. This does cause the liquids in the smeltery to no longer show their backfaces (as the bleeding caused them to show it in the first place) so if that was intended I can add a couple lines to bring it back.

This sounds weird. Will need to discuss the details on that, in general we don't want the backfaces to be culled. If it affects other rendering then the rendering code is fucked up. This is why we use models wherever possible >_>

>This is fine.

What makes you think that you get any say in this?

KnightMiner:
Well, for more details on that, when I extracted the fluid rendering code from the smeltery renderer into a common class, the Tinker Tank would only show backfaces if a smeltery was visiable.

Adding back the backface rendering in the liquid display is easy enough, with two main effects:

* Firstly, if you are inside the smeltery/tank you see the inside out cubes of fluids rather than the walls.
* Secondly, semitransparent liquids look a lot worse as the parts of the liquid in the back often get rendered after the ones in the front (meaning they show in the front). When there are no backfaces the small count of faces makes thisbless noticeable. Note that seeing through a semitransparent liquid through the one above is still an issue in either case that I cannot find a good fix for

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java
```diff
@@ -173,7 +175,16 @@ public static void renderStackedFluidCuboid(FluidStack fluid, double px, double
     }
 
     int xd = to.getX() - from.getX();
-    int yd = (int) (ymax - ymin);
+
+    // the liquid can stretch over more blocks than the subtracted height is if ymin's decimal is bigger than ymax's decimal (causing UV over 1)
```
bonii-xx: Hence the correction in the quad below. The way it's done is to ensure correct calculation of UVs over block transitions. This will most likely only break things. Got a screenshot of what was wrong before?

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java.rej
```diff
@@ -1,16 +0,0 @@
-diff a/src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java b/src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java	(rejected hunks)
```
bonii-xx: Git artifact

src/main/java/slimeknights/tconstruct/library/fluid/FluidHandlerExtractOnlyWrapper.java
```diff
@@ -9,19 +11,28 @@
...
+    return parent.get() != null;
   }
 
   @Override
   public IFluidTankProperties[] getTankProperties() {
-    IFluidTankProperties[] iFluidTankPropertiesArray = parent.getTankProperties();
-    if(iFluidTankPropertiesArray.length > 0) {
-      IFluidTankProperties fluidTankProperties = parent.getTankProperties()[0];
-      return new IFluidTankProperties[]{new FluidTankProperties(fluidTankProperties.getContents(), fluidTankProperties.getCapacity(), true, false)};
+    if(parent.get() != null) {
```
bonii-xx: hasParent

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileDrain.java
```diff
@@ -19,12 +21,13 @@
...
 
   private FluidHandlerExtractOnlyWrapper drainFluidHandler;
-  private WeakReference<TileEntity> oldSmeltery;
+  private WeakReference<TileEntity> oldSmelteryTank;
 
   @Override
   public boolean hasCapability(@Nonnull Capability<?> capability, @Nullable EnumFacing facing) {
     if(capability == CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY) {
-      return getSmeltery() != null;
+      TileEntity te = getMaster();
```
bonii-xx: getSmelteryTankHandler

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileDrain.java
```diff
@@ -34,15 +37,25 @@ public boolean hasCapability(@Nonnull Capability<?> capability, @Nullable EnumFa
   @Override
   public <T> T getCapability(@Nonnull Capability<T> capability, @Nullable EnumFacing facing) {
     if(capability == CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY) {
-      TileSmeltery smeltery = getSmeltery();
+      TileEntity te = getMaster();
+      if(te == null || !(te instanceof ISmelteryTankHandler)) {
```
bonii-xx: getSmelteryTankHandler

src/main/java/slimeknights/tconstruct/smeltery/client/GuiTinkerTank.java
```diff
@@ -0,0 +1,154 @@
...
+    // subtract the corner of the GUI location so the mouse location is relative to just the center, rather than the inventory center
+    mouseX -= guiLeft;
+    mouseY -= guiTop;
+
+    // Liquids
+    if(8 <= mouseX && mouseX < 114 && 16 <= mouseY && mouseY < 122) {
+      FluidStack hovered = getFluidHovered(122 - mouseY - 1);
+      List<String> text = Lists.newArrayList();
+
+      if(hovered == null) {
```
bonii-xx: Duplicate code

src/main/java/slimeknights/tconstruct/smeltery/client/GuiTinkerTank.java
```diff
@@ -0,0 +1,154 @@
...
+    }
+  }
+
+  @Override
+  protected void drawGuiContainerBackgroundLayer(float partialTicks, int mouseX, int mouseY) {
+    drawBackground(BACKGROUND);
+    drawContainerName();
+
+    // draw liquids
+    SmelteryTank liquids = tinkerTank.getTank();
```
bonii-xx: Duplicate code

src/main/java/slimeknights/tconstruct/smeltery/client/GuiTinkerTank.java
```diff
@@ -0,0 +1,154 @@
...
+  protected void mouseClicked(int mouseX, int mouseY, int mouseButton) throws IOException {
+    if(mouseButton == 0) {
+      mouseX -= guiLeft;
+      mouseY -= guiTop;
+      if(8 <= mouseX && mouseX < 114 && 16 <= mouseY && mouseY < 122) {
+        SmelteryTank tank = tinkerTank.getTank();
+        int[] heights = calcLiquidHeights(tank.getFluids(), tank.getCapacity());
+        int y = 122 - mouseY - 1;
+
+        for(int i = 0; i < heights.length; i++) {
```
bonii-xx: Duplicate code

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java
```diff
@@ -446,12 +459,22 @@ public static int blue(int c) {
...
+  }
+
+  /**
+   * Adds information for the tooltip based on the fluid stacks size.
+   *
+   * @param fluid  Input fluid, only used for type calculation
+   * @param amount Fluid amount
+   * @param text   Text to add information to.
+   */
+  public static void liquidToString(@Nullable Fluid fluid, int amount, List<String> text) {
```
bonii-xx: Wrong kind of extraction. You never want to pass null into this function. Extract the part you're actually interested in to be reused.

src/main/java/slimeknights/tconstruct/smeltery/client/TinkerTankRenderer.java
```diff
@@ -0,0 +1,92 @@
...
+      int brightness, color;
+      if(fluidStack != null) {
+        Fluid fluid = fluidStack.getFluid();
+        sprite = mc.getTextureMapBlocks().getTextureExtry(fluid.getStill().toString());
+        brightness = world.getCombinedLight(pos.offset(face), fluid.getLuminosity(fluidStack));
+        color = fluid.getColor(fluidStack);
+      }
+      else {
+        // if the fluid is missing (aka empty tank) use the default texture
+        sprite = mc.getTextureMapBlocks().getTextureExtry(TEXTURE_TINKER_TANK_BACKGROUND);
```
bonii-xx: Why are you rendering something at all if there is nothing to render? You're rendering exactly the same texture that already is there from the worldrenderer again?

src/main/java/slimeknights/tconstruct/smeltery/client/GuiTinkerTank.java
```diff
@@ -0,0 +1,154 @@
...
+import slimeknights.mantle.client.gui.GuiElement;
+import slimeknights.mantle.inventory.BaseContainer;
+import slimeknights.tconstruct.common.TinkerNetwork;
+import slimeknights.tconstruct.library.Util;
+import slimeknights.tconstruct.library.client.RenderUtil;
+import slimeknights.tconstruct.library.smeltery.SmelteryTank;
+import slimeknights.tconstruct.smeltery.network.SmelteryFluidClicked;
+import slimeknights.tconstruct.smeltery.tileentity.TileTinkerTank;
+
+public class GuiTinkerTank extends GuiContainer {
```
bonii-xx: This class basically was what you were to exctract for the gui.

bonii-xx:
>when I extracted the fluid rendering code from the smeltery renderer into a common class, the Tinker Tank would only show backfaces if a smeltery was visiable.

That means something broke when extracting it, or the Tinker Tank is missing something for rendering. You're fixing the symptoms, not the cause.

src/main/java/slimeknights/tconstruct/smeltery/client/GuiTinkerTank.java
KnightMiner: It was a little hard to move to a superclass since the smeltery uses on GuiMultiModule (which requires an inventory).

I could dump this stuff in render utils.

src/main/java/slimeknights/tconstruct/smeltery/client/TinkerTankRenderer.java
KnightMiner: To make things easier, the controller actually has a hole on the model and I just place the liquid behind it like tanks do.

If I don't have a fluid, I draw the back texture so we don't have an Xray

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java.rej
KnightMiner: Yep, accidently committed in the seared furnace commit. I was just deleting it here.

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java
KnightMiner: I will get a screenshot as soon as I can, but basicit the if on the top quad gets messed up.

If you see my comment below, before this fix with min of 2.3 and max of 5.1 would get a `yd` of 2 (floored from 2.8, set is {2.3, 3, 4, 5.1}).

After the fix, you get a `yd` of 3 (floored from 3.1, set is {2.3, 3, 4, 5, 5.1})

KnightMiner:
Not inside out, it's both inside out and normal are drawn. The issue with the bleeding you would find in the old code. See [here]https://github.com/SlimeKnights/TinkersConstruct/blob/master/src/main/java/slimeknights/tconstruct/smeltery/client/SmelteryRenderer.java#L81). As you notice, it disabled cull (note it's done after the fluids are drawn, so if this was right they should be unaffected) and never enables it again. 

The thing I mentioned about faces in the wrong order exists before my code due to how fluids are drawn per block space, try adding a lot of water to a smeltery (or I'll add screenshots soon)

src/main/java/slimeknights/tconstruct/library/client/RenderUtil.java
KnightMiner: Fixed in latest commit

src/main/java/slimeknights/tconstruct/smeltery/client/TinkerTankRenderer.java
bonii-xx: Just have the background be there always. Otherwise you'll have an xray with semitransparent fluids too. Also it's much easier and less error prone.

bonii-xx:
Just did some tests, and yeah the rendering is broken in the main branch. Must've been changes that happened between 1.8 and 1.10. They seem fixed in your branch, good job. I'll take a more detailed look at the rendering changes when i find more time, for now let's just keep them as long as they fix things.

---

Yes, just make it ticking. It's only one TE and it doesn't do anything special + it should be negligible if it's a full structure since you can just do nothing then. For the GUI stuff you probably will have to extract it somewhere.. doesn't have to be a parent class, they're just helper functions after all.

src/main/java/slimeknights/tconstruct/smeltery/client/TinkerTankRenderer.java
KnightMiner: I could throw it in the model I guess.

KnightMiner:
PR updated, it now includes the GUI for the tanks in a common class, the controller is ticking to check the structure, and the controller back face is done using a model

resources/assets/tconstruct/book/en_US/tinker_tank/structure.json
```diff
@@ -0,0 +1,12 @@
+{
+  "title":"",
+  "text": [
+    {
+      "text": "The tinker tank can be any size from 3x3 to 11x11, and with any height from a minimum of 3. Larger structues increase the size of the tank. In order to be a valid structure, it requires a floor and ceiling, along with a frame around the whole structure"
+    },
+    {
+      "text": "The tinker tank can be built using any combination of seared blocks, glass, tanks, and drains, though the floor and ceiling frome must be either seared blocks or drains. The ceiling additionally supports slabs, even in the frame, provided they are downwards facing.",
```
bonii-xx: typo: frome

resources/assets/tconstruct/models/block/tinker_tank_active.json
```diff
@@ -0,0 +1,35 @@
...
+        "side": "tconstruct:blocks/smeltery/seared_brick",
+        "back": "tconstruct:blocks/smeltery/tinker_tank_background"
+    },
+    "elements": [
+        {
+            "__comment": "main",
+            "from": [ 0, 0, 0 ], 
+            "to": [ 16, 16, 16 ], 
+            "faces": {
+                "north": { "texture": "#front", "uv": [ 0, 0, 16, 16 ], "cullface": "north" },
```
bonii-xx: Why does this have 5 faces? It should only need one?

resources/assets/tconstruct/models/block/tinker_tank_active.json
```diff
@@ -0,0 +1,35 @@
...
+                "up":    { "texture": "#side",  "uv": [ 0, 0, 16, 16 ], "cullface": "up"    },
+                "down":  { "texture": "#side",  "uv": [ 0, 0, 16, 16 ], "cullface": "down"  }
+            }
+        },
+        {
+            "__comment": "backface",
+            "from": [ 0, 0, 1 ], 
+            "to": [ 16, 16, 0 ], 
+            "faces": {
+                "north": { "texture": "#back", "uv": [ 0, 0, 16, 16 ], "cullface": "north" },
```
bonii-xx: Why does this have 5 faces? It should only need one

KnightMiner: It's back about 1 pixek, so you can see a bit of the empty edge from the right angle

resources/assets/tconstruct/book/en_US/tinker_tank/structure.json
bonii-xx: Also needs some description where the tank controller can go

resources/assets/tconstruct/models/block/tinker_tank_active.json
bonii-xx: Don't worry I got dis. Also you're DC'd in irc :(

bonii-xx:
I did some tweaking for the tank rendering. Pulled out the rendered area to the corner of the tank (minus one pixel) and made the active tank to only use one face. Also I made the active tank render water if no fluid is present so we have something to distinguish it from non-working tanks. Also made the fluid in the tank not render to the top because I was toying around with that and i kinda liked how it looked ;)

Patch can be found here: https://gist.github.com/bonii-xx/7361001b08547a8b381df598861c2fbe

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileTinkerTank.java
```diff
@@ -0,0 +1,297 @@
...
+  }
+
+  @Override
+  public Container createContainer(InventoryPlayer inventoryplayer, World world, BlockPos pos) {
+    return new ContainerTinkerTank(this);
+  }
+
+  @Override
+  public GuiContainer createGui(InventoryPlayer inventoryplayer, World world, BlockPos pos) {
+    return new GuiTinkerTank((ContainerTinkerTank) createContainer(inventoryplayer, world, pos), this);
```
bonii-xx: Unneeded cast

KnightMiner:
More updates. I ended up tweaking the patch you gave me slightly, mainly you forgot to update the TE render area size, so that is why I moved the position stuff directly to the TE.

Also includes things from the comments other than the patch

bonii-xx:
It is done! There still is some stuff we should refactor to avoid bugs happening in multiple places, but I want that done separately to keep the changesets smaller.

PR #2339, opened 2016-08-21
Web & Chulking

Yulife:


bonii-xx:
Took a look. Web needs a few adaptions, shulking looks very good.

- The web on the lumber axe is on the wrong side
- The one on the broadsword should be consistent with the other swords, if we're doing it like that
- Frying pan web makes no sense, should probably span to the handle
- Shulking on the excavator is missing
- Shulking is missing a model file (models/modifiers)

Yulife:
I'll definitely get around to them, however, IRL is pretty busy right now so it might be delayed a bit :)

Issue #2315, opened 2016-08-13
Compatibility with structure blocks


Lythom:
Hi,

Using minecraft 1.10.2 and TConstruct 1.10-2.3.3a, I got the game to crash when spawning a smeltery using structure blocks.

steps to reproduce :

1. Place a structure block named "smeltery" in save mode, configured with -1 1 -1 as relative position, and 3 3 3 as size.
2. build a minimal 3x3x2 smeltry on top of it
3. Use the save button on the structure block
4. Place another structureblock elsewhere in "load" mode named "smeltery". Click "load".

When the structure appears, on first next tick, the game crash and the following stacktrace appears in logs : http://pastebin.com/sQWTupyj. On reboot, the game doesn't crash anymore but the smeltery is "broken" (nothing works).

Reasons why you might consider compatibility with structure block :

- It's a vanilla mechanic builders and map makers might use to build their worlds / maps.
- It grants automatic compatibility with the 1.10.2+ versions of the [Capsule mod](http://imgur.com/gallery/xCWCX) (which relies on the structure block system).

Additional informations :

- I developped some tools to inspect tiles entities (it print a "give" command that would spawn a conform copy). I got a comparison between the initial and loaded smeltry : http://pastebin.com/raw/hQaZtPeQ. I highlight that most but not all coordinates are updated correctly after move : all {x,y,z,minPos,maxPos} looks ok, but all {xCenter, yCenter, zCenter} are wrong.
- With some instructions on how the problem should be solved (use relative coordinate ? recalculate smeltry integrity on load ?) I can come with some PR. I don't know much the codebase so any info that could help me take that try is welcome !

bonii-xx:
Does it also happen with https://minecraft.curseforge.com/projects/tinkers-construct/files/2319366 ?

Lythom:
Yes, this is reproductible as well with TConstruct-1.10.2-2.4.0.jenkins312.jar (+ Mantle-1.10.2-0.10.4.jar)

bonii-xx:
Looks like the parts still detect the original smeltery as valid (I assume it still exists?) and hence can't be reassigned to the new smeltery. Usually the smeltery will just rebuild its data if it's not valid anymore.

bonii-xx:
[...]

PR #2342, opened 2016-08-23
Basic structure block support for smeltery (#2315)

Lythom:
When the smeltery is copied using a structure block, the state is broken which can be identified by the smeltery being "active" while the variable this.info is null. This, by the way, causes the server the crash during the update.

This PR add a check to prevent the update if the state is identified as broken, and try to rebuild the smeltery if possible. The rebuild will work if the former smeltery is gone. If the former smeltery still exists, new loaded smeltery will not be valid and skip the tick updates, but at least won't crash the server anymore.

This PR completely allows smelteries to be moved or spawned as reward with capsules as well (see #2315), since the original smeltery will never exist anymore when the content is placed in the world.


bonii-xx:
Fixed within fb059f91ea34e9ac705946ae942959bfd7b8fde3

PR #2362, opened 2016-09-03
Fixed a crash caused by some other mods' recipes

The-Fireplace:
Fixed the following crash:
[http://paste.ubuntu.com/23129570/](http://paste.ubuntu.com/23129570/)
And potentially others.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
+++ b/src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
@@ -821,9 +821,13 @@ public class TinkerSmeltery extends TinkerPulse {
         ItemStack output = irecipe.getRecipeOutput().copy();
         int amount = known.get(fluid) / output.stackSize;
         output.stackSize = 1;
-        TinkerRegistry.registerMelting(new MeltingRecipe(RecipeMatch.of(output, amount), fluid));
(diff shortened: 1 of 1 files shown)


bonii-xx:
Wrong way to fix it. Should not require a catch. Should check that the log doesn't crash beforehand and then still log something useful. Log registry name additionally to the display name, and "null" as display name in case of null.

But then again it's questionable whether we should actually fix it, since it's a bug on the other mods side that should be fixed anyway.

The-Fireplace:
You're right, it is questionable if this should be fixed here or in Fluidity. On one hand, this mod's code is crashing, and this fix works. On the other, I should probably also figure out why something is null in one of my recipes, despite the fact that everything crafts without issues.
I am working right now to get it to provide useful information about the issue so I can pinpoint what exactly is triggering it. I will update the PR once I get it to provide useful information.

KnightMiner:
I would have Fluidity fix it. Having a null output for a recipe is going to break something eventually, be it Tinkers Construct or another mod, simply because it should not happen. Fixing it in Tinkers is just going to delay the problem to some other mod that expects an output.

Though, why is it crashing on ItemStack::toString()? We run ItemStack::copy() on the output above and it does not NPE there, so does the output of the recipe change since we asked the first time?

The-Fireplace:
Never mind, it is a Fluidity bug. Sorry.

PR #2373, opened 2016-09-08
Fix Tinker Tank bounding box not including the full liquid and the lighting location being offset in the wrong direction

KnightMiner:
Fixes issues caused by the multiblock code extraction as the stored position was changed from the liquid start to the inside start.

Also fixes the liquid lighting location as it was offset in the opposite direction during 9466eba

### DIFF
--- a/src/main/java/slimeknights/tconstruct/smeltery/client/TinkerTankRenderer.java
+++ b/src/main/java/slimeknights/tconstruct/smeltery/client/TinkerTankRenderer.java
@@ -90,7 +90,7 @@ public class TinkerTankRenderer extends SmelteryTankRenderer<TileTinkerTank> {
     }
 
     // draw the fluids inside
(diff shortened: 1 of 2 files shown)


src/main/java/slimeknights/tconstruct/smeltery/client/TinkerTankRenderer.java
```diff
@@ -90,7 +90,7 @@ public void renderTileEntityAt(@Nonnull TileTinkerTank tinkerTank, double x, dou
     }
 
     // draw the fluids inside
-    // we offset the minPos for lighting since its possible to have a solid block (always light 0) where the liquid starts
-    renderFluids(tank, tilePos, minPos.add(-1, 0, -1), maxPos.add(1, 0, 1), x, y, z, 0.0625f, minPos.add(2, 0, 2));
```
bonii-xx: Oh yeah right. We probably want a position inside the tank for the lighting

PR #2375, opened 2016-09-11
Add scythes

KnightMiner:
Scythes are back. Like before, they have an AOE attack and AOE plant breaking. They also now have a new right click ability which will harvest crops and replant them, plus perform an AOE attack centered on the player.

The right click move does come with a cooldown to prevent spam, and upon using will also reset the attack meter for the left click action.


bonii-xx:
Hm. Now we have it so that you can't replant properly at all. Also rightclick seems to create path? And the AOE seems to be the sword AOE that only deals 1 damage?

What does the rightclick do right now? I thought we changed that to rightclick harvest and replant?

KnightMiner:
Right click will do an AOE attack around the player,  which also breaks or replant crops (based on if they drop a seed). No idea why it would create path, I added no such code.

I will double check the AOE to see what the damage is, I know the right click was lower but for left click I might have accidently called the wrong function.

KnightMiner:
After a bit of testing, I am getting inconsistent results, as in from what I can tell a random entity within range is getting the full damage and the rest limited, though if you could link me a deobf build of TestDummy on IRC that would help a lot with testing (the one I have is 1.9)

TherminatorX:
That's how vanilla swords do the sweep attack, the one you target gets full damage, the rest 1 damage.

LovelyFlare:
@TherminatorX Get on skype -.-

KnightMiner:
Yes, we know. That is not how scythes are supposed to work though, they do equal damage to all targets

And if you read what I said, I said random. Sometimes the one I targeted got minimal damage and another got full

bonii-xx:
Before I forget it.. we totally should also add shearing to rightclick. We should probably have a short talk about the rightclick stuff.

KnightMiner:
I've updated the PR, both changing the right click behavior to block location and replacing the right click "spin attack" with a AOE shearing.

I also figured out why the AOE attack was dealing such low damage, and why it seemed to choose a random target for the full damage. Apparently the attack cooldown was getting reset between entities causing one entity to get full damage and the rest to get uncharged damage.

bonii-xx:
What a fun thing to happen. Do you know what triggers the cooldown reset?

KnightMiner:
[Apparently we do](https://github.com/SlimeKnights/TinkersConstruct/blob/master/src/main/java/slimeknights/tconstruct/library/utils/ToolHelper.java#L739), maybe adding a boolean to disable that would have been a better work around...

KnightMiner:
Well, I switched to a boolean there and it works fine, though I am not entirely sure why we are calling the resetCooldown() method in the ToolHelper attack function as it seems vanilla applies it later anyways (unless that was for alternative attack types).

bonii-xx:
Urgh, can you check if the rapier is also affected by that?

bonii-xx:
Looks like the offset on rightclick is sometimes wrong and doesn't match the displayed AOE blocks:

resources/assets/tconstruct/models/item/tools/scythe.tcon.json
```diff
@@ -0,0 +1,9 @@
+{
+    "textures": {
+        "layer0": "tconstruct:items/scythe/handle",
+        "layer1": "tconstruct:items/scythe/head",
+        "layer2": "tconstruct:items/scythe/binding",
+        "layer3": "tconstruct:items/scythe/accessory",
+		"broken1": "tconstruct:items/scythe/broken_head"
```
bonii-xx: Ident broken

src/main/java/slimeknights/tconstruct/tools/tools/Scythe.java
```diff
@@ -1,158 +1,306 @@
...
+
+    if(shorn) {
+      player.worldObj.playSound(null, player.posX, player.posY, player.posZ, SoundEvents.ENTITY_PLAYER_ATTACK_SWEEP, player.getSoundCategory(), 1.0F, 1.0F);
+      player.spawnSweepParticles();
+    }
+
+    return shorn;
+  }
+
+  public boolean harvestBlock(ItemStack stack, World world, EntityPlayer player, BlockPos pos, int fortune) {
```
bonii-xx: Best rename to something more appropriate so it's not confused with the regular leftclick blockbreaking (whose functions are all called harvest)

src/main/java/slimeknights/tconstruct/tools/tools/Scythe.java
```diff
@@ -1,158 +1,306 @@
...
   public ToolNBT buildTagData(List<Material> materials) {
-    HandleMaterialStats handle = materials.get(0).getStatsOrUnknown(MaterialTypes.HANDLE);
-    HeadMaterialStats head = materials.get(2).getStatsOrUnknown(MaterialTypes.HEAD);
-    HandleMaterialStats handle2 = materials.get(1).getStatsOrUnknown(MaterialTypes.HANDLE);
-    ExtraMaterialStats binding = materials.get(3).getStatsOrUnknown(MaterialTypes.EXTRA);
-
-    ToolNBT data = new ToolNBT();
-
-    return data;
+    return buildDefaultTag(materials);
```
bonii-xx: Still needs proper building math and bonus durability for being a big tool.

src/main/java/slimeknights/tconstruct/tools/tools/Scythe.java
```diff
@@ -1,158 +1,306 @@
...
+import slimeknights.tconstruct.library.tools.AoeToolCore;
+import slimeknights.tconstruct.library.tools.IAoeTool;
 import slimeknights.tconstruct.library.tools.ToolNBT;
 import slimeknights.tconstruct.library.utils.ToolHelper;
 import slimeknights.tconstruct.tools.TinkerTools;
 
-public class Scythe extends TinkerToolCore {
+public class Scythe extends AoeToolCore {
+
+  public static final String HARVEST_TAG = "harvesting";
```
bonii-xx: Unused constant

src/main/java/slimeknights/tconstruct/tools/tools/Scythe.java
```diff
@@ -1,158 +1,306 @@
...
-  }
 
-  protected void breakBlock(ItemStack stack, BlockPos pos, EntityPlayer player) {
-    // silktouch gives us shears :D
-    if(EnchantmentHelper.getEnchantmentLevel(Enchantments.SILK_TOUCH, stack) > 0) {
-      if(ToolHelper.shearBlock(stack, player.worldObj, player, pos)) {
-        return;
+  @Override
+  public boolean onBlockStartBreak(ItemStack stack, BlockPos pos, EntityPlayer player) {
+    if(!ToolHelper.isBroken(stack) && this instanceof IAoeTool && ((IAoeTool) this).isAoeHarvestTool()) {
```
bonii-xx: Redundant instanceof and casts

src/main/java/slimeknights/tconstruct/library/utils/ToolHelper.java
```diff
@@ -735,7 +738,7 @@ public static boolean attackEntity(ItemStack stack, ToolCore tool, EntityLivingB
         }
 
         // cooldown for non-projectiles
-        if(!isProjectile) {
+        if(!isProjectile && !skipCooldown) {
```
bonii-xx: Rename the parameter (and where it's called) to "startCooldown" or something similar, double negative is bad when thinking through the logic.

resources/assets/tconstruct/models/item/tools/scythe.tcon.json
KnightMiner: Bad indent.

Do you want tabs or spaces in JSON generally?

bonii-xx: I didn't really pay attention when writing them, so whatever's already there.

src/main/java/slimeknights/tconstruct/tools/tools/Scythe.java
KnightMiner: Why do I keep forgetting to remove these? Its leftover from a very bad solution to the harvesting block breaking which I nearly immediately scrapped.

src/main/java/slimeknights/tconstruct/tools/tools/Scythe.java
```diff
@@ -1,158 +1,306 @@
...
+  public boolean shearEntity(ItemStack stack, World world, EntityPlayer player, Entity entity, int distance, int fortune) {
+    if(!(entity instanceof IShearable)) {
+      return false;
+    }
+
+    IShearable shearable = (IShearable)entity;
+    if(shearable.isShearable(stack, world, entity.getPosition())) {
+      List<ItemStack> drops = shearable.onSheared(stack, world, entity.getPosition(), fortune);
+      Random rand = world.rand;
+      for(ItemStack drop : drops) {
```
bonii-xx: Only spawn items serverside, this creates clientside entityitems

src/main/java/slimeknights/tconstruct/tools/tools/Scythe.java
```diff
@@ -1,158 +1,306 @@
...
-        for(int zp = start.getZ() - d; zp != start.getZ() + d; zp++) {
-          if(MathHelper.abs_int(xp - start.getX()) + MathHelper.abs_int(yp - start.getY()) + MathHelper
-              .abs_int(zp - start.getZ()) >= distance) {
-            continue;
+  public ActionResult<ItemStack> onItemRightClick(@Nonnull ItemStack stack, World world, EntityPlayer player, EnumHand hand) {
+    if(ToolHelper.isBroken(stack)) {
+      return ActionResult.newResult(EnumActionResult.FAIL, stack);
+    }
+
+    RayTraceResult trace = this.rayTrace(world, player, false);
```
bonii-xx: This needs to pass true as the last parameter so it stops at crops (non-opaque-solid blocks)

bonii-xx:
Looks good so far. There's a few small details, but it's probably easier and faster if I just fix them quickly myself. The rightclick stuff turned out really good, and yes it's intended that autosmelt drops baked potatoes ;)

KnightMiner:
I figured, it was just unexpected that it did not replay at first before I remembered it now laced seeds :)

PR #2413, opened 2016-09-26
Update zh_CN.lang

3TUSK:
~~Summon @TartaricAcid for double checking. Personally I am confident with this.~~ (Already confirmed, never mind)


bonii-xx:
Great, thanks

PR #2425, opened 2016-09-29
Update fr_FR.lang

Trounk:
So like this ? :)

### DIFF
--- a/resources/assets/tconstruct/lang/fr_FR.lang
+++ b/resources/assets/tconstruct/lang/fr_FR.lang
@@ -940,5 +940,5 @@ option.tconstruct.progress=Progression de la recette
 
 
 # Config
-config.synced.restart=votre configuration a été mise à jour pour correspondre au serveur. Relancez Minecraft.
+config.synced.restart=Votre configuration a été mise à jour pour correspondre au serveur. Relancez Minecraft.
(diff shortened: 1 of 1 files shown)


LovelyFlare:
Yes, like that.

resources/assets/tconstruct/lang/fr_FR.lang
```diff
@@ -862,8 +938,7 @@ option.tconstruct.tank=Réservoir
 option.tconstruct.casting=Moulage
 option.tconstruct.progress=Progression de la recette
 
+
 # Config
-config.synced.restart=Votre configuration a été mise à jour pour correspondre au serveur. Relancez Minecraft.
+config.synced.restart=votre configuration a été mise à jour pour correspondre au serveur. Relancez Minecraft.
```
LovelyFlare: Why is this decapitalized?  This is the message the game gives in chat when the configs sync on server join.

Trounk: Hum hum, its a good question, maybe an error of copy-paste, I check :D

bonii-xx:
Great, thanks! I'll merge it in together with the russian version when it's there. If you find something else amiss, you can fix it that way.

PR #2427, opened 2016-09-30
Create pt_BR.lang

ArukaAlter:


bonii-xx:
Cool, you even translated the stuff for the release after the next release already (don't worry, they're fixed, but some more will be added later). What's up with all the empty lines around line 424?

ArukaAlter:
I wanted to skip that part, and for consistency, it's easier to find where new lines come when the english files updates.

bonii-xx:
Ah you left out the spaghetti completely, didn't see that. kk

PR #2429, opened 2016-10-01
Fix scythes not harvesting the clicked block (just AOE blocks)

KnightMiner:
You forgot to re-add that after you reworked the scythes :P


bonii-xx:
Oh right. I didn't notice because I was testing with EnderCore support and that harvested the clicked block on rightclick... derp.

PR #2481, opened 2016-10-12
Add longbows

KnightMiner:
Longbows are like shortbows, but have slightly more durability (to make up for the higher cost), more base power, and more speed/range. Their disadvantage is longer drawback time and worse accuracy.

----

To expand upon the conversation from IRC, testing this with vanilla arrows feels balanced since they take speed into account for damage, but testing with Tinkers arrows which ignore speed for damage basically makes drawspeed the only important factor for bows as speed/range becomes just about getting enough range to hit your target.


KamranMackey:
LGTM :+1:

Now we just need a response from @bonii-xx :)

JCMais:
@bonii-xx when should we expect (if you can say a date anyway) that version to hit curseforge?

KnightMiner:
When it is done

PR #2508, opened 2016-10-20
Chinese bug is here

subtank:
Line 2,it is a coding line can't translation


bonii-xx:
Ah yes. That is indeed bad. Thanks for fixing it!

liach:
line 11 made me laugh

PR #2524, opened 2016-10-27
ItemBlockTable tooltip changes

Tiax:
- Show actual item count on ItemBlockTable, when it holds an inventory in NBT.
- Show if PatternChest actually holds patterns or casts, in tooltip.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tools/common/item/ItemBlockTable.java
+++ b/src/main/java/slimeknights/tconstruct/tools/common/item/ItemBlockTable.java
@@ -2,8 +2,10 @@ package slimeknights.tconstruct.tools.common.item;
 
 import net.minecraft.block.Block;
 import net.minecraft.entity.player.EntityPlayer;
+import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.nbt.NBTTagList;
 
 import java.util.List;
 
@@ -12,7 +14,10 @@ import javax.annotation.Nonnull;
 import slimeknights.mantle.item.ItemBlockMeta;
 import slimeknights.mantle.util.LocUtils;
 import slimeknights.tconstruct.library.Util;
+import slimeknights.tconstruct.library.smeltery.ICast;
+import slimeknights.tconstruct.library.tools.IPattern;
 import slimeknights.tconstruct.shared.tileentity.TileTable;
+import slimeknights.tconstruct.tools.common.block.BlockToolTable;
 
 public class ItemBlockTable extends ItemBlockMeta {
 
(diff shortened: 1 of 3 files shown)


src/main/java/slimeknights/tconstruct/tools/common/item/ItemBlockTable.java
```diff
@@ -34,7 +39,40 @@ public void addInformation(@Nonnull ItemStack stack, @Nonnull EntityPlayer playe
...
+      }
+
+      NBTTagList items = inventory.getTagList("Items", 10);
+
+      // check if it's a PatternChest (damage value corresponds to TableTypes enum)
+      if (BlockToolTable.TableTypes.fromMeta(stack.getItemDamage()) == BlockToolTable.TableTypes.PatternChest) {
+        // determine if it holds casts or patterns
+        boolean holdsCasts = false;
+
+        for (int i = 0; i < items.tagCount(); ++i) {
```
bonii-xx: The chest can only hold either casts or patterns, no need for a loop, the first item is enough.

src/main/java/slimeknights/tconstruct/tools/common/item/ItemBlockTable.java
```diff
@@ -34,7 +39,40 @@ public void addInformation(@Nonnull ItemStack stack, @Nonnull EntityPlayer playe
...
+          Item item = ItemStack.loadItemStackFromNBT(items.getCompoundTagAt(i)).getItem();
+
+          if (item instanceof ICast || item instanceof IPattern) {
+            holdsCasts = (item instanceof ICast);
+            break;
+          }
+        }
+
+        if (holdsCasts) {
+          tooltip.addAll(LocUtils.getTooltips(Util.translateFormatted("tooltip.patternchest.holds_casts", items.tagCount())));
```
bonii-xx: Extract format string in variable, set string variable instead of holdsCasts

Tiax:
Reworked based on feedback. Still not sure I want to trust the NBT contents blindly - the loop would have ideally only had a single iteration anyway, unless the data is not as expected.

bonii-xx:
Well I didn't say that you should trust it blindly, just that we don't need a loop and to make it more readable.

src/main/java/slimeknights/tconstruct/tools/common/item/ItemBlockTable.java
```diff
@@ -34,7 +39,27 @@ public void addInformation(@Nonnull ItemStack stack, @Nonnull EntityPlayer playe
...
+        return;
+      }
+
+      NBTTagList items = inventory.getTagList("Items", 10);
+
+      // check if it's a PatternChest (damage value corresponds to TableTypes enum)
+      if (BlockToolTable.TableTypes.fromMeta(stack.getItemDamage()) == BlockToolTable.TableTypes.PatternChest) {
+        // determine if it holds casts or patterns
+        // assume the contents are valid, thus checking any stack in the list suffices
+        Item item = ItemStack.loadItemStackFromNBT(items.getCompoundTagAt(0)).getItem();
```
Tiax: in which case we'd need the loop again...

Say: slot 0 -> removed item, slot 1 -> valid cast/pattern

Tiax:
- Added loop again (still with a single iteration, in a perfect world) to find valid ItemStack to base the holds_casts/holds_patterns on
- rebased the changes into a single commit

PR #2525, opened 2016-10-27
Stacking after breaking Tinkers tables/chests, keep names when changed in anvil

Tiax:
- Placed part- and pattern chests now correctly stack with freshly crafted ones (if empty).
- Renamed Tinkers tables/chests now retain any custom names, when broken.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/shared/block/BlockTable.java
+++ b/src/main/java/slimeknights/tconstruct/shared/block/BlockTable.java
@@ -14,6 +14,7 @@ import net.minecraft.entity.player.EntityPlayer;
 import net.minecraft.item.Item;
 import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
+import net.minecraft.nbt.NBTTagString;
 import net.minecraft.tileentity.TileEntity;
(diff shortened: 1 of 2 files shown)


src/main/java/slimeknights/tconstruct/shared/block/BlockTable.java
```diff
@@ -169,18 +175,27 @@ private void writeDataOntoItemstack(@Nonnull ItemStack item, @Nonnull IBlockAcce
...
+      }
 
-      // save inventory?
+      // save inventory, if not empty
       if(noInventorySave && keepInventory(state)) {
-        NBTTagCompound inventoryTag = new NBTTagCompound();
-        table.writeInventoryToNBT(inventoryTag);
-        tag.setTag("inventory", inventoryTag);
-        table.clear();
+        if (!table.isInventoryEmpty()) {
```
bonii-xx: Weird that nobody ever noticed that, heh

Tiax: Guess you don't normally create more TinkerTables than you want to place, so it's usually a stack of 1 anyway :)

PR #2556, opened 2016-11-11
Add right click functionality to tool traits

JamiesWhiteShirt:
This implements right-click functionality for tool traits as specified in the title. The rules for which right-click function is used are simple.
1. If the tool has a built-in right-click function, this is applied.
2. If the built-in right-click function passes (returns EnumActionResult.PASS), the right click functions of traits are applied.
3. In the case of conflicting trait right-click functions, the first that returns != EnumActionResult.PASS is used.

The right click function is only called for the block that is clicked, not the AOE blocks.

[Here's an example trait that uses onToolUse.](https://gist.github.com/JamiesWhiteShirt/9336cc873022f6bc5a87d9e3888a9144)

I'm developing a mod which, among several things, will be adding a TConstruct tool trait. My intention with this tool trait is that it will add a specific right-click function to TConstruct tools. If you are curious why I need it, [this is the mod in question](https://minecraft.curseforge.com/projects/literal-ascension) where the trait will give the right click behaviour of the Carving Tool to any TConstruct tool.

I initially tried implementing this through PlayerInteractEvent.RightClickBlock. Unfortunately the event does not allow me to alter the EnumActionResult of the right-click to SUCCESS. The result is that there is no animation when the tool is used since. The animation will only happen if the EnumActionResult of the right click is SUCCESS.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/tools/ToolCore.java
+++ b/src/main/java/slimeknights/tconstruct/library/tools/ToolCore.java
@@ -18,6 +18,8 @@ import net.minecraft.item.ItemStack;
 import net.minecraft.nbt.NBTTagCompound;
 import net.minecraft.nbt.NBTTagList;
 import net.minecraft.util.DamageSource;
+import net.minecraft.util.EnumActionResult;
+import net.minecraft.util.EnumFacing;
 import net.minecraft.util.EnumHand;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.util.math.RayTraceResult;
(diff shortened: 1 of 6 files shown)


bonii-xx:
I intentionally didn't implement anything like that when I did traits, because that's what tools are for. The trait only modifies behaviour, not add. However two things come to my mind:

- Why are you adding the carving stuff via trait, and not via a carving tool?
- RightClickBlock + event cancelling should work fine, if the only thing you're missing is the animation you can call EntityPlayerSP.swingArm clientside

JamiesWhiteShirt:
I won't go too much into why I'm adding it as a trait. It's already a vanilla-type tool, but I want it as a TConstruct trait as well for those who use TConstruct tools.

It seems your tip pretty much solved my problem, so I don't require this trait functionality for my trait to work properly.

I can see why you intentionally didn't add this to traits. I figure this trait functionality is not something you want, so I'm closing the PR.

bonii-xx:
Looking forward to it :+1:

PR #2619, opened 2016-12-04
Upd Rus translation

Mexahoid:
Finally I've got some time for translations.


bonii-xx:
Thanks! You probably don't need the pages with the smeltery structure btw. If it's not present in your translation it should default to the en_us, so anything that doesn't have actual text in it should be fine.

bonii-xx:
I added a few more lines to the translation today, think you can fix them up quickly?

Mexahoid:
Yes I can.

PR #2659, opened 2016-12-17
Modified the glowing modifier to check for liquid blocks before glowing block placement

### LINKED ISSUE #2535: Glowing modifier underwater
So I went to raid an ocean monument the other day, and I have my glowing battle sign in my offhand. It tried to place the light ball thingy but it keeps getting deleted by the water, and the repeating process cause durability to drain. Is it possible to change it so that the light ball doesn't get deleted by the water or make it so that it doesn't check light level if I'm in water or something?

MightyCupcakes:
With this PR, the glowing modifier will no longer trying to add glowing blocks around the entity if the said entity is in water. Fixes #2535

### DIFF
--- a/src/main/java/slimeknights/tconstruct/shared/block/BlockGlow.java
+++ b/src/main/java/slimeknights/tconstruct/shared/block/BlockGlow.java
@@ -3,6 +3,7 @@ package slimeknights.tconstruct.shared.block;
 import com.google.common.collect.ImmutableMap;
 
 import net.minecraft.block.Block;
+import net.minecraft.block.BlockLiquid;
 import net.minecraft.block.SoundType;
 import net.minecraft.block.material.Material;
 import net.minecraft.block.properties.PropertyDirection;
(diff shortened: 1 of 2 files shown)


bonii-xx:
That does not fix all cases though, the placement logic needs a check whether the block where it's trying to replace is a liquid block. Maybe it needs a check for the isAir replacable thingie callback, but I'm not sure how that behaves with stuff etc.

MightyCupcakes:
True. Will investigate further.

Also I noticed that the glowing modifier checks for the general light level (i.e the one with skylight), causing it not to work in the overworld on blocks that have a sky light level of 15 (outside of caves) even though it is midnight and mobs can certainly spawn. Is this intended behaviour? I looked at how mobs check their light level and replicated it in the modifier and it will not work in the overworld during the night.

If yes, I will reverse my latest changes.

bonii-xx:
No it's not intended, I just never got around to investigate it further

MightyCupcakes:
Played around with it a little and saw no problems. It's ready for review.

PR #2677, opened 2016-12-26
Allow Slime Boots to be crafted in all five slime colors

KnightMiner:
Blue slime boots

Just a small thing I threw together because I want blue slime boots. They use the vanilla color tinting from leather armor, so its all still one item ID (just different metadata as slime boots don't have durability).

Only other things worth mentioning are right now the recipe uses a flat reference to the congealed slime blocks instead of adding color specific congealed slime oredict names (like the items use, I was not sure we would ever see that many congealed slime blocks outside Tinkers), and the old recipe (any slime type) was kept as a fallback recipe for mismatching/different slime types.


bonii-xx:
Ah yes, one of the things I was too lazy to do ;)

PR #2678, opened 2016-12-26
Update pt_BR.lang

ArukaAlter:


resources/assets/tconstruct/lang/pt_BR.lang
```diff
@@ -396,12 +399,18 @@ item.tconstruct.arrow_head.name=Ponta de Flecha
...
 
+item.tconstruct.bolt_core.name=Haste de dardo
+
 # Material specific tools and toolparts
 # item.tconstruct.<tool>.<material>
 # item.tconstruct.pickaxe.wood=Wooden Pickaxe
 
 # Gadgets
 item.tconstruct.slime_boots.name=Botas de Slime
+item.tconstruct.slime_boots.blue.name=item.tconstruct.slime_boots.green.name
```
bonii-xx: You don't have the green boots anywhere though

ArukaAlter:
But the name is the same across them (except for the blood boots), there's no color differentiation by name.

KnightMiner:
Yes, but you never actually define the green string, you define the one without a color (now unused). No color in the name is mainly to match slimeballs anyways.

Though it's worth mentioning slimeboots use vanilla localization rather than Tinkers, so setting it to another name is probably not going to localize it again.

ArukaAlter:
It will be at least 15 hours from now for me to be available to fix this properly.

PR #2687, opened 2016-12-30
Fix tiny typo in EFLN description (save -> safe)

SharkWipf:
Unless you meant they should be saved for mining of course.  
(Felt like overkill to submit an issue for this while it's a 1 letter change)

### DIFF
--- a/resources/assets/tconstruct/lang/en_US.lang
+++ b/resources/assets/tconstruct/lang/en_US.lang
@@ -429,7 +429,7 @@ item.tconstruct.fancy_frame.clear.name=Clear Item Frame
 item.tconstruct.throwball.glow.name=Glowball
 item.tconstruct.throwball.glow.tooltip=Not suited for consumption.
 item.tconstruct.throwball.efln.name=EFLN
-item.tconstruct.throwball.efln.tooltip=Explodes! Save for mining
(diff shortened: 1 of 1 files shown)


SharkWipf:
Didn't realize a PR would go through CI even before approval, for future reference, would you prefer stuff like this as issue or as PR?

C1ffisme:
@SharkWipf On the other hand, you could also interpret this line as "Save this for mining, don't bother using it on mobs." but I don't think that's what the sentence was meant to portray.

SharkWipf:
@C1ffisme especially given how much fun it is to use on mobs, they made sure it is still useable for [Tera-jumping](http://wiki.nosebud.de/wiki/Terajump) too

LovelyFlare:
It's certainly not safe for mining if you're being damaged by it when you're close to it...

SharkWipf:
@LovelyFlare I think they mean "safe" as in "it doesn't destroy drops", as opposed to regular TNT which does destroy a percentage of the drops.

C1ffisme:
@SharkWipf Ah, okay. I haven't tested the EFLN's too much yet.

LovelyFlare:
@SharkWipf That's an extremely bad way of wording it then...

bonii-xx:
PRs are fine, the hook is just there to check if it compiles. Teraflints obviously are the safest thing for mining ever.

PR #2716, opened 2017-01-10
Change the seared furnance to kill all mobs within itself when active

### LINKED ISSUE #2638: Mobs spawning inside Seared furnace
I built a 11x11 sear furnace 5 tall and have noticed that mobs are spawning inside the walls of side furnace (unsure why)

MightyCupcakes:
Mobs killed this way will not drop xp nor items. Fixes #2638 

There are many ways to achieve the same effect but I chose killing mobs after they spawn primarily because of the ease of implementation. Preventing mob spawn requires subscribing to the mob spawning event and there are potentially a lot of similar tile entites which will subscribe to the same event. And also when the tile entity is destroyed, it requires even more handling to make sure the event is not passed to the non-existent tile entity.

Adding glow might also work but it requires additional checks for light level and the subsequent destruction of said glow blocks in the event the structure is invalid.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
+++ b/src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
@@ -152,13 +152,11 @@ public class TileSearedFurnace extends TileHeatingStructureFuelTank<MultiblockSe
   protected void interactWithEntitiesInside() {
       AxisAlignedBB bb = info.getBoundingBox().contract(1).offset(0, 0.5, 0).expand(0, 0.5, 0);
 
-      List<Entity> entities = getWorld().getEntitiesWithinAABB(Entity.class, bb);
+      List<EntityLivingBase> entities = getWorld().getEntitiesWithinAABB(EntityLivingBase.class, bb);
 
-      for(Entity entity : entities) {
+      for(EntityLivingBase entity : entities) {
           if (entity instanceof EntityMob) {
-              EntityLivingBase entityLiving = (EntityLivingBase) entity;
-
-              entityLiving.setHealth(0F);
+              entity.setHealth(0F);
           }
       }
   }
--- a/src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
+++ b/src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
@@ -1,23 +1,27 @@
 package slimeknights.tconstruct.smeltery.tileentity;
 
+import java.util.List;
+
+import javax.annotation.Nonnull;
+
+import org.apache.logging.log4j.Logger;
+
 import net.minecraft.client.gui.inventory.GuiContainer;
+import net.minecraft.entity.Entity;
+import net.minecraft.entity.EntityLivingBase;
+import net.minecraft.entity.monster.EntityMob;
 import net.minecraft.entity.player.InventoryPlayer;
 import net.minecraft.inventory.Container;
 import net.minecraft.item.ItemFood;
 import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.FurnaceRecipes;
 import net.minecraft.util.ITickable;
+import net.minecraft.util.math.AxisAlignedBB;
 import net.minecraft.util.math.BlockPos;
 import net.minecraft.world.World;
 import net.minecraftforge.fml.relauncher.Side;
 import net.minecraftforge.fml.relauncher.SideOnly;
-
-import org.apache.logging.log4j.Logger;
-
-import javax.annotation.Nonnull;
-
 import slimeknights.mantle.common.IInventoryGui;
-import slimeknights.mantle.multiblock.IMasterLogic;
 import slimeknights.tconstruct.library.Util;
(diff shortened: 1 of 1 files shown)


src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -143,6 +148,20 @@ protected boolean onItemFinishedHeating(ItemStack stack, int slot) {
...
+  protected void interactWithEntitiesInside() {
+      AxisAlignedBB bb = info.getBoundingBox().contract(1).offset(0, 0.5, 0).expand(0, 0.5, 0);
+
+      List<Entity> entities = getWorld().getEntitiesWithinAABB(Entity.class, bb);
+
+      for(Entity entity : entities) {
+          if (entity instanceof EntityMob) {
+              EntityLivingBase entityLiving = (EntityLivingBase) entity;
+
+              entityLiving.setHealth(0F);
```
liach: I think dealing fire or lava damage consistently is better than a simple `setHealth`.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -143,6 +148,20 @@ protected boolean onItemFinishedHeating(ItemStack stack, int slot) {
   protected int getUpdatedInventorySize(int width, int height, int depth) {
     return 9 + (3 * width * height * depth);
   }
+  
+  protected void interactWithEntitiesInside() {
+      AxisAlignedBB bb = info.getBoundingBox().contract(1).offset(0, 0.5, 0).expand(0, 0.5, 0);
+
+      List<Entity> entities = getWorld().getEntitiesWithinAABB(Entity.class, bb);
+
+      for(Entity entity : entities) {
```
liach:
```java
for (EntityLivingBase entity : getWorld().getEntitiesWithinAABB(EntityLivingBase.class, bb);) {
```

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -54,6 +58,7 @@ public void update() {
       // we heat items every tick, as otherwise we are quite slow compared to a vanilla furnace
       if(tick % 4 == 0) {
         heatItems();
+        interactWithEntitiesInside();
```
KnightMiner: The smeltery only runs this every 20 ticks, or `tick == 0`

KnightMiner:
Do the mobs still drop items when killed this way? I am not entirely sure if there is a way to kill them without dropping items, but if not it would be a good idea to kill item entities inside as well, as otherwise they can cause lag as mobs inside spawn and are killed (especially if some mod adds an item that does not despawn)

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -143,6 +148,18 @@ protected boolean onItemFinishedHeating(ItemStack stack, int slot) {
...
   }
+  
+  protected void interactWithEntitiesInside() {
+      AxisAlignedBB bb = info.getBoundingBox().contract(1).offset(0, 0.5, 0).expand(0, 0.5, 0);
+
+      List<EntityLivingBase> entities = getWorld().getEntitiesWithinAABB(EntityLivingBase.class, bb);
+
+      for(EntityLivingBase entity : entities) {
+          if (entity instanceof EntityMob) {
+              entity.setHealth(0F);
```
liach: I think there is a `entity.setDead();` method, more convenient than killing.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -143,6 +150,18 @@ protected boolean onItemFinishedHeating(ItemStack stack, int slot) {
...
     return 9 + (3 * width * height * depth);
   }
+  
+  protected void interactWithEntitiesInside() {
+      AxisAlignedBB bb = info.getBoundingBox().contract(1).offset(0, 0.5, 0).expand(0, 0.5, 0);
+
+      List<EntityLivingBase> entities = getWorld().getEntitiesWithinAABB(EntityLivingBase.class, bb);
+
+      for(EntityLivingBase entity : entities) {
+          if(entity instanceof EntityMob) {
```
bonii-xx: Add an extra check if the entity is alive

src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSearedFurnace.java
```diff
@@ -1,23 +1,26 @@
 package slimeknights.tconstruct.smeltery.tileentity;
 
+import java.util.List;
```
bonii-xx: No reformatting in PRs

bonii-xx:
I suppose this is the easiest way to get it done quickly. I wanted to avoid something like that, but I don't think we'll get a solution anytime soon otherwise.

MightyCupcakes:
Done. Followed the import order in the other tile entities.

PR #2720, opened 2017-01-14
created zh_TW.lang 

CLOVERsquare:
I translated it. But probably only 70%.I'll keep translating :)


resources/assets/tconstruct/lang/zh_TW.lang
```diff
@@ -0,0 +1,965 @@
...
+tile.tconstruct.tooltables.partbuilder.name=部件加工台
+tile.tconstruct.tooltables.toolstation.name=工具組裝台
+tile.tconstruct.tooltables.patternchest.name=模具箱
+tile.tconstruct.tooltables.patternchest.tooltip=可存放模具或是鑄模
+tile.tconstruct.tooltables.partchest.name=部件箱
+tile.tconstruct.tooltables.partchest.tooltip=可存放大量的工具部件
+tile.tconstruct.toolforge.name=工具鍛造台
+
+tile.tconstruct.soil.grout.name=水泥
+tile.tconstruct.soil.grout.tooltip=我是水泥！
```
liach: I suggest you to borrow some idea from the simplifed Chinese localization for ones like this
https://github.com/SlimeKnights/TinkersConstruct/blob/master/resources/assets/tconstruct/lang/zh_CN.lang#L16

3TUSK: Wait for seconds. You have this: https://github.com/SlimeKnights/TinkersConstruct/blob/1.7.10/resources/assets/tinker/lang/zh_TW.lang

PR #2755, opened 2017-02-04
fix client crash with players morphed into non-bipedal creatures

### LINKED ISSUE #2754: Mod conflict with metamorph (client crash)
When a player is morphed into a creature that is not a biped, another player looking at them will experience a crash with the following log.

https://gist.github.com/chimericdream/e695d8141ca7db0cf09e1f6b2a89253f

chimericdream:
Fixes #2754 

I realize this isn't necessarily a bug in Tinker's, but it's still frustrating, and I think the fix is pretty straightforward.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tools/ranged/RangedRenderEvents.java
+++ b/src/main/java/slimeknights/tconstruct/tools/ranged/RangedRenderEvents.java
@@ -31,18 +31,13 @@ public class RangedRenderEvents {
       left = EnumHand.MAIN_HAND;
     }
 
-    try {
+    if (event.getRenderer().getMainModel() instanceof ModelBiped) {
       if(isCarryingLoadedCrossbow(player, right)) {
         ((ModelBiped) event.getRenderer().getMainModel()).rightArmPose = ModelBiped.ArmPose.BOW_AND_ARROW;
       }
       else if(isCarryingLoadedCrossbow(player, left)) {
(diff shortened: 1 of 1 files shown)


bonii-xx:
No. You don't go and catch exceptions like that, you go and check instances before casting. Generally speaking you should never cast something before ensuring it's the right class, exactly because of stuff like this, and yes this also makes it a bug in TiC.

chimericdream:
Normally I'd completely agree. In this instance, I made the (incorrect) assumption that TiC explicitly wanted/needed instances of `ModelBiped` for rendering purposes, since other models may not even have right or left hands in which to render tools.

However, I also readily admit that I have almost zero experience working with MC mods, so I didn't really expect this to be the right solution on the first go.

Is the assumption that the right/left hand rendering should only happen for bipeds correct? Or do non-bipeds also have a "bow and arrow" pose that can be used?

chimericdream:
I've updated the PR, but not really knowing what the Forge/Minecraft APIs look like, it's still very much a "blind" fix. Feel free to close this PR if there is a more appropriate fix for rendering the player's model.

bonii-xx:
Yes, it requires a ModelBiped. Your mistake was using exceptions to handle it. Exceptions are fine.. as long as they're never thrown, then they result in a performance impact. Hence you don't catch the exception, but you prevent it from ever being thrown. And that has nothing to do with Minecraft in general, that's a standard coding practice.

The fix is correct now.

PR #2840, opened 2017-04-11
Fix incompatibility between Squake and Slime Boots

### LINKED ISSUE #2453: Slime Boots cause incompatibility with Squake Mod
The new slime boots cause the Squake movement system to freak out.

[What Squake should look like.](https://gfycat.com/AliveMadEft)

[What it looks like with the boots equipped.](https://gfycat.com/ElatedPrestigiousKittiwake)

---

I'm also reporting this problem to the maker of Squake, but I don't really know whose mod's at fault...

Theray070696:
Would fix #2453


laz2727:
It's not a fix, it's a dirty hack and you have to put these in your mod.

bonii-xx:
Squake simply is incompatible with anything that changes movement. Just don't equip slimeboots.

PR #2925, opened 2017-05-25
Fix bucket casting recipe not properly setting the output

KnightMiner:
I would have just committed directly to master, but I made a change to the API that I want your opinion on.

Basically, I swapped the BucketCastingRecipe from using a singleton instance to taking an item in the constructor so anther mod can add a recipe for their fluid container by calling the constructor. I wanted to be certain you were fine with removing the old protected constructor as it was likely unused in our API anyways (and the 1.11 build is still Alpha)


bonii-xx:
Guess that works easily now

Issue #2928, opened 2017-05-26
Armor creation seemingly impossible with current api


Lance5057:
Hi I'm the Tinkers Defense dev. So according to mezz mc now requires all armor to extend itemArmor, I guess the way I pulled it off in 1.7 was a fluke thing. Problem with that is though now I can't extend ToolCore. So I tried creating a toolcore specifically for armor stuff but immediately ran into the issue that almost everything in the api only wants toolcore and none of the related interfaces. So at least as it seems now I'm unable to create my armors again. Is it possible that the api could be changed to allow for items that only implement an interface instead of using an extension?
Thanks,
Lance5057

bonii-xx:
What part of the api are you trying to use, that requires you to use ToolCore..?

Lance5057:
TinkerRegistry is the big one. ToolBuildGuiInfo also requires a TinkersItem which requires an extension.

bonii-xx:
ToolBuildGuiInfo is for Tools. If you're making armor you should have your own block with its own GUI. Might be that ToolBuildGuiInfo is missing an abstraction layer, but I don't remember what it's used for ad-hoc. Code-wise everything else should be separated accordingly.

bonii-xx:
I suggest you take all the stuff needed and pull it into the ITinkerable interface, and send me a PR with that.

PR #2961, opened 2017-06-11
PR for Issue #2928

Lance5057:
Moves methods from TinkerItem to ITinkerable
Changes Registry to use ITinkerable instead of ToolCore
Changes various methods throughout to be compatible with Registry changes

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/book/content/ContentTool.java
+++ b/src/main/java/slimeknights/tconstruct/library/book/content/ContentTool.java
@@ -64,8 +64,8 @@ public class ContentTool extends TinkerPage {
       toolName = parent.name;
     }
     if(tool == null) {
-      tool = TinkerRegistry.getTools().stream()
-                           .filter(toolCore -> toolName.equals(toolCore.getIdentifier()))
+      tool = (ToolCore) TinkerRegistry.getTools().stream()
+                           .filter(toolCore -> toolName.equals(((ToolCore) toolCore).getIdentifier()))
                            .findFirst()
                            .orElseThrow(() -> new RuntimeException("Unknown tool " + toolName));
     }
--- a/src/main/java/slimeknights/tconstruct/library/book/sectiontransformer/ToolSectionTransformer.java
+++ b/src/main/java/slimeknights/tconstruct/library/book/sectiontransformer/ToolSectionTransformer.java
@@ -10,6 +10,7 @@ import slimeknights.mantle.client.book.data.PageData;
 import slimeknights.tconstruct.library.TinkerRegistry;
 import slimeknights.tconstruct.library.book.content.ContentListing;
 import slimeknights.tconstruct.library.book.content.ContentTool;
+import slimeknights.tconstruct.library.tinkering.ITinkerable;
 import slimeknights.tconstruct.library.tools.ToolCore;
 
 @SideOnly(Side.CLIENT)
@@ -24,11 +25,11 @@ public class ToolSectionTransformer extends ContentListingSectionTransformer {
     // only add tool pages if the tool exists
     if(page.content instanceof ContentTool) {
       String toolName = ((ContentTool) page.content).toolName;
-      Optional<ToolCore> tool = TinkerRegistry.getTools().stream()
-                                              .filter(toolCore -> toolName.equals(toolCore.getIdentifier()))
+      Optional<ITinkerable> tool = TinkerRegistry.getTools().stream()
+                                              .filter(toolCore -> toolName.equals(((ToolCore) toolCore).getIdentifier()))
                                               .findFirst();
 
-      tool.ifPresent(toolCore -> listing.addEntry(toolCore.getLocalizedName(), page));
+      tool.ifPresent(toolCore -> listing.addEntry(((ToolCore)toolCore).getLocalizedName(), page));
     }
     else {
       super.processPage(book, listing, page);
--- a/src/main/java/slimeknights/tconstruct/library/tinkering/ITinkerable.java
+++ b/src/main/java/slimeknights/tconstruct/library/tinkering/ITinkerable.java
@@ -1,6 +1,10 @@
 package slimeknights.tconstruct.library.tinkering;
 
 import net.minecraft.item.ItemStack;
+import net.minecraft.nbt.NBTTagCompound;
+import net.minecraftforge.fml.relauncher.Side;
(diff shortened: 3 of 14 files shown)


src/main/java/slimeknights/tconstruct/library/tinkering/ITinkerable.java
```diff
@@ -18,4 +22,32 @@
...
+  //From TinkersItem
+  public List<PartMaterialType> getRequiredComponents();
+  //not sure why these both exist but I'll include both just in case
+  public List<PartMaterialType> getToolBuildComponents();
+  
+  public ItemStack buildItemFromStacks(ItemStack[] stacks);
+  
+  public ItemStack buildItem(List<Material> materials);
+  
+  public NBTTagCompound buildItemNBT(List<Material> materials);
```
bonii-xx: Implementation detail, does not belong into the interface. Should actually be protected in the class.

src/main/java/slimeknights/tconstruct/library/tinkering/ITinkerable.java
```diff
@@ -18,4 +22,32 @@
...
+  
+  public NBTTagCompound buildData(List<Material> materials);
```
bonii-xx: Implementation detail, does not belong into the interface.

src/main/java/slimeknights/tconstruct/tools/common/client/module/GuiButtonsToolStation.java
```diff
@@ -42,8 +43,8 @@ public void updatePosition(int parentX, int parentY, int parentSizeX, int parent
       addSideButton(button);
     }
 
-    for(Item item : parent.getBuildableItems()) {
-      ToolBuildGuiInfo info = TinkerRegistryClient.getToolBuildInfoForTool(item);
+    for(ITinkerable item : parent.getBuildableItems()) {
+      ToolBuildGuiInfo info = TinkerRegistryClient.getToolBuildInfoForTool((Item) item);
```
bonii-xx: This will crash.

src/main/java/slimeknights/tconstruct/library/book/content/ContentTool.java
```diff
@@ -64,8 +64,8 @@ public void load() {
       toolName = parent.name;
     }
     if(tool == null) {
-      tool = TinkerRegistry.getTools().stream()
-                           .filter(toolCore -> toolName.equals(toolCore.getIdentifier()))
+      tool = (ToolCore) TinkerRegistry.getTools().stream()
+                           .filter(toolCore -> toolName.equals(((ToolCore) toolCore).getIdentifier()))
```
bonii-xx: This will crash.

src/main/java/slimeknights/tconstruct/library/book/sectiontransformer/ToolSectionTransformer.java
```diff
@@ -24,11 +25,11 @@ protected void processPage(BookData book, ContentListing listing, PageData page)
     // only add tool pages if the tool exists
     if(page.content instanceof ContentTool) {
       String toolName = ((ContentTool) page.content).toolName;
-      Optional<ToolCore> tool = TinkerRegistry.getTools().stream()
-                                              .filter(toolCore -> toolName.equals(toolCore.getIdentifier()))
+      Optional<ITinkerable> tool = TinkerRegistry.getTools().stream()
+                                              .filter(toolCore -> toolName.equals(((ToolCore) toolCore).getIdentifier()))
```
bonii-xx: This will crash.

bonii-xx:
It looks like you want to expose a lot of internal implementation details. That's not really a good idea, and changing the registry stuff will also break every other mod that relies on it, I can't really make such a breaking change.

All the registry stuff and TinkersItem etc. is built to provide a tool system. You want armor. I understand that you want to reuse the stuff TiC uses (and it makes sense), but I don't think the ItemArmor requirement can be married with the system without breaking abstraction. All the stuff how tools are built internally don't belong into the interfaces/API.

You'll have to basically copy what TiC does and implement it for your own classes. The registry returning tools also is intentional, since.. well everywhere it's used is for tools. Maybe you can "cheat" the system by having a TinkersItem delegate item that you use to build the NBT, then you can use a lot of the helper functions provided. Currently I don't see a way to get armor into the system without breaking  things and/or introducing a lot of uncertainty.

src/main/java/slimeknights/tconstruct/library/book/content/ContentTool.java
Lance5057: Neither of the book related issues caused a crash, I checked thoroughly because I expected one there too.

Lance5057:
I'm not sure why you're expecting it to be a breaking change. I plugged my mod in without changing a thing and it ran without a hitch. I'll be honest I was a little surprised myself.

I understand that its a tool system but if it could be expanded even farther it seems silly not to, especially with all the people who want it.

I actually tried doing this already before even coming here, to implement it in this way would basically make it so it was no longer and addon mod and just a copycat that does different things. It seemed rather pointless to continue trying it that way which is why I came here.

bonii-xx:
>I'm not sure why you're expecting it to be a breaking change.

I'm not expecting it to, it is. Every mod out there that uses those functions of the TinkerRegistry will not work anymore as is.

>I understand that its a tool system but if it could be expanded even farther it seems silly not to, especially with all the people who want it.

It is. Your problem stems from the change that armor needs to extend ItemArmor, which is a problem in minecraft/forge itself, not the API.

>I actually tried doing this already before even coming here, to implement it in this way would basically make it so it was no longer an addon mod and just a copycat that does different things.

I realize that, hence why I asked for a PR so I can see what's needed. And I'm telling you that that approach is better API wise. There are a few things that can probably be abstracted into different new interfaces so the underlying logic of ToolBuilder or GUI stuff can use it without relying on TinkersItem/ToolCore. That would NOT be a breaking change, since everything using toolcore right now would automatically get the interfaces due to inheriting from toolcore. However your approach is basically turning ToolCore into the API, which is bad. As I've more or less said above, your problem is not with the API, but with the implementation. The API is built so you can have different implementations.

As suggested earlier, have you tried using delegate/proxy items based on ToolCore to interface with the TiC helpers? Besides that the easiest way is to do your own implementation as you said above.

Lance5057:
Proxy items were my very last resort as I find them to be a rather ugly solution, but I suppose that's my only option so I'll look into doing that.

Issue #3066, opened 2017-07-30
WAILA integration AWOL (modid issue?)


miiichael:
Greetings, after updating my pack to 1.11, I noticed that tconstruct-specific WAILA info (eg. casting basin contents/progress) was absent.

I also see in my logs:

`[13:27:26] [Client thread/INFO] [Pulsar-tconstruct/tconstruct]: Skipping Pulse WailaIntegration; missing dependency: Waila`

I'm therefore wondering if the problem is that it's looking for 'Waila', when (since the 1.11 lowercasening of modids) it should instead be looking for 'waila', as all my 1.10 logs contain many messages of the form:

`[13:00:49] [Client thread/WARN]: * The modid Waila is not the same as it's lowercase version. Lowercasing will be enforced in 1.11`

miiichael:
And [here's the fml-client-latest.log](https://github.com/SlimeKnights/TinkersConstruct/files/1185617/fml-client-latest.log.txt) for a brief play using just Forge, tconstruct and waila.

bonii-xx:
Yes, the modid is incorrect due to capitalization. I think @alexbegt has a handle on the 1.12 alternatives, maybe he can take a look.

DoomsDoorWay:
I'm using HWYLA and I'm having the same issue and from what I can tell from the integration implemented by other mods its a simple fix of just making it lowercase "[waila/]: Enabling mod waila"

KnightMiner:
Yeah, I think 1.11 enforced lowercase mod IDs, but we forgot to update it here

DoomsDoorWay:
Yeah I just took a very sledgehammer approach and modified the waila class without doing any decompiling ect using only 7zip and notepad++ to lowercase the Waila and strangely it actually worked lol and the integration is working now

miiichael:
Thanks! Though I don't suppose there will be a 1.11.2 release for this? *looks longingly at curseforge*

bonii-xx:
Nope.

miiichael:
Dang.

PR #3086, opened 2017-08-06
Fix for waila integration in 1.11 fixes issue #3066

DoomsDoorWay:
Lowercase waila mod requirement for compatibility with 1.11 case sensitivity.


KnightMiner:
We have already released our last build for 1.11 as far as I'm aware, and I already have this fixed in 1.12, just not pushed.

Also, this PR does not function at all, you lowercased a variable name which simply causes a complication error, the mod ID is a string below. Please actually test PRs instead of just committing on Github.

Issue #3079, opened 2017-08-04
Bouncepads


bonii-xx:
# Bouncepads in TiC2
The basic premise of bouncepads is the same as always: You touch them. They go *floing* and off you go.

The old bouncepads however had a few issues that made it really hard to construct reliable transportation routes. Most notably where you landed depended mostly on where you entered the bouncepad.. or more accurately where in the bouncepad you'd end up during the tick.

I hence propose 3 different variants of bouncepads:

## Regular Bouncepads
These are our lovely, green, bouncepads from 1.7.10 and before

##Awesome Bouncepads (temp. name)
To solve the reliability problem we simply center the player before applying any force, meaning when entering a bouncepad you'll always start from exactly the same location. Since players can control the "fall" midair it might be necessary to also apply an extra effect that prevents player movement midair (should be doable clientside by preventing/resetting input)

##Awesomest Bouncepads (temp. name)
You know how awesome the awesome bouncepads are? Imagine combining that with the effect the Slimesling gives. Far reach. Lots of boing.

# Thoughs 'n Stuff
How are the do the different variants crafted? Green/Blue/Magmaslime? Do we allow each color to be used independently, and add the awesome™ effects by crafting them with redstone/lapis/something else? Is the "don't-move-midair" effect too restrictive and should be optional? How can it be turned on/off (crafting them into a different variant? just shift clicking?)

## Technical differences To 1.7.10
We also want to eliminate some restrictions of the old bouncepads, similar to the slime channels/punji sticks.
That means: All slime colors should be possible, and bouncepads should be placeable on all sides of a block.
Rendering should be a fixed model, not a TESR.
Also the crafting recipe should probably give 2 Bouncepads per slime block overall. They were pretty expensive before and are only half a block high.

Landmaster:
@bonii-xx How would the mechanics of a bounce pad on the walls or ceiling work (especially the latter)?

laz2727:
The "don't-move-midair" effect is probably fine on a precision bouncepad. Maybe a version without it should replace a basic bouncepad, since it's a straight upgrade.

Crafting idea: pour molten stuff on a basic one! That is surely a good idea that won't mess anything up.

PR #3121, opened 2017-08-23
Basic bounce pad ("Awesome(st)" bounce pads not yet implemented)

Landmaster:
This PR is a rudimentary implementation of the basic bounce pads which can be attached to any face and come in all 5 slime types. They are functional, but some details have not been worked out yet.
Current recipe: 1 Slime Channel + 1 Slimeball = 1 Bouncepad (Yields about 1.7 bouncepads per slime block)
Discussed somewhat in #3079


bonii-xx:
Just throwing it out there that I didn't have time to look at it yet, but I haven't forgotten you. ;)

KnightMiner:
I have had a chance to test this a bit, and have a couple comments:

* Bounce pads seem to be reliable for players as long as I do not move
* Mobs are very unreliable as the AI causes them to passively move almost every time. If we do not prevent players from moving, other mobs should definitely be stopped.
* Items do not work at all. This is probably due to vanilla Minecraft code applying a negative momentum to all items on the ground, you can see how I fixed it for upwards slime channels [here](https://github.com/SlimeKnights/TinkersConstruct/blob/master/src/main/java/slimeknights/tconstruct/gadgets/block/BlockSlimeChannel.java#L320-L322)
* Block model looks a little weird sometimes due to faces being out of order, but that is mostly vanilla's fault as it hates rendering transparent blocks together. Not sure if there is anything we can do to mask it (we could maybe split the inner cube into an above fluid and below fluid model and see if that helps)

KnightMiner:
I am going to close this as it was never really finished and its been a couple years, plus 1.12 development is basically done. If someone wants to bring this back for 1.15, feel free.

PR #3129, opened 2017-08-26
jar tasks gets finalized by reobfJar task

BrainStone:
- Fixes issues with uploading an deobfuscated jar to the maven


BrainStone:
I can go down further the version tree.

bonii-xx:
Not necessary I think.

PR #3180, opened 2017-09-29
Small performance improvement for model CacheKey

mezz:
Ran across this while profiling items rendering in JEI and looking for any simple small fixes.
`NBTTagCompound#toString` is apparently *super* slow, and the nbt itself can be used as a cache key just as well without converting it to a string first.

The changes to the `TagUtil` checks are because vanilla actually does those same checks now, so they were made redundant.

Here's profiling before and after: (9.8% to 0.1%)
cachekey-before

cachekey-after


bonii-xx:
Nice. Yeah, that wasn't possible before, NBT got changed quite a bit in the last versions... to the better. Heh. Thanks!

PR #3182, opened 2017-09-30
Improve FPS in JEI by culling quads of BakedSimpleItem when in the GUI

### LINKED ISSUE #3176: Tinker's weapons have one quad per pixel and repeat them 7 times
Most items seem to have one quad, the face.
Tinker's seem to have one quad per pixel in the image.
Compounding this issue, the items use Mantle's `BakedSimple` which gives the same quads for all 6 sides and the null side.

This makes rendering these items extremely slow, enough to bring JEI to its knees without some aggressive optimizations. I ended up looking into this in hopes of removing some optimizations from JEI that are causing minor compatibility issues. It seems like the fixes here should be relatively straightforward and the issues were just not noticed before.

mezz:
Fixes #3176

This basically implements the same culling logic found in fry's fix here:
https://github.com/MinecraftForge/MinecraftForge/commit/f42137907e0a87adf1522fded5827086fc0b13f2 
The gist of it is that you can cull most of the quads off of items when they are in the gui because you can't see the sides.

Combined with the previous fixes (https://github.com/SlimeKnights/TinkersConstruct/commit/48b420bfce63159fa31c75eb896a94794cf17661 and #3180), this brings my FPS over 60 even when rendering every single Tinker's item at once in JEI!
Before this PR I got about 40 FPS, before all the fixes together I was getting like 9. Not bad :D

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/client/model/BakedSimpleItem.java
+++ b/src/main/java/slimeknights/tconstruct/library/client/model/BakedSimpleItem.java
@@ -5,30 +5,75 @@ import com.google.common.collect.ImmutableMap;
 
 import java.util.List;
 
-import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import javax.vecmath.Matrix4f;
 
-import slimeknights.mantle.client.model.BakedSimple;
+import net.minecraft.block.state.IBlockState;
+import net.minecraft.client.renderer.block.model.BakedQuad;
+import net.minecraft.client.renderer.block.model.IBakedModel;
+import net.minecraft.client.renderer.block.model.ItemCameraTransforms;
+import net.minecraft.client.renderer.block.model.ItemOverrideList;
+import net.minecraft.client.renderer.texture.TextureAtlasSprite;
+import net.minecraft.util.EnumFacing;
+import net.minecraftforge.client.model.ItemLayerModel;
+import net.minecraftforge.client.model.PerspectiveMapWrapper;
+import net.minecraftforge.common.model.TRSRTransformation;
+import org.apache.commons.lang3.tuple.Pair;
 
-public class BakedSimpleItem extends BakedSimple {
+/**
+ * Based on {@link ItemLayerModel.BakedItemModel}
+ * Implements culling for items in the GUI
+ */
+public class BakedSimpleItem implements IBakedModel {
+  private final ImmutableList<BakedQuad> quads;
+  private final TextureAtlasSprite particle;
+  private final ImmutableMap<ItemCameraTransforms.TransformType, TRSRTransformation> transforms;
+  private final IBakedModel otherModel;
+  private final boolean isCulled;
+  private final ItemOverrideList overrides;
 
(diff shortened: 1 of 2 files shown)


src/main/java/slimeknights/tconstruct/library/client/model/BakedSimpleItem.java
```diff
@@ -5,30 +5,75 @@
...
 
-  public BakedSimpleItem(List<net.minecraft.client.renderer.block.model.BakedQuad> quads, ImmutableMap<net.minecraft.client.renderer.block.model.ItemCameraTransforms.TransformType, net.minecraftforge.common.model.TRSRTransformation> transforms, net.minecraft.client.renderer.texture.TextureAtlasSprite particle, boolean ambientOcclusion, boolean isGui3d, net.minecraft.client.renderer.block.model.ItemOverrideList overrides) {
-    super(quads, transforms, particle, ambientOcclusion, isGui3d, overrides);
+  public boolean isAmbientOcclusion() { return true; }
+  public boolean isGui3d() { return false; }
+  public boolean isBuiltInRenderer() { return false; }
+  public TextureAtlasSprite getParticleTexture() { return particle; }
+  public ItemOverrideList getOverrides() { return overrides; }
+  public List<BakedQuad> getQuads(@Nullable IBlockState state, @Nullable EnumFacing side, long rand) {
+    if(side == null) return quads;
```
bonii-xx: Really mezz, no braces? You're better than that ;P

src/main/java/slimeknights/tconstruct/library/client/model/BakedSimpleItem.java
```diff
@@ -5,30 +5,75 @@
...
-  public BakedSimpleItem(ImmutableList<net.minecraft.client.renderer.block.model.BakedQuad> quads, ImmutableMap<net.minecraft.client.renderer.block.model.ItemCameraTransforms.TransformType, net.minecraftforge.common.model.TRSRTransformation> transforms, net.minecraft.client.renderer.block.model.IBakedModel base) {
-    super(quads, transforms, base);
+  public BakedSimpleItem(ImmutableList<BakedQuad> quads, ImmutableMap<ItemCameraTransforms.TransformType, TRSRTransformation> transforms, IBakedModel original) {
+      this(quads, original.getParticleTexture(), transforms, original.getOverrides(), null);
   }
 
-  public BakedSimpleItem(List<net.minecraft.client.renderer.block.model.BakedQuad> quads, ImmutableMap<net.minecraft.client.renderer.block.model.ItemCameraTransforms.TransformType, net.minecraftforge.common.model.TRSRTransformation> transforms, net.minecraft.client.renderer.block.model.IBakedModel base) {
-    super(quads, transforms, base);
+  private BakedSimpleItem(ImmutableList<BakedQuad> quads, TextureAtlasSprite particle, ImmutableMap<ItemCameraTransforms.TransformType, TRSRTransformation> transforms, ItemOverrideList overrides, @Nullable IBakedModel unculledModel) {
```
bonii-xx: Without looking at forge code.. when is this constructor ever called with a culled model? I couldn't find where it all comes together in the forge PR, kinda feels like something is missing.

mezz: The code to do this is tricky af, you use the public constructor to create the unculled model.
That constructor calls the private one to create the culled model.
The two are referencing each other, and switch depending on if you are rendering in GUI or not.

bonii-xx: Yes, that's what I expected it to do. Apparently I overlooked that part in the constructor. Thanks

PR #3213, opened 2017-10-27
Added Casting Channels

KnightMiner:
Adds back casting channels. Channels no longer dynamically choose flow direction, instead setting the direct on placement and through right click.

Some additional information on mechanics:

* Channels will automatically connect to the block they are placed on, but not any other sides
* Channels cannot extract liquids. Instead, a channel is typically filled using a faucet.
* Channels will prioritize downwards flowing if possible. If they cannot flow down, they will split evenly between all sides.
* Channels are redstone sensitive. Toggling redstone will toggle whether the downwards connection is enabled

For a bit more information, I have a writeup on the [Ceramics Wiki](https://github.com/KnightMiner/Ceramics/wiki/Channel)


src/main/java/slimeknights/tconstruct/smeltery/network/ChannelConnectionPacket.java
```diff
@@ -0,0 +1,53 @@
...
+    TileEntity te = Minecraft.getMinecraft().world.getTileEntity(pos);
+    if(te instanceof TileChannel) {
+      ((TileChannel) te).updateConnection(side, connect);
+    }
+  }
+
+  @Override
+  public void handleServerSafe(NetHandlerPlayServer netHandler) {
+    // clientside only
+    throw new UnsupportedOperationException("Serverside only");
```
Landmaster: Should not this be `"Clientside only"`?

KnightMiner: That's based on the fluid update packet. I assume boni intended it to mean serverside only to send it, but it could have been a typo there

bonii-xx:
Errmagerrrrrrrd

bonii-xx:
Ok, so from a function wise perspective they work pretty well. A few things I noticed:

- I built a channel, and then wanted to put a tank below it. The block below the channel was free. It seems to have tried to toggle the downflow, so i couldn't place the tank there - Behaves a bit weird there.
- The indications for the flow direction aren't clear. I can't tell which way is which. Instead of a 4x4 area in the middle of the connection I'd rather go for a more arrow shaped mark (see below)
- It'd be nice to also have some kind of indication on the outside, or maybe when you hover over the block
- WAILA/TOP don't seem to display any content information. If the fluid is stuck you don't know which fluid and how much is in there.


Super fancy ASCII art for the marker:
Current:
OOOOO
OXXXXO
OXXXXO
OOOOO

Suggested for flowing "left":
OOXXO
OXXOO
OXXOO
OOXXO

KnightMiner:
Fixed the first issue and the last issue for TOP (WAILA is being stubborn, nut sure why its not working there)

Also added a new texture for the top of channels, as shown below:

Not sure what to do for an indicator on the sides since I have so little texture space to work with.

laz2727:
OQOO
QQQQ

KnightMiner:
Hard part is thats a bit vague as now left vs right is a 2 pixel difference. Then again, its better than nothing.

Issue #3233, opened 2017-11-13
Null result from CastingRecipe#getResult


Shadows-of-Fire:
So, I'm porting foundry, and somewhere along the line there's an itemstack from a `slimeknights.tconstruct.library.smeltery.CastingRecipe`'s `getResult` that is null.  As of 1.11, all ItemStack instances should be non-null.  The crashing line in my code is [this](https://github.com/Shadows-of-Fire/Foundry/blob/1.12.2/src/main/java/exter/foundry/integration/ModIntegrationTiCon.java#L164), which is called during post-init.
This crash was in a deobf environment.
TConstruct Version: `TConstruct-1.12-2.7.4.34`
Mantle Version: `Mantle-1.12-1.3.1.21`
The crash report file: https://gist.github.com/Shadows-of-Fire/746d2a1053bbd4d1bb1a6de9a9e3546f
The FML Log: https://gist.github.com/Shadows-of-Fire/fabaa83df238422d70402e382a4fcc51

KnightMiner:
Can you reproduce without any other mods except yours and Tinkers? There is a chance some addon adds a broken recipe.

Shadows-of-Fire:
I was actually just handed another crash saying that I was adding broken recipes to TiC, which I was not receiving in my dev env.  I'll look into it more when my comp finishes updating

PR #3244, opened 2017-11-20
Fixes a null ItemStack return

Shadows-of-Fire:
So, clearly this returning null at all meant there's an invalid OreCastingRecipe somewhere, but this null return is setting so many things on fire I can't really get to the bottom of what is causing the issue.  Probably what is causing #3233


KnightMiner:
Looks good. Surprised this one has been unnoticed for so long.

For the future, target 1.12 for PRs, I think boni's plan was to keep master up to date with the last release.

PR #3249, opened 2017-11-26
Add kamas, essentially a non-AOE version of the scythes

KnightMiner:
Kamas are originally a tool from Natura (where the textures come from), but now are available as a Tinkers Construct tool as essentially a non-AOE scythe. Behavior is as follows:

* Kamas can break plants and webs faster, similar to shears
* ~~When given the silk touch modifier, kamas will now shear blocks~~
* Kamas will shear blocks instead of just breaking them
* Kamas can shear entities on right click
* Kamas can right click harvest a crop
* Kamas do slightly less damage than hatchets, but attack slightly faster
* Unlike the scythe, kamas will not perform an AOE attack, even with expanders


bonii-xx:
Do they have natural silktouch? So they can harvest leaves and webs? Might be worth to add it directly.

KnightMiner:
Right now no, but it would be trivial to make the kamas always use shearing behavior. That would actually make a lot of sense since axes already quickly break plants.

KnightMiner:
I just pushed a commit which makes kamas use shearing by default. Scythes still require silk touch to use shears.

Not sure if we want to tweak anything else about the kamas now, as they are technically shears, but better now.

bonii-xx:
Yes, that's the idea. If they aren't specialized they're not worth it over hatchets. Scythe should require silktouch ofc. Also, fuck vanilla shears. >_>

PR #3254, opened 2017-11-30
Extend events functionality

Foghrye4:


bonii-xx:
Why do you need an event to prevent repairing? What's your use case?

src/main/java/slimeknights/tconstruct/library/events/TinkerCraftingEvent.java
```diff
@@ -25,7 +25,7 @@
  */
 @Cancelable
 public class TinkerCraftingEvent extends TinkerEvent {
-  private final ItemStack itemStack;
+  private ItemStack itemStack;
```
bonii-xx: No, the events do not allow to change the outcome, this needs to stay immutable.

Foghrye4:
One dude ask me to add Tinker's Construct support to my mod. Actually indeed I don't need those. I could just change NBT tag of an output item instead, correct? Idea of my mod "Iblis" is a item quality system. A more experienced player in creating item, a better item he will craft. See also https://minecraft.curseforge.com/projects/iblis

bonii-xx:
Interesting mod. So your use case would be that the player can't repair a tool he can't craft yet? Like, somebody gives him a manyullyn sword when he's lvl1, so he can't repair it?
In general I'm not convinced that preventing repair causes anything but confusion, preventing the usage of the tool alltogether is probably easier and more constructive.

Foghrye4:
Ok, I will use existing API.

Issue #1962, opened 2016-03-20
Huge lag when activating a faucet with redstone repeatedly (2.2.0b)


homemmakako:
I had 3 faucets pouring into casting tables and they worked just fine by hand. However, after making a redstone clock (with a redstone comparator) to quickly activate all of the faucets, my fps went from 40 to 5. I also tried using a lever and right clicking it continuously insted of a clock (I thought the problem could be the high frequency of the clock) but the outcome was the same, except the the fps went to 12 instead of 5.

bonii-xx:
Do you have a screenshot of the setup?

homemmakako:
http://prntscr.com/ahsjd4

bonii-xx:
Works fine for me so far.

Judging by your already low 40fps without anything going on I wager it's the graphic updates of the redstone causing it.

laz2727:
Probably the lamp.

2016-03-22 22:09 GMT+05:00 boni <notifications@github.com>:

TherminatorX:
@laz2727 The lamp is not powered by the redstone dust in this screenshot, and is only an indicator of the clock being on.

That said, redstone by itself can cause lighting-related lag when switching states. Always light up your circuitry to prevent this from becoming a problem. (This is just advice and may not solve the root of the problem)

TherminatorX:
Interesting quirk found: faucets will drain when they detect a change in redstone power level, not just having or not having a signal. This is definitely an issue, and it might be the cause of this.

TherminatorX:
@homemmakako Until Boni fixes this 'pour' performance :wink:, I suggest rearranging your setup to this:
http://imgur.com/wthd34R
Hooray for vanilla mechanics! (And basins depositing solid contents into hoppers)
Just be careful logging out while it's running, as per #1967

homemmakako:
Yeah, I guess I'll have to wait, though that other setup seems like a good idea as well. Oh, and btw @TherminatorX , nice pun :D

codewarrior0:
>I wager it's the graphic updates of the redstone causing it.

What's curious is that this only happens when the redstone clock is directly adjacent to the smeltery. Putting the clock one block away from the smeltery, yet still in the same chunk, does not exhibit the FPS loss or the stutter.

Just a hunch: Could this have anything to do with the way the smeltery multiblock will rebuild itself whenever a block is changed adjacent to it?

bonii-xx:
[...]

PR #3257, opened 2017-12-02
backport redstone fix to 1.8.9

dgriffen:
I've been stuck on 1.8.9 mods since I want to use Thaumcraft and the redstone smeltery lag is just awful. This PR is to backport the fix in #1962 to version 1.8.9


bonii-xx:
We don't support 1.8.9 anymore and we're not planning on doing any more releases for it.

dgriffen:
All right, I'll just use the build privately then.

PR #3258, opened 2017-12-03
Add new ore preference system, to allow oredict outputs to use a specific mod if available

KnightMiner:
As originally discussed in #2884, this adds the feature to choose a preference for oredict outputs. For example, adding Immersive Engineering to the top of the list will cause IE copper and steel to be used for outputs over Thermal Expansion or Chisel blocks.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/common/config/Config.java
+++ b/src/main/java/slimeknights/tconstruct/common/config/Config.java
@@ -18,6 +18,7 @@ import java.util.List;
 import slimeknights.mantle.pulsar.config.ForgeCFG;
 import slimeknights.tconstruct.TConstruct;
 import slimeknights.tconstruct.library.Util;
+import slimeknights.tconstruct.library.utils.RecipeUtil;
 
 public final class Config {
 
@@ -45,6 +46,10 @@ public final class Config {
   public static double oreToIngotRatio = 2;
   public static String[] craftingStationBlacklist = new String[] {
       "de.ellpeck.actuallyadditions.mod.tile.TileEntityItemViewer"
+  };
+  private static String[] orePreference = {
+      "minecraft",
+      "tconstruct"
   };
   //public static List<String> craftingStationBlacklist = Collections.emptyList();
 
@@ -193,7 +198,11 @@ public final class Config {
       craftingStationBlacklist = prop.getStringList();
       propOrder.add(prop.getName());
 
-
+      prop = configFile.get(cat, "orePreference", orePreference);
+      prop.setComment("Preferred mod ID for oredictionary outputs. Top most mod ID will be the preferred output ID, and if none is found the first output stack is used.");
+      orePreference = prop.getStringList();
+      RecipeUtil.setOrePreferences(orePreference);
+      propOrder.add(prop.getName());
     }
     // Worldgen
     {
--- a/src/main/java/slimeknights/tconstruct/common/config/Config.java
+++ b/src/main/java/slimeknights/tconstruct/common/config/Config.java
@@ -49,7 +49,12 @@ public final class Config {
   };
   private static String[] orePreference = {
       "minecraft",
-      "tconstruct"
+      "tconstruct",
+      "thermalfoundation",
+      "forestry",
+      "immersiveengineering",
+      "embers",
+      "ic2"
   };
   //public static List<String> craftingStationBlacklist = Collections.emptyList();
 
--- a/src/main/java/slimeknights/tconstruct/library/MaterialIntegration.java
+++ b/src/main/java/slimeknights/tconstruct/library/MaterialIntegration.java
@@ -1,14 +1,12 @@
 package slimeknights.tconstruct.library;
 
 import net.minecraft.block.Block;
-import net.minecraft.item.ItemStack;
 import net.minecraft.item.crafting.IRecipe;
 import net.minecraftforge.fluids.Fluid;
 import net.minecraftforge.fluids.FluidRegistry;
 import net.minecraftforge.oredict.OreDictionary;
 import net.minecraftforge.registries.IForgeRegistry;
 
-import java.util.List;
 import slimeknights.tconstruct.TConstruct;
 import slimeknights.tconstruct.common.config.Config;
 import slimeknights.tconstruct.library.materials.Material;
@@ -145,14 +143,7 @@ public class MaterialIntegration {
   private void registerRepresentativeItem() {
     // also set the representative item
     if(material.getRepresentativeItem().isEmpty() && representativeItem != null && !representativeItem.isEmpty()) {
-      List<ItemStack> ore = OreDictionary.getOres(representativeItem, false);
-      if(!ore.isEmpty()) {
-        ItemStack itemStack = ore.get(0).copy();
-        if(itemStack.getMetadata() == OreDictionary.WILDCARD_VALUE) {
-          itemStack.setItemDamage(0);
-        }
-        material.setRepresentativeItem(itemStack);
-      }
+      material.setRepresentativeItem(representativeItem);
     }
   }
 
--- a/src/main/java/slimeknights/tconstruct/library/smeltery/OreCastingRecipe.java
+++ b/src/main/java/slimeknights/tconstruct/library/smeltery/OreCastingRecipe.java
@@ -4,56 +4,44 @@ import net.minecraft.init.Blocks;
 import net.minecraft.item.ItemStack;
 import net.minecraftforge.fluids.Fluid;
 import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.oredict.OreDictionary;
+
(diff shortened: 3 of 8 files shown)


src/main/java/slimeknights/tconstruct/common/config/Config.java
```diff
@@ -46,6 +47,10 @@ private Config() {
   public static String[] craftingStationBlacklist = new String[] {
       "de.ellpeck.actuallyadditions.mod.tile.TileEntityItemViewer"
   };
+  private static String[] orePreference = {
+      "minecraft",
+      "tconstruct"
```
bonii-xx: Should probably put some more defaults here so people don't get chisel for copper etc. by default. TE always is a good fit ("thermalfoundation" I think), and immersive engineering.

src/main/java/slimeknights/tconstruct/library/MaterialIntegration.java
```diff
@@ -145,14 +143,7 @@ public void integrate() {
   private void registerRepresentativeItem() {
     // also set the representative item
     if(material.getRepresentativeItem().isEmpty() && representativeItem != null && !representativeItem.isEmpty()) {
-      List<ItemStack> ore = OreDictionary.getOres(representativeItem, false);
```
bonii-xx: Wondering what the impact of existing mods is for that. Like XU2 that adds materials. Or IE.

src/main/java/slimeknights/tconstruct/library/smeltery/OreCastingRecipe.java
```diff
@@ -4,42 +4,56 @@
...
+import slimeknights.tconstruct.library.utils.RecipeUtil;
 
 /**
  * A casting recipe that takes its output from an oredict entry.
  * Used for ingot casting etc.
  */
 public class OreCastingRecipe extends CastingRecipe {
 
+  /** @Deprecated use oreName instead */
+  @Deprecated
```
bonii-xx: I wonder if it wouldn't be better if we made a new class for this, and deprecated the old class. Seems a bit cleaner, but less compatible (although I hope nobody casts to that class without checking for instance first..)

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
```diff
@@ -0,0 +1,86 @@
...
+import java.util.List;
+import java.util.Map;
+import java.util.Optional;
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.NonNullList;
+import net.minecraftforge.oredict.OreDictionary;
+import slimeknights.mantle.client.CreativeTab;
+
+public class RecipeUtil {
```
bonii-xx: final

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
```diff
@@ -0,0 +1,86 @@
...
+   * @param oreName  String to check
+   * @param items    Items to search
+   */
+  private static void cachePreference(String oreName, List<ItemStack> items) {
+    // search through each preference name, finding the first item for the name
+    ItemStack preference = null;
+    for(String mod : orePreferences) {
+      Optional<ItemStack> optional = items
+          .stream()
+          .filter(stack -> !stack.isEmpty() && stack.getItem().getRegistryName().getResourceDomain().equals(mod))
```
bonii-xx: anyMatch()
btw I can strongly recommend https://medium.com/@tagir_valeev/use-stream-api-simpler-or-dont-use-it-at-all-ea0a44a4b1ff
Very useful for stream stuff

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
```diff
@@ -0,0 +1,86 @@
...
+
+  /**
+   * Gets the preferred ore for the given oredict name
+   * @param oreName  Oredictionary key
+   * @return  The preferred ItemStack, or ItemStack.EMPTY if the name is empty
+   */
+  public static ItemStack getPreference(String oreName) {
+    // if we found one before, use that
+    if(preferenceCache.containsKey(oreName)) {
+      return preferenceCache.get(oreName).copy();
```
bonii-xx: Very good

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
```diff
@@ -0,0 +1,86 @@
...
+    }
+    // if we found no preference, just use the first available stack
+    if(preference == null) {
+      preference = items.get(0);
+    }
+
+    // ensure we do not return a wildcard value
+    if(preference.getMetadata() == OreDictionary.WILDCARD_VALUE) {
+      NonNullList<ItemStack> subItems = NonNullList.create();
+      preference.getItem().getSubItems(CreativeTab.SEARCH, subItems);
```
bonii-xx: Is that Serverside-Safe?

bonii-xx:
Looks good overall, just a few details I'm pondering on. I'll leave them to your discretion.

src/main/java/slimeknights/tconstruct/library/MaterialIntegration.java
KnightMiner: With the way I set it up, if they manually set the item there is no change since thats what the old code did with the ore names. If they use an oredict name will be the only changes, but they did not have control over that before anyways

src/main/java/slimeknights/tconstruct/library/smeltery/OreCastingRecipe.java
KnightMiner: I considered that, though I was not sure what to name it.

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
KnightMiner: Yes, it is now since one of the 1.12 builds. Might need to double check if that was 1.12.1 only though...

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
KnightMiner: We used that in some of the model stuff, didn't we. Only downside is we would have to cache if we got no result, if that is acceptable

src/main/java/slimeknights/tconstruct/common/config/Config.java
bookerthegeek: Industrial Craft would also be a good fit in this list

src/main/java/slimeknights/tconstruct/library/MaterialIntegration.java
bonii-xx: Ok, great

src/main/java/slimeknights/tconstruct/library/smeltery/OreCastingRecipe.java
bonii-xx: OreCastingRecipe2, OreCastingRecipeNew, BetterOreCastingRecipe, UseThisInsteadOfTheOldOneOreCastingRecipe, the terrible possibilities are endless!
I guess yours or maybe SmartOreCastingRecipe or possibly even just OreDictCastingRecipe.. hrm.

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
bonii-xx: No, the remaining code would have to be adjusted to fit a single call, but that's just some code reordering. You already cache the first oredict entry in your code, it'd just require to move all the remaining code (aka get oredict items) into the cache method.

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
bonii-xx: Nice. I'm not up to date with 1.12+ changes.

src/main/java/slimeknights/tconstruct/library/utils/RecipeUtil.java
KnightMiner: Looking into It, I guess I can just return null from thay function if the item does not exist (so no cache) then use Map.getOrDefault to return ItemStack.EMPTY if the result is null.

KnightMiner:
Pushed some commits addressing the concerns

While testing after making some of these, I noticed an issue with our JEI integration which relates to this. It seems because of the way JEI now handles validations (on load instead of when viewing recipes), our empty OreCastingRecipes still show (and by extendsion empty PreferenceCastingRecipes).

According to mezz, it should be safe to just not load those recipes if they are empty by the time JEI loads as the oredict should be stable by then, though if I do that would it make sense to also cache the empty results from the preferences or leave those uncached?

Issue #3241, opened 2017-11-17
Unbreakable tool broken


blaziken663:
I just made an unbreakable pickaxe via the creative modifiers, just adding reinforcement 5 times, yet my pickaxe still broke when i was mining. 
![image](https://user-images.githubusercontent.com/33751149/32944771-efa148ea-cb90-11e7-9d1a-8f4f8804bb2d.png)

KnightMiner:
You have magically brittle from extra utilities, good chance it does not do unbreakable checks when it magically breaks the tool

Issue #3206, opened 2017-10-20
Broken "unbreakable" tool with mending moss cannot be repaired (1.12.1)


Sukasa:
Using Tinkers' Tool Levelling and Extra Utilities 2 on a server.  Hammer was broken but I didn't get around to repairing it soon enough.  Seems that the Mending Moss on it brought it up to "full" health (but still broken), and plunking it in a tool forge no longer lets me use its repair material to repair it.  Doesn't consume any iron if I try, and stays at full health + repaired.

NBT data attached.
[1d6f3406-10e2-431a-9ed0-1866ed8bc37d.zip](https://github.com/SlimeKnights/TinkersConstruct/files/1403497/1d6f3406-10e2-431a-9ed0-1866ed8bc37d.zip)

(As for why it broke while unbreakable, I'm going to place blame squarely on the ExU2 "magically brittle" modifier)

KnightMiner:
Last I checked, magically brittle does not properly break tools, it just applies the broken tag. Try with just our materials.

laz2727:
ExU bug.

KnightMiner:
Since there has been no response, I am closing this as an Extra Utilities bug. If you  want a reopen reproduce with just our materials.

Issue #3173, opened 2017-09-24
[1.12] Unbreakable pickaxe not able repair


GQuan1975:
Unbreakbale pickaxe (5x reinforcement, tool rod is from magical wood) is broken. I suppose it's due to "Magical Brittle" from magical wood. But it's not possible repair this tool with any material used in this pickaxe.

bonii-xx:
The materials used are probably missing integration then. What materials did you use for the pickaxe head?

GQuan1975:
It's Manyullyn pickaxe, before it's full reinforced, i can repair it.

GQuan1975:
Problem is in TinkersItem.java row 352

**
    // now do it all over again with the real items, to actually repair \o/
    ItemStack item = repairable.copy();

    while(item.getItemDamage() > 0) {
      int amount = calculateRepairAmount(materials, repairItems);

      // nothing to repair with, we're therefore done
      if(amount <= 0) {
        break;
      }

      ToolHelper.repairTool(item, calculateRepair(item, amount));
      // save that we repaired it :I
      NBTTagCompound tag = TagUtil.getExtraTag(item);
      tag.setInteger(Tags.REPAIR_COUNT, tag.getInteger(Tags.REPAIR_COUNT) + 1);
      TagUtil.setExtraTag(item, tag);
    }
**

repairing is in while, where checking only damage, but my pickaxe have no damage, only is mark broken  by Magically Brittle

laz2727:
Well, if the mod can break a tool without actually damaging it, that's
technically their bug since it's impossible with only TiC.

bonii-xx:
Then it's a bug in magical brittle. Tools being broken with durability is an invalid state. Please report to tema.

laz2727:
Hmm. I wonder, is it possible to repair the tool with moss/wood trait and keep it broken? It was possible in 1.7.10.

KnightMiner:
2230ba2 should make tools which are broken with no damage repairable again.

Extra Utilities still needs a more proper way to break the tools. The best way is probably just them calling `stack.setItemDamage(stack.getMaxDamage())`. I cannot easily make `ToolHelper.breakTool()` as `ToolCore.damageItem()` calls `breakTool()`.

PR #3269, opened 2017-12-10
Make ToolHelper#breakTool check for unbreakable tools.

### LINKED ISSUE #3242: Magically Brittle bug
I found a bug where a tool with magically brittle and unbreakable on it will randomly break and stay broken with noway to repair it. This is in the Direwolf20 1.12 pack

Bluexin:
Fixes #3242 #3241 #3206 #3173
I checked the ExU code, which uses the ToolHelper method for breaking tools, so this prevents it (but ExU doesn't accept PR's, and their public issue tracker seems to be left unattended).

Not sure which branch I'm supposed to target, this should work on most branches.
Can adapt the PR if needed.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/utils/ToolHelper.java
+++ b/src/main/java/slimeknights/tconstruct/library/utils/ToolHelper.java
@@ -522,6 +522,7 @@ public final class ToolHelper {
 
   public static void breakTool(ItemStack stack, EntityLivingBase entity) {
     NBTTagCompound tag = TagUtil.getToolTag(stack);
+    if (tag.getBoolean(ModReinforced.TAG_UNBREAKABLE)) return;
     tag.setBoolean(Tags.BROKEN, true);
     TagUtil.setToolTag(stack, tag);
 


bonii-xx:
Wrong fix. The bug is that you can't repair them. Calling breakTool should always break the tool.

Bluexin:
Not really.
The issue is you can't repair them because they're in a wrong state (broken
with max dura).
This check prevents them from breaking in the first place, when they're not
supposed to (preventing the wrong state).
 the method should *not* break the tool if it is unbreakable (that's what
the name implies).
Adding a way to repair unbreakable tools is a bit odd (esp since this is to
circumvent bad design on exu's side).

laz2727:
You can easily break a tool with exu even if it's not unbreakable, with
some luck and some obsidian.

Bluexin:
oh well I'll look into doing that then.
Might take just a bit longer to do, this is not my only project u_u

bonii-xx:
Unbreakable is just a bit.. badly named. It's a vanilla term, which has no concept of broken tools. Unbreakable just means it doesn't take durability damage. It doesn't mean it can't be broken by code. You can still set the tag manually. XUs trait BREAKS the tool. It doesn't care about your durability. It is behaving correctly. That you can't repair them.. now that's not intended.

Bluexin:
Alright I get it ^-^
(though imo in that case the issues shouldn't have been closed with
messages along the lines of "it's exu's fault, they're setting the tools in
an invalid state, wontfix")

bonii-xx:
Yes, that was before we knew what was actually going on, my bad

PR #3271, opened 2017-12-10
Update zh_cn.lang

DYColdWind:


KnightMiner:
Can you also add in [this new commit](https://github.com/SlimeKnights/TinkersConstruct/commit/33a1f05d124ae4c6f18b9b91f2beb73d562a6c4b)? It adds the missing book page for kamas which should be translated as well.

PR #3274, opened 2017-12-11
zh_cn & en_us supplement

DYColdWind:
@KnightMiner  I think I should not misunderstand what you mean
Because I used to modify the warehouse
So, about ‘resources/assets/tconstruct/lang/zh_cn.lang’mistake...


KnightMiner:
You should not need to create those files yourself, just change your pull request to work from the 1.12 branch.

Issue #3293, opened 2017-12-22
[1.10.2] Breaking Smeltery Causes a crash


MJRLegends:
Crash log: https://pastebin.com/D7zZD4MB
TinkersConstruct Version: 1.10.2-2.6.5
Mantle: 1.10.2-1.1.5

Im going to be submitting a fix in a min through pull request, would be amazing if you could take a look at it

MJRLegends

bonii-xx:
1.10 support has stopped quite a while ago, sorry.

PR #3294, opened 2017-12-22
Fixed crashing when breaking smeltery (Issue #3293 )

MJRLegends:
This fix is already in the MC 1.11.2 version so this is a backport of that fix to MC 1.10.2
https://github.com/MJRLegends/TinkersConstruct/blob/1.11.2/src/main/java/slimeknights/tconstruct/smeltery/tileentity/TileSmeltery.java#L270

Fixes issue #3293


bonii-xx:
We are not going to do any new releases <1.12.2

MJRLegends:
No problem, it was the first time seeing a assert so i thought it was a new way i didnt know but with other things set it does nothing from my quick google on it.

bonii-xx:
You are correct. Asserts are used in debug environment, to ensure things are working correct. The game would crash at this position. But it only works in dev environments, since the release versions don't have debug information.

PR #3328, opened 2018-01-10
Correct and add cross-player velocity interactions

alexbobp:
This pull request corrects a couple SPacketEntityVelocity packet sends to make the frying pan and presumably other weapons apply velocity updates to other players properly.

It also adds optional interaction of the slimesling with other entities, including other players, mostly using code borrowed from the frying pan.  (that was how I found the bug).  The feature is off by default and configurable.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/common/config/Config.java
+++ b/src/main/java/slimeknights/tconstruct/common/config/Config.java
@@ -37,6 +37,7 @@ public final class Config {
   public static boolean reuseStencil = true;
   public static boolean craftCastableMaterials = false;
   public static boolean chestsKeepInventory = true;
+  public static boolean slingOther = false;
   public static boolean autosmeltlapis = true;
   public static boolean obsidianAlloy = true;
   public static boolean claycasts = true;
@@ -143,6 +144,11 @@ public final class Config {
       prop = configFile.get(cat, "chestsKeepInventory", chestsKeepInventory);
       prop.setComment("Pattern and Part chests keep their inventory when harvested.");
       chestsKeepInventory = prop.getBoolean();
+      propOrder.add(prop.getName());
+
+      prop = configFile.get(cat, "slimeslingOther", slingOther);
+      prop.setComment("Slime sling pushes away other entities.");
+      slingOther = prop.getBoolean();
       propOrder.add(prop.getName());
 
       prop = configFile.get(cat, "enableClayCasts", claycasts);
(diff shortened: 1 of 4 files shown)


bonii-xx:
Ah, so the packts were sent for the wrong entity if I'm seeing this correctly? What does the other stuff do?
 The slimesling has to be aimed at blocks by design.

alexbobp:
The other code allows hitting another entity with a slimesling, which slings both you and that entity away from each other.  It's implemented as a config option, off by default, so it will have zero impact on users who don't choose to turn it on.  I wrote the feature because it was greatly desired on the server I play on.

alexbobp:
By the way this probably doesn't affect the current PR but I have future plans to add another config option to make the slimesling consume hunger to use.  This feature will probably seem more balanced when that one is in place.  But in any case, as a config option it should be harmless?

bonii-xx:
The slimesling should not have the capability to knock away other things, that's what frying pans are for. It'd also only lead to unwanted interactions most of the time.
Currently no plans to make it use hunger. The slimesling is in a pretty good place currently, it doesn't require further adjustments.

alexbobp:
Hm, I am sorry to hear that I can't have this even as a config option.  I disagree about unwanted interactions (when people want to discharge a sling without slinging, they just aim up or switch items).  Personally I really want this behavior for my servers.  I will go ahead with building out a TiCon fork for my use and anyone who likes my weird ideas, since I don't see how to do this in an addon mod.  I'll keep submitting PRs of interesting changesets I make in case your team finds any of them appealing (unless you tell me to stop)

Do you want a PR submitted with *just* the package corrections?  I could do that, though it might be just as convenient for you to manually fix the two lines.

alexbobp:
BTW boni in the future is it ok to contact you on irc with my ideas so I can save everyone time if a PR wouldn't likely be accepted?

KnightMiner:
We have a discord now which might be simpler.

No reason for a fork just for this behavior, just subscribe to the Forge event for the same callback. You will need to verify the item given to the event is a slimesling, but it's much preferred to someone else maintaining a second copy of the mod.

bonii-xx:
In general it is preferable to message us ahead of time if you plan on doing PRs, yes. That's how it's usually done. You can find us on IRC or Discord, however I'm rather hard to get a hold of lately, so an issue on github is fine too.
Yes, please have a separate PR with the packet fixes, it's a bugfix after all.
Also what does discharging without wanting to sling having to do with the issue? With unwanted interactions I mean stuff like flinging your friend to certain death by accident because he happened to walk by.

bonii-xx:
Closing this since, last time I double checked this after we talked about it, everything looked correctly. I think it was just the confusing way it was set up. Everything should be fine as is.

PR #3331, opened 2018-01-12
Fix addInformation crash with null player

asiekierka:
Seems to happen on a large modpack, predicated by the enabling of CoFHCore

Also adds some debug code for ToolClientEvents crashing if FMLClientHandler.instance().hasError() and thus the model registry was not populated with baked models (see ModelManager) - you might want to handle this more gracefully

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/utils/ToolHelper.java
+++ b/src/main/java/slimeknights/tconstruct/library/utils/ToolHelper.java
@@ -840,7 +840,11 @@ public final class ToolHelper {
   }
 
   public static float getActualDamage(ItemStack stack, EntityLivingBase player) {
-    float damage = (float) player.getEntityAttribute(SharedMonsterAttributes.ATTACK_DAMAGE).getAttributeValue();
+    float damage = (float) SharedMonsterAttributes.ATTACK_DAMAGE.getDefaultValue();
+    if (player != null) {
+      damage = (float) player.getEntityAttribute(SharedMonsterAttributes.ATTACK_DAMAGE).getAttributeValue();
+    }
+
     damage += ToolHelper.getActualAttack(stack);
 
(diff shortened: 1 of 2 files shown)


src/main/java/slimeknights/tconstruct/library/utils/ToolHelper.java
```diff
@@ -840,7 +840,11 @@ public static float calcCutoffDamage(float damage, float cutoff) {
   }
 
   public static float getActualDamage(ItemStack stack, EntityLivingBase player) {
-    float damage = (float) player.getEntityAttribute(SharedMonsterAttributes.ATTACK_DAMAGE).getAttributeValue();
+    float damage = (float) SharedMonsterAttributes.ATTACK_DAMAGE.getDefaultValue();
```
bonii-xx: Oh, I like this. Where does the call with null originate from?

asiekierka: addInformation() called during building the search tree used by the creative tab on the client.

asiekierka: addInformation() called during building the search tree used by the creative tab on the client.

bonii-xx: Shouldn't that have the local player? Or is that done before the server starts? Iirc the creative tabs are only called once the player is ingame.

asiekierka: @bonii-xx They aren't, they're also called once in init. addInformation() can also be called on the server-side by any other mod.

bonii-xx: Probably something along those lines, meh.

Issue #3337, opened 2018-01-15
Using the scythe on cobweb crashes the game


Mysteroo:
Hosting a Lan server, had a Pig-Iron Scythe in an abandoned mineshaft.

Tried using it on some cobwebs - don't remember if I attacked or right clicked, but regardless it crashed the server when I tried to do it. Then again when I tried after restarting.

KnightMiner:
If it crashed, we need the crash report. Please post it to a site such as pastebin then link it here.

temp1011:
I've recreated this. Just Tinker's 2.9.0.55 and Mantle 1.3.1.21
[crash report](https://github.com/SlimeKnights/TinkersConstruct/files/1637031/crash-2018-01-16_22.13.33-client.txt)

Harvest level not set?

KnightMiner:
Seems getToolClass is being called with a null string. I assumed that was not valid, but it shouldn't hurt to avoid that.

PR #3343, opened 2018-01-19
Fixes Scythe Crash

temp1011:
Fixes https://github.com/SlimeKnights/TinkersConstruct/issues/3337 (and possibly as well for any other tool that might try and target webs)


KnightMiner:
That's not the line causing the crash. Contains is perfectly safe with a null string.

KnightMiner:
Closing since 1ab09e1 fixes the crash.

PR #3376, opened 2018-02-11
Store the UUID and other data in player heads

josephcsible:
This makes heads not spontaneously become Steve heads in various circumstances.


bonii-xx:
That was vanilla code, it probably changed since 1.8

Issue #3441, opened 2018-03-28
[Suggestion] 1.12.2- Ex Nihilo/ skyblock Kama change


Treehawk-Watts:
With the addition of the kama, early skyblock with Ex Nihilo has been drastically changed. By enabling "shearing" on leaves, dirt creation is almost trivial right away. It would be really helpful for pack makers if there was an option in the config to move the kama to only be crafted at the tool forge, rather than also allowing it at the tool station. This would let us keep the cool new tool while gating it behind the metal blocks.

I realize suggestions aren't typically accepted but it would be a very minor change, so I thought it would be worth mentioning. Thanks.

TheQuadShot:
Although I've not played all the skyblock modpacks, the ones I have all had wooden shears and vein miner, so how does the Kama change that? and why make it more difficult than the "default" process for making dirt?

Treehawk-Watts:
In the packs I've played, the default is to use ex nihilo crook on leaves to get saplings, and then compost the saplings. Having shears makes the crook much less useful, since you can just compost the leaves themselves. I've never heard of wooden shears, but that would create the same problem as the kama.

bonii-xx:
No, I'm not going to support the boring earlygame grind, that provides no gameplay at all. If you want to force people to not use TiC tools before they have metal, you can change the crafting recipes of the different tables.

PR #3442, opened 2018-03-28
Pull Request for #3441 - Add config option for kama

Treehawk-Watts:
Simply adds and makes functional a config option for moving the kama from tool station+tool forge to tool forge only, default set to false. This is to make skyblock/ex nihilo more difficult early. See #3441.


bonii-xx:
Kama is a shear replacement. Its big brother, the scythe, already is in the toolforge.

Issue #3377, opened 2018-02-11
Double wither skeleton head drop


josephcsible:
It's currently possible for a single wither skeleton to drop two heads - one from the vanilla 2.5-5.5% chance, and one from Beheading. This seems like it shouldn't be possible.

KnightMiner:
@bonii-xx: think this is worth patching? It should be as simple as searching the drops list for the head before adding the head from beheading into the list

PR #3458, opened 2018-04-23
Prevent heads dropping twice

ALongStringOfNumbers:
This should fix https://github.com/SlimeKnights/TinkersConstruct/issues/3377


bonii-xx:
Wont work since EntityItem equality is not what you're expecting it to be. But seeing that reminds me to fix it, i'll just fix the check, thanks for the start.

PR #3470, opened 2018-05-03
Fix Faucets draining from the wrong side of the source

codewarrior0:
Fixes issues including:

A faucet on the underside of an Immersive Engineering Wooden Barrel will not drain fluids unless the *top* side of the barrel has been hammered.

A faucet on the underside of an Astral Sorcery Lightwell will not drain fluids at all, since the Lightwell only allows fluid interaction from below, and the faucet is trying to do it from above.


bonii-xx:
Derp

PR #3504, opened 2018-05-26
Added ko_kr.lang

RedPrintR:
Finished


bonii-xx:
You seem to be missing all modifiers, ingame and in the books. Intended?

RedPrintR:
Yes, it is intended. I only translated modifiers except tool tips

PR #3555, opened 2018-06-23
update ja_jp files

koh-gh:
Add m&u ja_jp translation.
Tweak ja_jp.lang.


bonii-xx:
Looks like you're still missing some files. They should be untranslated for now so the default english ones are used. Also you don't need the technical ones like the smeltery/furnace structure.

koh-gh:
Is it better to delete currently unused pages than to keep them?(For example, welcome.json)
I intentionally have not translated them.

koh-gh:
Thanks. I have removed some unused files and shared files.

PR #3574, opened 2018-07-13
Update ko_kr.lang

ksb12:


bonii-xx:
See https://github.com/SlimeKnights/TinkersConstruct/pull/3716

PR #3623, opened 2018-08-23
Added es_es lang file.

Allinxter:
Planing on doing a full spanish localization. Will translate the books later.
Also, if I may ask, how do I mofify the naming order? For example, translated but not tweaked  the Manyullyn Crossbow, wuld be translated to Manyulín Ballesta, which while understandable is gramatically incorrect. The correct spelling order would be something like "Ballesta de Manyulín".


SweetSugarSugar:
Try looking at the French translation, I don't remember how I did it but I managed to change the prefixes around.

resources/assets/tconstruct/lang/es_es.lang
```diff
@@ -0,0 +1,992 @@
...
+#################
+
+potion.carry=Cangureando
+
+#################
+#   Materials   #
+#################
+# material.<identifier>.name
+# material.<identifier>.prefix
+# material.wood.prefix=Wooden %s - Used for stuff like "Wooden Pickaxe" if no specific name is given
```
bonii-xx: See here for changing the order of the materials in tools

Allinxter: Oh, thanks. That makes sense in why Wood Tools have the correct identification.

PR #3699, opened 2018-12-14
update fr

Neerwan:


bonii-xx:
It's no use adding the english description to the french translation. The english one will be used as default for missing entries.

bonii-xx:
Just open a new one when you're ready

PR #3703, opened 2018-12-28
Fix #3702

### LINKED ISSUE #3702: Smeltery Forming and unforming causing a client freeze
**Issue description:**

If you build a 4x1 smeltery (or larger) with the controller on the base, level (Floor), a tank above it and a drain to the right, formed like so

X=Seared Brick | C=Controller | T=Tank | D=Drain | A=AIR

Y=1
XXXXXX
XXXXXX
XXCXXX

Y=2
XXXXXX
XAAAAX
XXTDXX

And break one of the seared brick on the upper level, the smeltery will form and unform, causing the liquid to render and unrender causing the client to lag and completely freeze. using a mod to force the smeltery to not render the liquid resolves this issue.

**If crashed, link to crash report (use a site such as pastebin):**
https://gist.github.com/850ed18f47d86c481045e086a611348f

**Versions:**
* Minecraft:  1.12.2
* Forge: 2772
* Mantle: 1.3.3.39
* Tinkers Construct: 10.1.87
* SpongeForge: RC3505

**Can it be reproduced with just Tinkers Construct, or are other mods required to reproduce the bug?**
Requires a dedicated server with sponge. (tested on the latest version of DW20)

LXGaming:
Prevents Multiblock structures from becoming active if a BlockMultiblockController Y position is equal to or lower than the floor layer

### DIFF
--- a/src/main/java/slimeknights/tconstruct/smeltery/multiblock/MultiblockCuboid.java
+++ b/src/main/java/slimeknights/tconstruct/smeltery/multiblock/MultiblockCuboid.java
@@ -38,6 +38,11 @@ public abstract class MultiblockCuboid extends MultiblockDetection {
     int masterY = center.getY();
     center = getOuterPos(world, center, EnumFacing.DOWN, 64).up();
 
+    // below lowest internal position
+    if (masterY < center.getY()) {
+      return null;
+    }
+
     // distances to the edges including the outer blocks
     int edges[] = new int[4];
     // order: south/west/north/east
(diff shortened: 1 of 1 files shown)


src/main/java/slimeknights/tconstruct/smeltery/multiblock/MultiblockCuboid.java
```diff
@@ -71,6 +71,11 @@ public MultiblockStructure detectMultiblock(World world, BlockPos center, int li
       }
     }
 
+    // below lowest internal position
```
bonii-xx: Good analysis and correct fix. However you should move the check up directly after the center is calculated (Line 40).

LXGaming: Good point, Done.

PR #3793, opened 2019-03-20
Swedish translation

Regnander:


resources/assets/tconstruct/lang/sv_se.lang
```diff
@@ -0,0 +1,1041 @@
...
+gui.error.incompatible_enchantments=Modifieraren %s kan inte kombineras med förtrollningen %s
+gui.error.no_rename=Detta verktyg kan inte byta namn
+
+################
+#     GUIs     #
+################
+
+# General things used in multiple GUIs
+gui.general.hover=Lägg musen över uppslagen för mer information
+gui.general.temperature.celsius=%s °c
```
KnightMiner: I'd remove the space here, it looks a bit awkward in game and is not in the english version. I also switched this to a capital C in the english one recently, so just make sure this is correct for your language's symbol for Celsius

Regnander: Swedish spelling rules says that there must be a space before a grade sign in temperature related situations, but it looks kinda tight and will most likely overflow if it reaches four digits, so I'll remove it.
C and F should be uppercase, but I must have made a blunder there. Thanks for being observant!

Issue #3730, opened 2019-01-17
Wooden Rail with Trapdoor stops inserting after first slot runs out


fer22f:
**Issue description:** If you have a hopper minecart with the first slot empty (which happens when it runs out), it stops outputting the other slots:
![image](https://user-images.githubusercontent.com/3057302/51290655-7efc8c80-19eb-11e9-9d7a-4df2d0e30cb8.png)

**Versions:**
* Minecraft: 1.12.2
* Forge: 14.23.5.2807
* Mantle: 1.3.3.39
* Tinkers Construct: 1.12-2.12.0.113

**Can it be reproduced with just Tinkers Construct, or are other mods required to reproduce the bug?** I believe it can be reproduced just with Tinkers Construct.

Looking at the code, it seems like it would give up after the first slot becomes empty:
https://github.com/SlimeKnights/TinkersConstruct/blob/23034cb63e98bba06faf1cdc4074009daf93be1f/src/main/java/slimeknights/tconstruct/gadgets/block/BlockWoodRailDropper.java#L30-L37

J-Dill:
I was able to recreate this problem by putting the Wood Rail Dropper on top of a dropper: only the first stack of items was pulled from the hopper. 

My above fix allows the Wood Rail Dropper to search through the rest of the slots in the Minecart with a Hopper.

J-Dill:
Also, I noticed that the Wood Rail Dropper allows items to be moved much faster than a hopper can actually move items. Is this intended?

J-Dill:
@bonii-xx Should this issue be closed out then? Or is there a reason to still keep it open?

SweetSugarSugar:
@j-dill Issues are kept open until a release is pushed

PR #3795, opened 2019-03-23
Fixed issue #3730

J-Dill:
Wood Rail Dropper was only allowing the first slot of Minecart With Hopper to be emptied.


src/main/java/slimeknights/tconstruct/gadgets/block/BlockWoodRailDropper.java
```diff
@@ -29,6 +29,8 @@ public void onMinecartPass(World world, EntityMinecart cart, BlockPos pos) {
     assert itemHandlerCart != null;
     for(int i = 0; i < itemHandlerCart.getSlots(); i++) {
       ItemStack itemStack = itemHandlerCart.extractItem(i, 1, true);
+      if(itemStack.isEmpty()) // If current slot is empty, move on to the next slot.
```
bonii-xx: Add brackets, remove superfluous comment

J-Dill:
There we go, I believe that is what you were looking for :)

bonii-xx:
Thanks for the fix, totally forgot about that issue

PR #3834, opened 2019-04-21
Update and enhancements french language!

Neerwan:


SweetSugarSugar:
Hi!  I was the one who wrote part of the original translation.  I looked through it briefly, and saw that you put exclaimation points after a space.  I am a student so I have no idea how that is typed but I don't think there should be spaces.   Also, I saw that you capitalized "slime" in Slimeleaf.   I believe it shouldn't be capitlized as slime is a descriptor/adjective.

Thanks for taking time to update it!  I haven't really gotten around to doing so.... ^^;

Neerwan:
Hello, in French, spaces before and after exclamation points are mandatory, see here: https://leconjugueur.lefigaro.fr/frponctuationtypographie.php

At least it's the case in metropolitan France.

For the smile, as we left it in original langage, perhaps it should be capitalized. This can be changed if you think it's incorrect.

Thanks for your feedback!

Neerwan:
Here are the screenshots for the Slime thing: https://i.imgur.com/G6QMpVY.png ; https://i.imgur.com/DKzXXAr.png

PR #3953, opened 2019-09-22
Feature: Ender Ores (Enderitum, Titanium)

spnda:
First of all: *I know that boni said that suggestions are not taken and features will only be added if there is any demand. Though the End has always been quite empty and I think it could have some Ores added to it. It only took me 1 day of playing Tinker's Construct to get a Manyullyn/Ardite Cleaver which I felt was quite easy to do. Maybe also lowering the default rarity of Ardite and Cobalt would be a good idea.
Furthermore boni also said that Copper, Aluminium etc. were removed as many other mods have these Ores. As far as I know there are only very few mods adding new Ores into the End.
That's why I added **Enderitum** and **Titanium**. They would add a new reason to go to the End and will be able to make even more advanced weapons with a little longer searching for the ores.*

**tldr**: The End has been empty and I feel it needs a few additions, Enderitum and Titanium.

## Enderitum 
It can be molten and used as a Material. It has Attack Damage worse than Manyullyn but better than Cobalt. Its durability is somewhere between Cobalt and Ardite and the mining speed is slightly faster than Cobalt. Enderitum has the "Alien" Trait. It is a epic material and is roughly about as rare as Diamonds.

Alien currently works as a trait for Enderitum but it could use a second Trait. I wasn't able to come up with a good Trait, sadly.

## Titanium
It can be molten and used as a Material. It has better Attack Damage than Manyullyn and also has only slightly worse Durability than Ardite. The Mining speed of Titanium is slightly better than Enderitum. Titanium is also an epic material and is very rare. 

I am still trying to find an applicable trait which Titanium could use. I thought of something like Resistance or Sharpness and maybe as a Modifier Silk Touch.

## Other Changes
- The Harvest Level of Manyullyn has been bumped to Enderitum. Because of that, Titanium and Enderitum can only be mined using a Manyullyn pickaxe/hammer.
- (Code Change) I've added BlockOre2 (tconstruct:ore_2) which is the same as BlockOre (tconstruct:ore) but it has a bumped Harvest Level.
- Added EndOreGenerator to generate Enderitum and Titanium.

## Preview
image

## Note
These Textures were fully made by myself.
If anyone has any suggestions regarding naming, textures, balancing, code and **especially** the traits please let me know. Any suggestions are welcome.
I am also thinking of making Slime Island versions for the End... or maybe other dimensions?


SweetSugarSugar:
You're free to include this as an addon but the current balance is fine as-is.  The End has End Stone as a material.  Also there are a couple of common mods that already add materials to The End such as Draconic Evolution (with an addon).  You're free to implement these changes in your own addon but it likely won't be implemented into Tinkers Construct.

spnda:
I definitely will make my own Addon if this PR gets declined (which I would fully understand).
I just felt like Tinker's Construct could use some Ender Ores of its own.

bonii-xx:
Let me preface with that we wont add it for several reasons that go against TiCs design. I'd like to give some design insights there:

- All materials must have a unique trait
- The material must serve a purpose. It must not invalidate (or simply "be better") than other materials, at any point in the game. Why ever use Manyullyn when you can just use Titanium? So far it looks to only add power for powers sake.
- Adding ores for the sake of ores/metals - They don't have any other use than for tool creation. Although in the End that's not really a problem, so it's a thing that might be violated. But it's always a great opportunity to get creative.
- I actually bumped the spawn rate of nether ores in the rework. The materials need to be available enough for repairing etc. It may be difficult to find, but it should not be hard to amass then (that's why a single purple slime island is more than enough for knightslime tools and keeping them intact)
- Titanium will clash with other mods. Definitively needs another name.


In general, we might add some endgame style End material, now that there's more End content and accessibility in Vanilla too. Scavenging for material on the End islands might be fun. But this will only happen when we have a niche to fill, and it fits.

bonii-xx:
As an extra note, the colors you chose are hard to read. Enderitium is purple, similar to Manyullyn, and Titanium looks a lot like iron.

KnightMiner:
One thing that should be mentioned is that for feature PRs, you really should talk to one of us on Discord first so we can discuss the design and let you know if it would fit with our overall vision. That being said, I really do like the idea of an end ore, its a pretty unique place to find an ore material.

I am not a huge fan of this implementation, mostly for the things boni said but I have a couple opinions to add:
* I am not a fan of the paradigm in mods of adding new materials with a tougher mining level, its just not fun to have to tier though all the metals just to mine the one you care about. Vanilla is not bad as its really easy to get up to iron; many mods (including Tinkers) push this a bit with blocks needing obsidian tools. Making Enderitum only minable with manyullyn is too much as now you have an ore with a mining level that only our tools mine, will get really annoying if a pack has Tinkers and you don't use it.
* I want to second what boni said about unique traits, if its the same as another material except with better stats, why would you use the old one anymore? you lose the Tinkering aspect to the mod when one choice is clearly better (especially a problem when end stone found in the same place as Enderitum). I suggest coming up with a unique trait first, one that simply is powerful enough that an end requirement makes sense but does not make other materials useless.
* Tinkers Construct is not really designed to have an "early game" and a "late game". Sure, some materials are in general stronger than others, but we don't aim to have a complex progression; you still have a reason to use wood even when you have access to manyullyn.
* I would personally stick with just one ore in the end if its just for tools. If each of the ores had something besides tools thats one thing, but right now they are not very different (that is why we got rid of copper, tin and the rest, they were just iron with different stats). I am aware cobalt/ardite don't fit this vision, that is something on my list to deal with in the future: making cobalt and ardite have some uses outside of tools.

So overall I like the idea of an end ore, I think you just need to flesh out this idea a bit more. Of course, you are free to make an addon if you have a different design goal than us; my addon Tinkers Complement can serve as an example for the structure of a Tinkers Construct addon.

spnda:
Thanks boni and KnightMiner for responding. I understand what each of you said.
I'll for now work on my own Addon and see what I'll do there. I'll post more on the discord.

PR #3960, opened 2019-10-14
1.14 cleanup

bonii-xx:
PR for reviewing


progwml6:
Should we move the injected cleanup here up to mantle to make it useful for our other mods?

KnightMiner:
That would make sense, though since it has to be a static method its a bit annoying to do so (would probably want a static import to avoid having to type the utility class name for every field.

PR #3962, opened 2019-10-15
Registry rework

bonii-xx:
Change our util for registries to be a bit cleaner.
In theory we could just call .setRegistryName inline on each item.. but it looks cleaner like that.
Also the ItemBlock utils are very useful.

Ditched the list-variants since they don't make sense from a design standpoint.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/world/TinkerWorld.java
+++ b/src/main/java/slimeknights/tconstruct/world/TinkerWorld.java
@@ -27,7 +27,6 @@ import net.minecraftforge.fml.event.lifecycle.FMLCommonSetupEvent;
 import net.minecraftforge.fml.event.lifecycle.InterModEnqueueEvent;
 import net.minecraftforge.fml.event.lifecycle.InterModProcessEvent;
 import net.minecraftforge.registries.ForgeRegistries;
-import net.minecraftforge.registries.IForgeRegistry;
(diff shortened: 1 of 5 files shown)


src/main/java/slimeknights/tconstruct/world/TinkerWorld.java
```diff
@@ -126,113 +127,114 @@ public TinkerWorld() {
...
-    registerBlockItem(registry, purple_slime_vine_middle, tabWorld);
-    registerBlockItem(registry, purple_slime_vine_end, tabWorld);
-
-    registerBlockItem(registry, blue_slime_vine, tabWorld);
-    registerBlockItem(registry, blue_slime_vine_middle, tabWorld);
-    registerBlockItem(registry, blue_slime_vine_end, tabWorld);
+    SpawnEggItem blueSlimeSpawnEgg = new SpawnEggItem(WorldEntities.blue_slime_entity, 0x47eff5, 0xacfff4, (new Item.Properties()).group(ItemGroup.MISC));
+    registry.register(blueSlimeSpawnEgg, "blue_slime_spawn_egg");
+
+    registry.registerBlockItem(green_slime_dirt, tabWorld);
```
bonii-xx, line 176: Actually considering if I don't make it so that the tab also is saved inside the ItemRegistryAdapter

alexbegt (approved):

PR #3966, opened 2019-10-18
WIP Rework registry/item/block handling

bonii-xx:
WIP Move all blocks/items, break up pulse structure.
All items/blocks shall always be registered.


src/main/java/slimeknights/tconstruct/blocks/WorldBlocks.java
```diff
@@ -0,0 +1,217 @@
...
+  public static final SlimeBlock blood_slime = injected();
+  public static final SlimeBlock magma_slime = injected();
+  public static final SlimeBlock pink_slime = injected();
+
+  public static final CongealedSlimeBlock congealed_green_slime = injected();
+  public static final CongealedSlimeBlock congealed_blue_slime = injected();
+  public static final CongealedSlimeBlock congealed_purple_slime = injected();
+  public static final CongealedSlimeBlock congealed_blood_slime = injected();
+  public static final CongealedSlimeBlock congealed_magma_slime = injected();
+  public static final CongealedSlimeBlock congealed_pink_slime = injected();
```
KnightMiner, line 49: We also use congealed slime for gadgets in the slime channel recipe, would it make sense to move this to common?

You could argue the same for the slime blocks as they would be expected even apart from the world module.

src/main/java/slimeknights/tconstruct/blocks/CommonBlocks.java
```diff
@@ -0,0 +1,125 @@
...
+
+  /* Crafting related items */
+  public static final GroutBlock grout = injected();
+  public static final GraveyardSoilBlock graveyard_soil = injected();
+  public static final ConsecratedSoilBlock consecrated_soil = injected();
+  public static final SlimyMudBlock slimy_mud_green = injected();
+  public static final SlimyMudBlock slimy_mud_blue = injected();
+  public static final SlimyMudBlock slimy_mud_magma = injected();
+  public static final FirewoodBlock lavawood = injected();
+  public static final FirewoodBlock firewood = injected();
```
KnightMiner: These are only used by the tools module, I wonder if it would be worth moving them to ToolsBlocks once we port that module over.

Likewise grout is only used by the smeltery.

KnightMiner (approved):

src/main/java/slimeknights/tconstruct/blocks/WorldBlocks.java
bonii-xx: I classified the things by "is in the world/generated" as well as how they show up in the creative tabs. Might move them

src/main/java/slimeknights/tconstruct/blocks/CommonBlocks.java
bonii-xx: Same as with the slimeblocks

src/main/java/slimeknights/tconstruct/TConstruct.java
```diff
@@ -69,6 +83,19 @@ public TConstruct() {
...
+      eventBus.register(CommonBlocks.class);
+      eventBus.register(WorldBlocks.class);
+      eventBus.register(GadgetBlocks.class);
+      eventBus.register(DecorativeBlocks.class);
+      eventBus.register(CommonItems.class);
+      eventBus.register(WorldItems.class);
+      eventBus.register(GadgetItems.class);
+      eventBus.register(ToolItems.class);
+      eventBus.register(FoodItems.class);
+    }
```
KnightMiner: There are event bus subscriber annotation we can use instead of this if we wanted. Ensures it gets registered as soon as possible.

alexbegt (approved):
These changes look good 👍

PR #3985, opened 2019-11-15
Materials - a humble beginning

bonii-xx:
The very base of material infrastructure, including loading.


src/main/java/slimeknights/tconstruct/library/materials/Material.java
```diff
@@ -0,0 +1,201 @@
...
+  }
+
+  @Override
+  public Collection<ITrait> getAllTraits() {
+    return traitsByStats.values().stream()
+      .flatMap(Collection::stream)
+      .collect(Collectors.toList());
+  }
+
+  public void setShard(ItemStack stack) {
```
KnightMiner, line 164: Do we still need shards? I thought we don't have any parts that cost 0.5 material anymore.

src/main/java/slimeknights/tconstruct/library/materials/client/MaterialRenderInfo.java
```diff
@@ -0,0 +1,26 @@
+package slimeknights.tconstruct.library.materials.client;
+
+import net.minecraft.client.renderer.texture.TextureAtlasSprite;
+import net.minecraft.util.ResourceLocation;
+
+/**
+ * Determines the type of texture used for rendering a specific material
+ */
+public interface MaterialRenderInfo {
```
KnightMiner: IMaterialRenderInfo?

src/main/java/slimeknights/tconstruct/library/traits/ITrait.java
```diff
@@ -0,0 +1,194 @@
...
+   * @param wasCritical If the hit was a critical hit.
+   * @param wasHit      If the target was actually hit. False when the entity was still invulnerable, or prevented the damage because of some other reason.
+   */
+  void afterHit(ItemStack tool, LivingEntity player, LivingEntity target, float damageDealt, boolean wasCritical, boolean wasHit);
+
+  /** Called when the player holding the tool blocks an attack. */
+  void onBlock(ItemStack tool, PlayerEntity player, LivingHurtEvent event);
+
+  /** Called when the player holding the tool is damaged. Only called if the player is NOT blocking! onBlock is called in that case.*/
+  default void onPlayerHurt(ItemStack tool, PlayerEntity player, LivingEntity attacker, LivingHurtEvent event) {}
```
KnightMiner, line 135: Why are some methods default and some not? I'd almost make all methods default as not all traits use most methods

src/main/java/slimeknights/tconstruct/library/materials/Material.java
bonii-xx: Good point, I need to check where else it's used. It'd actually be great if we got rid of the "leftovers" part in tool building.

src/main/java/slimeknights/tconstruct/library/materials/client/MaterialRenderInfo.java
bonii-xx: Dang. I hate how minecraft/forge uses the IFoobar interface notation.. but I want to stay consistent... thanks

src/main/java/slimeknights/tconstruct/library/traits/ITrait.java
bonii-xx: Leftovers, haven't actually touched the class logic yet. Good point though. I need to check how much of this belongs into abstract/defaults. The only point there is a default is because it was a later addition, so it doesn't break compatibility.

src/main/java/slimeknights/tconstruct/library/materials/Material.java
KnightMiner: One case that is worth cosidering is using a >1 material piece (like a log) to craft a smaller part.
* We could just require every material have a part of <=1 and use that part as the "shard", but then we still have to deal with leftovers which we want to avoid.
* Another option would be requiring an exact match, e.g. logs can only be used if you want a part costing 4
* We could discard excess material. This would be the simpliest to code, but would cause questions of why a log gives only 1 tool rod.
* Finally, we could return multiple of the part. Math gets a bit weird if not an exact match, so maybe discard excess beyond a single use of the material? Overall probably the toughest to code.

Personally, I like giving multiple matches then discarding excess, but that would be complicated.

bonii-xx: Last option definitively in this case. Use what you can and discard the rest. In most cases you'll be using "1 worth" items anyway. And you can always craft logs/ironblocks etc. into their smaller parts.

src/main/java/slimeknights/tconstruct/library/materials/json/MaterialStatJsonWrapper.java
```diff
@@ -22,6 +25,11 @@ public MaterialStatJsonWrapper(List<BaseMaterialStatsJson> stats) {
     return stats;
   }
 
+  @Nullable
+  public MaterialId getMaterialId() {
+    return materialId != null ? new MaterialId(materialId) : null;
```
KnightMiner, line 30: Why not convert this in the constructor so we do not keep throwing away MaterialId objects every time this is called?

alexbegt (approved):

src/main/java/slimeknights/tconstruct/library/materials/json/MaterialStatJsonWrapper.java
bonii-xx: Because deserialization does not use the constructor, and I was too lazy to write a special deserializer for it.

PR #4006, opened 2020-01-12
BouncedPacket exploit fix

Rysaen:
BouncedPacket should check server-side if the player has the slime boots in the right equipment slot. Without any check, the server will set the player's fall distance to zero everytime it receives a BouncedPacket, and that's exploitable.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tools/common/network/BouncedPacket.java
+++ b/src/main/java/slimeknights/tconstruct/tools/common/network/BouncedPacket.java
@@ -21,8 +21,9 @@ public class BouncedPacket extends AbstractPacketThreadsafe {
 
   @Override
   public void handleServerSafe(NetHandlerPlayServer netHandler) {
-	if(netHandler.player.getItemStackFromSlot(EntityEquipmentSlot.FEET).getItem() instanceof ItemSlimeBoots)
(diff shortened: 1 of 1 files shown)


src/main/java/slimeknights/tconstruct/tools/common/network/BouncedPacket.java
```diff
@@ -19,7 +21,8 @@ public void handleClientSafe(NetHandlerPlayClient netHandler) {
 
   @Override
   public void handleServerSafe(NetHandlerPlayServer netHandler) {
-    netHandler.player.fallDistance = 0;
+	if(netHandler.player.getItemStackFromSlot(EntityEquipmentSlot.FEET).getItem() instanceof ItemSlimeBoots)
```
KnightMiner: Your indentation is way off, we use 2 space indentation. We also always use curly brackets for single if statements

Rysaen: Sorry, didn't know. Should be okay now.

bonii-xx:
Tbh if somebody goes all the way to fake it, he might as well ;)
But of course it should be done for completeness sake. Thanks for the fix.

PR #4007, opened 2020-01-12
Materials: can you get back Flamestring from Nature (for bowstring)?

ircmaan:
Some years ago there was light integration with mod Nature  - 3or4 Flamestrings (loot from Nature spiders) was used for bowstring with 1.2x durability.
Can you get it back? It will be great.
Thanx anyway


KnightMiner:
Why is this a pull request? That is so much more work than making an issue.

Tinkers currently does not add integration for other mods, so normally its left to the other mod too add integration.

PR #4062, opened 2020-03-17
Enable whitelisting of slime island world gen dimensions

### LINKED ISSUE #3132: whitelist dim ids for slime islands
i wanna make slime islands only generate in overworld. other mods like ultilty worlds create many dim ids and i dont want slime islands spawning in them.

would be nice thanks

### LINKED ISSUE #3519: [1.12.2] [Request] Slime Island Whitelist
I would like to ask if it is possible to add an option to turn the blacklist in the config into a white list. Using Mystcraft or RFTools dimension let spawn slime island in many dimension where I can't prevent spawns.

### LINKED ISSUE #3831: I'm either stupid or this is a major problem.
So uh, big oofy. Y'know the slime islands? They're kinda um... spawning everywhere. I don't mean they're filling the sky but... I mean they're literally in EVERY dimension other than the Nether and End. Literally every single one. Yes, even on THE MOON. I looked in the config and it said that the "islands only spawn in the overworld" rule is marked true. They're still spawning though. There does happen to be a blacklist, which should really be able to be switched to a whitelist, and the Nether and End are marked on it. I am NOT going to waste my time manually finding the IDs of over 30 dimensions for this. Any help would be appreciated. Thanky. ^^

**Versions: 2.12.0.135**
* Minecraft: 1.12.2
* Forge: 14.23.5.2768
* Mantle: 1.3.3.49
* Tinkers Construct: 2.12.0.135

I'm pretty sure this is not a cross-mod related issue, because it just seems that the config is not registering.

### LINKED ISSUE #3933: Add option to config to change the Slime Island Blacklist to be a Whitelist instead
title says it

reason: because in most packs it's much easier to say what dimensions you want the Islands to generate in than to say where you don't want them to generate.

in the default case it should just have dimension 0 in there. since that will break the least amount of mods you add afterwards that have dimensions (without having to worry about the config), like Advanced Rocketry for example.

CJKay:
This change allows the user to toggle whether the list of slime island
dimension identifiers represents a blacklist or a whitelist. This allows
users with a large number of (possibly generated) dimensions to limit
the dimensions in which slime island generation takes place to a smaller
number without manually enumerating each one.

The following changes to the mod configuration options have been made:

 - `slimeIslandBlacklist` has been renamed to `slimeIslandDimensions`
 - `slimeIslandDimensionsIsBlacklist` has been added

`slimeIslandDimensions` describes the list of dimensions in which slime
island generation will be whitelisted/blacklisted and defaults to the
old default value of `slimeIslandBlacklist`, which was `-1` and `1` (the
Nether and the End).

`slimeIslandDimensionsIsBlacklist` describes whether
`slimeIslandDimensions` represents a blacklist or a whitelist, and
defaults to `true` (blacklist). This maintains backwards compatibility
with the previous behaviour.

Resolves #3132, resolves #3519, resolves #3831 and resolves #3933.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/common/config/Config.java
+++ b/src/main/java/slimeknights/tconstruct/common/config/Config.java
@@ -73,7 +73,8 @@ public final class Config {
   public static boolean genIslandsInSuperflat = false;
   public static int slimeIslandsRate = 730; // Every x-th chunk will have a slime island. so 1 = every chunk, 100 = every 100th
   public static int magmaIslandsRate = 100; // Every x-th chunk will have a slime island. so 1 = every chunk, 100 = every 100th
-  public static int[] slimeIslandBlacklist = new int[]{-1, 1};
+  public static int[] slimeIslandDimensions = new int[]{-1, 1};
+  public static boolean slimeIslandDimensionsIsBlacklist = true;
   public static boolean slimeIslandsOnlyGenerateInSurfaceWorlds = true;
   public static boolean genCobalt = true;
   public static int cobaltRate = 20; // max. cobalt per chunk
@@ -271,9 +272,16 @@ public final class Config {
       magmaIslandsRate = prop.getInt();
       propOrder.add(prop.getName());
 
-      prop = configFile.get(cat, "slimeIslandBlacklist", slimeIslandBlacklist);
-      prop.setComment("Prevents generation of slime islands in the listed dimensions");
-      slimeIslandBlacklist = prop.getIntList();
+      configFile.renameProperty(cat, "slimeIslandBlacklist", "slimeIslandDimensions");
+
+      prop = configFile.get(cat, "slimeIslandDimensions", slimeIslandDimensions);
+      prop.setComment("List of dimensions in which to enable or disable generation of slime islands");
+      slimeIslandDimensions = prop.getIntList();
+      propOrder.add(prop.getName());
+
+      prop = configFile.get(cat, "slimeIslandDimensionsIsBlacklist", slimeIslandDimensionsIsBlacklist);
+      prop.setComment("Whether the list of slime island dimensions behaves as a blacklist or a whitelist");
+      slimeIslandDimensionsIsBlacklist = prop.getBoolean();
       propOrder.add(prop.getName());
 
       prop = configFile.get(cat, "slimeIslandsOnlyGenerateInSurfaceWorlds", slimeIslandsOnlyGenerateInSurfaceWorlds);
--- a/src/main/java/slimeknights/tconstruct/world/worldgen/SlimeIslandGenerator.java
+++ b/src/main/java/slimeknights/tconstruct/world/worldgen/SlimeIslandGenerator.java
@@ -103,12 +103,12 @@ public class SlimeIslandGenerator implements IWorldGenerator {
   }
 
   protected boolean shouldGenerateInDimension(int id) {
-    for(int dim : Config.slimeIslandBlacklist) {
+    for(int dim : Config.slimeIslandDimensions) {
       if(dim == id) {
-        return false;
+        return !Config.slimeIslandDimensionsIsBlacklist;
       }
     }
-    return true;
+    return Config.slimeIslandDimensionsIsBlacklist;
   }
 
   @Override


KnightMiner:
This is not going to get accepted without config migrations, since you renamed the property, you need to make sure old configs properly migrate over.

bonii-xx:
Also we probably still don't want to maintain all the logic without automated testing, which is not doable in 1.12 :L
This issue never was about actually implementing it, but the implications it has on maintenance.

CJKay:
> 
> 
> This is not going to get accepted without config migrations, since you renamed the property, you need to make sure old configs properly migrate over.

Done - I've adapted the new version to fall onto the new behaviour only if the user has not modified `slimeIslandBlacklist`.

> 
> 
> Also we probably still don't want to maintain all the logic without automated testing, which is not doable in 1.12 :L
> This issue never was about actually implementing it, but the implications it has on maintenance.

I don't think this carries a major maintenance cost - the logic is very simple, and the backwards-compatibility logic can be very easily cut out when the time comes.

As it stands, I believe the combined maintenance cost borne by all of this mod's users far outweighs the time it took me to write and validate this, and it was the frustration of having to enumerate all of my server's dimensions that drove me here in the first place. With four submitted issues regarding this already I would suggest it's worth considering even just to stop people submitting them.

src/main/java/slimeknights/tconstruct/common/config/Config.java
```diff
@@ -271,11 +272,30 @@ public static boolean syncConfig() {
       magmaIslandsRate = prop.getInt();
       propOrder.add(prop.getName());
 
+      int[] slimeIslandBlacklist = slimeIslandDimensions;
+
       prop = configFile.get(cat, "slimeIslandBlacklist", slimeIslandBlacklist);
-      prop.setComment("Prevents generation of slime islands in the listed dimensions");
+      prop.setComment("DEPRECATED: Prevents generation of slime islands in the listed dimensions (overrides `slimeIslandDimensions` and `slimeIslandDimensionsIsBlacklist` if modified)");
```
KnightMiner: Do not add deprecated properties, thats even worse than no migration. `configFile.renameProperty`

CJKay: Apologies, first time mod-modder - not familiar with the APIs.

bonii-xx:
> > This is not going to get accepted without config migrations, since you renamed the property, 
> > Also we probably still don't want to maintain all the logic without automated testing, which is not doable in 1.12 :L
> > This issue never was about actually implementing it, but the implications it has on maintenance.
> 
> I don't think this carries a major maintenance cost - the logic is very simple, and the backwards-compatibility logic can be very easily cut out when the time comes.

That's nice, but the thing still is that it's a change that, if just one ! is placed incorrectly, results in all the people out there not generating slime islands anymore. And the logic is simple, but not exactly readable.

> As it stands, I believe the combined maintenance cost borne by all of this mod's users far outweighs the time it took me to write and validate this, and it was the frustration of having to enumerate all of my server's dimensions that drove me here in the first place. With four submitted issues regarding this already I would suggest it's worth considering even just to stop people submitting them.

Your situation is the exception. I understand it's frustrating, but it's not something I'm risking unless it's really necessary. You can easily go ahead and blacklist the first few hundred dimensions and never have to care about it again.

Last but not least, we don't have any more planned releases for 1.12.2, since it's EOL for us. But we'll consider it  for 1.15.

CJKay:
> That's nice, but the thing still is that it's a change that, if just one ! is placed incorrectly, results in all the people out there not generating slime islands anymore. And the logic is simple, but not exactly readable.

I'm not quite sure I understand. Of course, if there is a misplaced negation in the logic then that would be a bug... but that's why I verified (manually, with all the possible permutations) that there are no misplaced negations. There is, after all, just one, and I assure you it is in the correct place! 🙂

> Your situation is the exception. I understand it's frustrating, but it's not something I'm risking unless it's really necessary. You can easily go ahead and blacklist the first few hundred dimensions and never have to care about it again.

I don't think there is enough evidence to suggest my situation is the _exception_. It's certainly a behaviour of the mod that I simply learned to live with until recently (and prior to that simply didn't know how to work around). I don't imagine I am the first to wish slime islands were not spawning in my RFTools Dimensions.

> Last but not least, we don't have any more planned releases for 1.12.2, since it's EOL for us. But we'll consider it for 1.15.

Fair enough, best of luck with your endeavours!

bonii-xx:
Sorry if that came across wrong, I have on doubt that your code is correct as is. But I have enough experience with MC worldgen code to know that these things break easily. And added config options increase the complexity of it, meaning it breaks even easier. Hence the extra maintenance. We can verify it now, but that doesn't mean it'll stay true if it's changed if any future update is necessary.

> I don't think there is enough evidence to suggest my situation is the exception.

Of course it's an assumption, but if I compare it to the number of downloads/player using TiC and RFTools together, it most likely is.

> I don't imagine I am the first to wish slime islands were not spawning in my RFTools Dimensions.

You're not. And you wont be the last. But the behaviour should be that an RFTools overworld dimension behaves the same as the actual overworld. If people don't want it for specific RFTools dimensions.. they have to blacklist it. And blacklisting the first 100 dimensions is quickly done if you want to do it in advance.

Thank you for your understanding, I know it's not the optimal solution, but it's the best way to keep 1.12 as stable as possible.

DarianLStephens:
It has been around a year since this was proposed, with only two adjustment in the form of the added tags.
Reading through the relevant issues, there has always been significant pushback against it by the author, when it would solve very many issues and introduce relatively little complexity.

This issue was a factor in me considering simply not adding Tinkers to the server, and with how many dimensions I've had to blacklist (Currently 25 and climbing), I'm at the point where I'm considering just turning off slime island generation altogether for the hassle they cause.
After all, some are already generated in the Overworld, people can just go to those if they need the things they offer.

KnightMiner:
@DarianLStephens did you even read the comments here? 1.12 is end of life, we are not adding more features, and honestly I doubt it's getting another build at this point. The only reason this is still open is a reminder to set up config for 1.16 properly

DarianLStephens:
It doesn't detract from stability, because one of the arguments against the feature was that it would introduce unneeded complexity with regards to updates.
However, if 1.12 is no longer being maintained, then there's no need to worry about having to keep the feature updated, as it only needs to go in once.
After all, if you plan to add it to the 1.16 version, then it's not a question of it being difficult, and no longer of people being unwilling to add it.

This is a small feature with major implications for pack making.
However, if this never makes it in to any 1.12 version, official or otherwise, I will simply disable the slime islands, because they provide very little actual benefit beyond trivializing slime.

KnightMiner:
1.12 is no longer being updated. Meaning first off there is no update to add this in, and second off there is no update to fix issues caused by this change.

The threat of disabling slime islands is not going to make me add this. Turn them off if you must.

CJKay:
@DarianLStephens if you really want this, I'd advise just cloning my 1.12 branch and building that.

KnightMiner:
Closing this as we decided to no longer release more 1.12 builds.

PR #4087, opened 2020-04-17
1.15: I want to help

hanetzer:
As I'm not getting much of a response on the discord regarding the asking there before
a pr, figured I'd do it here. This is just a small commit, with more to follow. I've added (so
far) 89 recipes from the 1.12 version to this branch (currently uncommitted) and would
like to get some kind of feedback going forward.

I have a number of ideas/questions I'd like to about. For instance, 1.15 (maybe before)
added a new stonecutter tool which makes 1:1 block:stair conversions possible, and it
can transform one kind of stoneish block into another (for instance it can turn sandstone
directly into cut, chiseled, slab, cut slab, stair, and wall in 1:1 ratios); should we add similar
functionality to the seared family of blocks? Can be done.

Same for the new blast furnace, though ideally one would smelt in a smeltery, having a
blast furnace for the earlygame is useful. Ardite and Cobalt ore can be added to that
recipe table, it just smelts twice as fast, no other benefits.

What are your guy's thoughts on this?


KnightMiner:
I think stone cutter makes sense for swapping seared and brownstone variants, the current recipe was just added so it has a recipe but its really not great. Of course, that is only for types that are "the same", so things like cracked bricks would keep their furnace recipe.

Cobalt and Ardite might make sense as a blast furnace only recipe at normal speed, not not sure. One of the biggest reasons a lot of stuff is not enabled is we are still reworking how modules are done, and working on migrating recipes to data generators.

hanetzer:
> I think stone cutter makes sense for swapping seared and brownstone variants,
> the current recipe was just added so it has a recipe but its really not great. Of
> course, that is only for types that are "the same", so things like cracked bricks
> would keep their furnace recipe.
> 
I'll be honest, I never once dealt with brownstone in the 1.12 version, so I'm not
sure what you mean here; the redstone+sandstone recipe in 1.12?
Yeah, So all the ones that are just 1:1 conversion in the grid would make sense,
but the ones that require a pass through the furnace should stay there. But, should
a recipe that takes you straight from tconstruct:seared_stone to tconstruct:seared_road_stairs
be added?
> Cobalt and Ardite might make sense as a blast furnace only recipe at normal
> speed, not not sure. One of the biggest reasons a lot of stuff is not enabled is
> we are still reworking how modules are done, and working on migrating recipes
> to data generators.
As in, you cannot smelt cobalt or ardite in a normal cobblestone-ring furnace but
has to be smelted in either a blast furnace or the full smeltery, and in the blast
furnace it should be as slow as smelting say iron ore in a normal furnace?

What do you mean, recipe to data generator?

Also, how granular should I make the commits for the recipe json files? Like, class
of item (ores, seared stuff, slime stuff) per commit? (obviously 'real' code changes
should be granular).

hanetzer:
@alexbegt hey, I see you're doing a lot of work on this as well. Do you think it would
be too much issue to get the 'plain data' into the 1.15 branch? I was just checking out
your tables branch and realized you had quite a lot of recipes already done and I'd like
to avoid duplicate work if possible :)

src/main/java/slimeknights/tconstruct/blocks/BlockProperties.java
```diff
@@ -54,8 +54,8 @@
   static Block.Properties PUNJI = builder(Material.PLANTS, NO_TOOL, SoundType.PLANT).hardnessAndResistance(3.0F);
 
   // SMELTERY
-  static Block.Properties SMELTERY = builder(Material.ROCK, NO_TOOL, SoundType.METAL).hardnessAndResistance(3.0F, 20.0F);
-  static Block.Properties SMELTERY_GLASS = builder(Material.ROCK, NO_TOOL, SoundType.METAL).hardnessAndResistance(3.0F, 20.0F).notSolid();
+  static Block.Properties SMELTERY = builder(Material.ROCK, NO_TOOL, SoundType.METAL).hardnessAndResistance(3.0F, 9.0F);
```
bonii-xx, line 57: Why would you change the resistance? Their values are intentional.

bonii-xx:
Help is generally welcome, but try to get a hold of knightminer or me before starting doing stuff, otherwise you're risking doing work that wont be used. As knightminer already said, several things are still being figured out with regard to how we want to tackle things in 1.15.

As for adding things, there wont be anything new added anytime soon. The main goal is to get the main mod up and working. Additional things that aren't part of core TiC are better off as an addon. TiC itself uses the same codebase as any addon would.

src/main/java/slimeknights/tconstruct/blocks/BlockProperties.java
hanetzer: Reason is in the commit message. 1.15's blocks are apparently 100% impervious (or
close to it) to tnt, whereas 1.12 blocks react to tnt about the same as 1.15's end brick.

hanetzer:
> Help is generally welcome, but try to get a hold of knightminer or me before starting
> doing stuff, otherwise you're risking doing work that wont be used. As knightminer
> already said, several things are still being figured out with regard to how we want to
> tackle things in 1.15.
> 
> As for adding things, there wont be anything new added anytime soon. The main goal
> is to get the main mod up and working. Additional things that aren't part of core TiC
> are better off as an addon. TiC itself uses the same codebase as any addon would.

Agreed on the adding things bit, but there are some things which are really simple to do
and would make the mod 'fit better' in 1.15, such as stonecutter recipes (just a few json
files), very low hanging fruit. As mentioned by myself and @KnightMiner earlier in this
thread the handling of the blast furnace wrt ardite/cobalt is an interesting topic.

hanetzer:
So I had an idea which may simplify the formation of smeltery and friends. Instead of a
hard-coded list of blocks in SmelteryBlocks.java, how about we use the new tagging
system to add something like "tconstruct:smeltery_valid" and so on and use the lookup
to determine it? Then mods which extend tconstruct can add their own blocks to the valid
list by using said tags.

bonii-xx:
The list being hardcoded is intentional. We don't want other blocks to be usable by design (and to prevent complications)

hanetzer:
> The list being hardcoded is intentional. We don't want other blocks to be usable by design
> (and to prevent complications)

Fair enough. Just spitballing ideas.
FTR, seared stone should only be able to come from casting the liquid, seared cobble
should only be from casting the liquid over cobble, and the cracked seared bricks should
only come from some form of smelting of seared bricks, even with all the new methods
in 1.15 to change blocks? Just making sure.

bonii-xx:
Seared stone from casting, yes. The others.. I think we introduced crafting recipes to get them easier?
Anything -> cracked by furnace is correct as well.

KnightMiner:
Our crafting recipe is just "brick type 1" -> "brick type 2". I personally think we should ditch all thoser recipes except the vanilla ones (stone -> bricks, or maybe stone -> paver to be like granite/etc.), then use the stone cutter for the fancy patterns.

bonii-xx:
Oh, yeah. I keep forgetting that that's a vanilla thing now.. I agree

hanetzer:
I can do that. Who's in charge here who can tell me to do it definitely do it?

hanetzer:
> Our crafting recipe is just "brick type 1" -> "brick type 2". I personally think we should
> ditch all thoser recipes except the vanilla ones (stone -> bricks, or maybe stone ->
> paver to be like granite/etc.), then use the stone cutter for the fancy patterns.

So like, starting from plain seared stone you can go stone->paver->bricks, and you can
do the seared brick (ingot  type) ->seared bricks and standard stair/slab recipes 'by hand'
but you need a stonecutter to get anything that isn't plain bricks or a paver?

KnightMiner:
Me and boni are in charge here, alexbegt is also an official voice.

----

Yeah, I say support
* stone -> paver -> bricks in 2x2 crafting, like vanilla stone bricks recipe
* seared brick item -> seared bricks block, like clay bricks recipe
* two brick slabs -> square bricks (chiseled), like the vanilla chiseled brick recipe
* cobble -> stone and bricks -> cracked bricks in a furnace
* stone to any of paver, brick, brick_fancy, brick_square, brick_triangle, brick_small, road, tile, and creeper in the stone cutter.
* changing any of the bricks from the previous list into each other using the stone cutter, should be easy using a tag  containing that list. None of stone, cobble, or cracked bricks can be carved in the stone cutter.
* All blocks will support normal stair and slab crafting among the same texture. Both in the crafting table and in the stone cutter

hanetzer:
> Me and boni are in charge here, alexbegt is also an official voice.
> 
> Yeah, I say support
> 
>     * stone -> paver -> bricks in 2x2 crafting, like vanilla stone bricks recipe
> 
Easy
>     * seared brick item -> seared bricks block, like clay bricks recipe
> 
Easy
>     * two brick slabs -> square bricks (chiseled), like the vanilla chiseled brick recipe
> 
As in, `tconstruct:seared_bricks_slab` on top of one another
>     * cobble -> stone and bricks -> cracked bricks in a furnace
> 
I don't quite understand what you mean here. Is not the seared_cobble recipe
dependent on smeltery (seared stone poured over cobble in basin)?
>     * stone to any of paver, brick, brick_fancy, brick_square, brick_triangle, brick_small, road, tile, and creeper in the stone cutter.
> 
>     * changing any of the bricks from the previous list into each other using the stone cutter, should be easy using a tag  containing that list. None of stone, cobble, or cracked bricks can be carved in the stone cutter.
> 
As in, no stonecutter recipes take those as input, or you cannot create them in the stonecutter?
>     * All blocks will support normal stair and slab crafting among the same texture. Both in the crafting table and in the stone cutter
> 
So you mean, no going straight from seared_stone to seared_road_slab?

KnightMiner:
Seared cobble is made from seared stone on cobble, but it should also smelt into seared stone for consistency with vanilla cobble. Likewise, bricks smelt into cracked bricks.

Stonecutter does not output stone, cobble, or cracked bricks. Cobble and cracked bricks are not input options.

Yes, no direct from seared stone to seared road slabs.

hanetzer:
> Seared cobble is made from seared stone on cobble, but it should also smelt into seared stone for consistency with vanilla cobble. Likewise, bricks smelt into cracked bricks.
>
Oh, that's interesting. so seared_cobble->furnace->seared_stone just like cobble->furnace->stone,
gotcha.
> Stonecutter does not output stone, cobble, or cracked bricks. Cobble and cracked bricks are not input options.
> 
Ok, figured the first, but as cobble and cracked bricks have slab and stairs variants I figured it would
be sensible to add a stonecutter recipe.
> Yes, no direct from seared stone to seared road slabs.
> 
Ok so I can use stonecutter to turn seared_stone into all the other non-cracked, non-cobble
blocks, but the only way to get seared_bricks_stairs from it is to put in seared_bricks.

KnightMiner:
I meant don't accept them as input for the texture recipe, so you cannot use the stone cutter to turn cobble into bricks. They still have their own recipes for stairs and slabs, separate recipes.

hanetzer:
Ah yeah. that's what I was thinking. I'll prolly have to redo the earlier commit which
added the seared recipes to take in account the new ideas for recipes.

bonii-xx:
> > Seared cobble is made from seared stone on cobble, but it should also smelt into seared stone for consistency with vanilla cobble. Likewise, bricks smelt into cracked bricks.
> 
> Oh, that's interesting. so seared_cobble->furnace->seared_stone just like cobble->furnace->stone,
> gotcha.

It's simply conforming to player expectations. Way too many mods try to be complicated. ;)

PR #4096, opened 2020-04-28
Improve performance of smeltery

ImMorpheus:


KnightMiner:
A single recipe cache will barely help, the smeltery has more than one slot, so there is a potential for a different recipe in every slot. Essentially, this would fix the case of an iron block sitting in your smeltery unable to melt causing lag, but as soon as you add an iron ingot it will start to lag again.

We need a solution that takes into account that the smeltery can be working on multiple recipes at the same time.

----

Perhaps a better solution is to cache the recipe per slot in `updateHeatRequired`, then ideally you can just pull the already fetched recipe during melting and save more efficiency. Would even have a small benefit if you keep changing items, as it halves the number of recipe fetches. Would need to make sure it handles an item partway through melting on world reload of course.

ImMorpheus:
> the smeltery has more than one slot, so there is a potential for a different recipe in every slot. 

True, but that's the worst case scenario. Often players would just dump a stack of iron (for example) ores in the smeltery. 

> this would fix the case of an iron block sitting in your smeltery unable to melt causing lag, but as soon as you add an iron ingot it will start to lag again.

Not really, the method is called as soon as the `itemTemperature` is more than the `itemTempRequired` and for each slot.
To negate the performance improvement of the cache you would have to:
1) Melt one item
2) Melt more than one item but place them in such a way that an item is always different from the one in the previous slot.

This means if you try to melt 64 iron ore, you would only query the recipe registry once.

PrincessOfEvil:
1) a pretty common use case is to use the smeltery for alloys.
2) would probably be better to have a tiny *data structure* of a cache, with size in single digits. Having a recipe per slot *is* uncommon.

KnightMiner:
Yes, smelting 64 iron ore is a usecase of the smeltery, but not the only one. I personally will often dump all my ores from mining into the smeltery as I tend to store my metals in liquid form, and if the mod I help produce treats my playstyle as illegitimate, something is wrong there.

I do agree that a recipe per slot is rather uncommon, though it would be simple to implement and its just a small amount of RAM usage. Of course, if you hook a hopper up and are rapidly changing items that would not help much. Keeping a cache of the last 5 recipes might be sufficient instead, I'm pretty sure Java has some sort of priority queue that would handle it. Likewise, to make such a cache useful, it should be used for the temperature, not just the melting result.

PrincessOfEvil:
Nah, I never seen people just dump all ores, honestly. That's just asking
for random alloys.

>

KnightMiner:
Not when your smeltery immediately ejects fluids to a Tinker Tank. The time between ores finishing is enough to prevent alloys.

bonii-xx:
No further changes to the smeltery will be done on 1.12 to ensure stability. We'll tackle the smeltery in 1.15+, probably by using a more composite approach.

PR #4104, opened 2020-05-09
smeltery work

hanetzer:
Ok. Consider this a continuation of #4087, mentioning it here for click purposes.
Not ready for merge. Some commits are, and some should never be merged at all,
as they're just there for my own help in debugging.


hanetzer (commented):
Screw the list. Inline comments it is.

PR #4128, opened 2020-06-09
Fix invalid client side inventory when opening Part Chest or Pattern …

shedaniel:
…Chest directly


shedaniel:
It looks like you can't interact with the inventory once you fill the first 3x8 inventory, but that should be a gui issue and I will look into that later.

shedaniel:
Fixed that as well.

PR #4131, opened 2020-06-12
smeltery: add working faucet

hanetzer:
Seems to mostly be in order. Asked @shedaniel to add the pretty stuff to it,
feel free to poke holes in stuff I forgot ;)


src/generated/resources/data/tconstruct/advancements/recipes/tinkers_smeltery/smeltery/faucet.json
```diff
@@ -0,0 +1,32 @@
+{
```
KnightMiner, line 1: Missing cache in commit for these new resources

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
```diff
@@ -66,6 +68,7 @@
   public static final BuildingBlockObject searedRoad = BLOCKS.registerBuilding("seared_road", SMELTERY, TOOLTIP_BLOCK_ITEM);
   public static final BuildingBlockObject searedTile = BLOCKS.registerBuilding("seared_tile", SMELTERY, TOOLTIP_BLOCK_ITEM);
   public static final EnumObject<SearedTankBlock.TankType,SearedTankBlock> searedTank = BLOCKS.registerEnum("seared", SearedTankBlock.TankType.values(), (type) -> new SearedTankBlock(SMELTERY_GLASS), (b) -> new TankItem(b, SMELTERY_PROPS));
+  public static final BlockItemObject<FaucetBlock> searedFaucet = BLOCKS.register("faucet", () -> new FaucetBlock(SMELTERY.notSolid()), TOOLTIP_BLOCK_ITEM);
```
KnightMiner: `SMELTERY` is expected to be final, you should not be modifying it. Make a new property for the faucet. While properties are not stored in the block, its better to assume that each variable should be unique.

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
```diff
@@ -0,0 +1,122 @@
...
+import slimeknights.tconstruct.smeltery.tileentity.FaucetTileEntity;
+
+import javax.annotation.Nullable;
+import java.util.Random;
+
+public class FaucetBlock extends ContainerBlock {
+  public static final DirectionProperty FACING = DirectionProperty.create("facing", (direction) -> {
+    return direction != Direction.DOWN;
+  });
+  protected static final VoxelShape UP_SHAPE = Block.makeCuboidShape(4.0D, 10.0D, 4.0D, 12.0D, 16.0D, 12.0D);
```
KnightMiner: Make an `EnumMap` with these. Can use a `static {}` block to initialize.

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
```diff
@@ -0,0 +1,122 @@
...
+
+  @Nullable
+  @Override
+  public BlockState getStateForPlacement(BlockItemUseContext context) {
+    Direction dir = context.getFace().getOpposite();
+    if (dir == Direction.DOWN) {
+      dir = context.getPlacementHorizontalFacing().getOpposite();
+    }
+    return this.getDefaultState().with(FACING, dir);
+    //return this.getDefaultState().with(FACING, dir.getAxis() == Direction.Axis.Y ? Direction.UP : dir);
```
KnightMiner: Remove this since you handed it above.

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
```diff
@@ -0,0 +1,122 @@
...
+  }
+
+  protected void fillStateContainer(StateContainer.Builder<Block, BlockState> builder) {
```
KnightMiner, line 110: Missing `@Override`, if you are using IDEA, there is a save actions extension that will add it automatically you can use.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
```diff
@@ -0,0 +1,231 @@
...
+      reset();
+    }
+  }
+
+  protected void reset() {
+    isPouring = false;
+    stopPouring = false;
+    drained = null;
+    direction = Direction.DOWN;
+    lastRedstoneState = false;
```
KnightMiner: Remove this last line, it led to a bug in 1.12 with block updates. We never patched it because it became expected behaior.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
```diff
@@ -0,0 +1,231 @@
...
+      isPouring = true;
+      direction = Direction.values()[compound.getInt("direction")];
+      stopPouring = compound.getBoolean("stop");
+    }
+    else {
+      reset();
+    }
+  }
+
+  public void onActivationPacket(FluidStack fluid) {
```
KnightMiner, line 233: I know we do not have a ton of java docs in 1.12, but ideally we should have them. My rule of thumb is each method should have either a javadoc or an `@Override`

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
```diff
@@ -0,0 +1,231 @@
...
+        int filled = toFill.fill(drained, IFluidHandler.FluidAction.SIMULATE);
+        if (filled > 0) {
+          // drain the liquid and transfer it, buffer the amount for delay
+          this.drained = toDrain.drain(filled, IFluidHandler.FluidAction.EXECUTE);
+          this.isPouring = true;
+          pour();
+
+          // sync to clients
+          if (!getWorld().isRemote()) { // TODO: getWorld() instanceof WorldServer
+            // TODO: TinkerNetwork.sendToClients((WorldServer) getWorld(), pos, new FaucetActivationPacket(pos, drained));
```
KnightMiner: Why is this commented out? You probably want `ServerWorld`

KnightMiner (changes requested):
Some minor changes needed. I can add the TESR after merging if needed.

src/generated/resources/data/tconstruct/advancements/recipes/tinkers_smeltery/smeltery/faucet.json
hanetzer: Yeah I don't generate cache until its close to being done. that file is a nightmare to
deal with in rebasing.

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
hanetzer: Sure.

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
hanetzer: Fair enough. I basically did an inverted hopper thing.

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
hanetzer: Ah yeah. this was a note to self type thing.

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
hanetzer: Huh. Neat. Lemme look into that. You are talking about fillStateContainer, correct?

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
hanetzer: aight, and its now time we can clean slate it.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
hanetzer: Ah, because I wasn't sure at the time. Do we currently have a faucet activation
packet, or do I need to write it?

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
alexbegt: Yes, fillStateContainer is missing the `@Override`

src/generated/resources/data/tconstruct/advancements/recipes/tinkers_smeltery/smeltery/faucet.json
KnightMiner: It should just automatically merge without conflicts. But then again, that is cache.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
KnightMiner: I have some plans for a proper behavior version of that basically. Changing what stops the faucet from pouring

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
KnightMiner: We had one in 1.12. You need to port it.

src/generated/resources/data/tconstruct/advancements/recipes/tinkers_smeltery/smeltery/faucet.json
hanetzer: 'should' and 'does' are not always the same. Like I said, once everything else is approved I'll regen it clean.

src/main/java/slimeknights/tconstruct/smeltery/block/FaucetBlock.java
hanetzer: Think I've handled this in a satisfactory way. Its modeled after how WallTorchBlock
is handled.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
hanetzer: Think I did this right, however, faucet'ing fluids between two tanks
will spam the logs with:
```[Render thread/WARN] [minecraft/ClientPlayNetHandler]: Unknown custom packet identifier: minecraft:tconstruct```
I am probably missing something simple.

Comment (hanetzer) [reply src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java]: looks like I have to setPacketHandled.

File: src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java (hunk 2 above)
Comment (hanetzer) [reply src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java]: Is something akin to
```java
private static final Block.Properties FAUCET = SMELTERY.notSolid();
```
valid in this case? Idea doesn't seem to care either way.

hanetzer: Screw it. I'll just effectively dupe the property, like we do for glass. Would use glass,
but even though they're mechanically identical, semantically they are not, and they
may diverge over time.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
hanetzer: Considering to do something akin to FluidUpdatePacket, make it an interface
and document it there. Sound fine?

KnightMiner: If we use it in multiple places, make an interface. I doubt the faucet packet will be used again elsewhere.

src/generated/resources/data/tconstruct/advancements/recipes/tinkers_smeltery/smeltery/faucet.json
KnightMiner: Looks like the faucet comments is the last thing. When you push those javadocs, also push this file and it should be ready to merge. I have an idea of how to do the faucet fluid using a model TESR hybrid, so skip that for now

src/main/java/slimeknights/tconstruct/smeltery/tileentity/FaucetTileEntity.java
hanetzer: Think we're good there. If needed we can factor it out later.

src/generated/resources/data/tconstruct/advancements/recipes/tinkers_smeltery/smeltery/faucet.json
hanetzer: Done.

hanetzer:
Alrighty, shedaniel hooked us up with a good enough for now working 
TE renderer. Ready for merge whenever :)

hanetzer:
~Hang on. Still have some fluidstack == null/ fluidstack != null checks which should be isEmpty~
Actually nevermind. switching null checks to isEmpty checks makes it stop working after 144mb

KnightMiner (approved):

PR #4132, opened 2020-06-13
potential platform datagen issue

hanetzer:
This commit is based upon a clean checkout of commit 59e4cc5d87560a77c349203c6f93a054c242ca4f,
with no local changes. Running `runData` on my system (gentoo linux ~amd64) always produces this
rearrangement of the slimeball tags. I've been removing this from my commits in order to prevent
spurious code churn where I'm not actually changing anything. Do you think this may be an issue
we should bring up to forge upstream?


KnightMiner:
Me running datagen does not move that, appears to be a system sorting issue. Might be worth asking about on the forge discord.

I really want more details on what that file does mostly, Forge commits it so it does something useful, but it causes a lot of problems.

hanetzer:
it appears to just be a sorted (by filename) md5 filename list.

KnightMiner:
Closing as this is fixed in Forge

PR #4134, opened 2020-06-19
casting: This is not a drill.

hanetzer:


src/main/java/slimeknights/tconstruct/library/registration/RecipeSerializerDeferredRegister.java
```diff
@@ -0,0 +1,34 @@
...
+
+import net.minecraft.item.crafting.IRecipeSerializer;
+import net.minecraftforge.eventbus.api.IEventBus;
+import net.minecraftforge.registries.DeferredRegister;
+import net.minecraftforge.registries.ForgeRegistries;
+import slimeknights.tconstruct.library.registration.object.RecipeSerializerObject;
+
+import java.util.function.Supplier;
+
+public class RecipeSerializerDeferredRegister extends RegisterWrapper<IRecipeSerializer<?>> {
```
KnightMiner: Delete this, all it does is apply the wrapper, which is unneeded wrapping. Use the base deferred register

src/main/java/slimeknights/tconstruct/library/registration/object/RecipeSerializerObject.java
```diff
@@ -0,0 +1,20 @@
+package slimeknights.tconstruct.library.registration.object;
+
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.item.crafting.IRecipeSerializer;
+import net.minecraft.util.ResourceLocation;
+
+import java.util.function.Supplier;
+
+public class RecipeSerializerObject<RECIPE extends IRecipeSerializer<?>> implements Supplier<RECIPE> {
```
KnightMiner: Delete this, `RegistryObject` can already do everything this wrapper can. I only add the wrappers to provide new useful methods, just for the sake of having them.

src/main/java/slimeknights/tconstruct/library/smeltery/IFaucetDepth.java
```diff
@@ -0,0 +1,10 @@
+package slimeknights.tconstruct.library.smeltery;
+
+/**
+ * Block interface to determine how far the seared faucet liquid flows into a block
+ */
+public interface IFaucetDepth {
```
KnightMiner: We removed this from the faucet, no need to bring it back. I want a proper solution using models, which you can ignore for this PR

src/main/java/slimeknights/tconstruct/plugin/jei/casting/CastingVariantCategory.java
```diff
@@ -0,0 +1,19 @@
+package slimeknights.tconstruct.plugin.jei.casting;
+
+import mezz.jei.api.gui.drawable.IDrawableAnimated;
+import mezz.jei.api.gui.drawable.IDrawableStatic;
+import mezz.jei.api.helpers.IGuiHelper;
+import mezz.jei.api.recipe.category.IRecipeCategory;
+
+public abstract class CastingVariantCategory<T> implements IRecipeCategory<T> {
```
KnightMiner: Merge this into AbstractCastingCategory, I am not sure why mezz made so many subclasses in the furnace.

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
```diff
@@ -92,12 +104,51 @@
...
 
   /*
    * Items
    */
   public static final ItemObject<Item> searedBrick = ITEMS.register("seared_brick", SMELTERY_PROPS);
+  public static final ItemObject<Item> blankCast = ITEMS.register("blank_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> ingotCast = ITEMS.register("ingot_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> nuggetCast = ITEMS.register("nugget_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> gemCast = ITEMS.register("gem_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> pickaxeHeadCast = ITEMS.register("pickaxe_head_cast", SMELTERY_PROPS);
```
KnightMiner, line 117: For now, would be better to add only casts for items we have, so pickaxe head, small binding, and tool rod. We will be removing some of the old tools and parts later and I would rather not need migrations on ForgeCraft.

You can leave the textures/models if you want, most we will need later. Just want to avoid adding items we do not need for now.

src/main/java/slimeknights/tconstruct/smeltery/block/CastingBlock.java
```diff
@@ -0,0 +1,47 @@
...
+import net.minecraft.util.math.BlockRayTraceResult;
+import net.minecraft.world.IBlockReader;
+import net.minecraft.world.World;
+import slimeknights.mantle.block.InventoryBlock;
+import slimeknights.tconstruct.library.smeltery.IFaucetDepth;
+import slimeknights.tconstruct.smeltery.tileentity.CastingTileEntity;
+
+import javax.annotation.Nonnull;
+
+public class CastingBlock extends InventoryBlock implements IFaucetDepth {
```
KnightMiner: Remove `IFaucetDepth`, interface was a poor solution.

KnightMiner: Make abstract

src/main/java/slimeknights/tconstruct/smeltery/block/CastingBlock.java
```diff
@@ -0,0 +1,47 @@
...
+
+  protected CastingBlock(Properties builder) {
+    super(builder);
+  }
+
+  @Nonnull
+  @Override
+  public TileEntity createTileEntity(BlockState blockState, IBlockReader iBlockReader) {
+    return null;
```
KnightMiner: Abstract method, I don't want an accidental null

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.World;
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.registries.ForgeRegistries;
+import net.minecraftforge.registries.ForgeRegistryEntry;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public abstract class AbstractCastingRecipe implements IRecipe<IInventory> {
```
KnightMiner: Use a custom interface instead of IInventory. Will let you include fluid in the interface and removes the need for two separate matches methods. The whole design of vanilla recipes is you can choose the container type, so we can add more arguments later.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+  protected final IRecipeType<?> type;
```
KnightMiner: Do we need to pass in type as a field? We could just override it in the subclass.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+  protected final ResourceLocation id;
+  protected final String group;
+  private final Ingredient cast;
+  private final FluidStack fluid;
+  private final ItemStack result;
+  protected final int coolingTime;
+  private final boolean consumed;
+  private final boolean switchSlots;
```
KnightMiner: Some javadoc comments on these fields would be good, explain what they do. Especially the booleans.

You can also use `@Getter` on these to have Lombok make the getter methods for you.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+  }
+
+  @Override
+  public ResourceLocation getId() {
+    return this.id;
+  }
+
+  @Override
+  public IRecipeSerializer<?> getSerializer() {
+    return null;
```
KnightMiner: Don't return null, make abstract. We must have a serializer.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+    public Serializer(Serializer.IFactory<T> factoryIn) {
+      this.factory = factoryIn;
+    }
+
+    @Override
+    public T read(ResourceLocation recipeId, JsonObject json) {
+      Ingredient cast = Ingredient.EMPTY;
+      String s = JSONUtils.getString(json, "group", "");
+      boolean consumed = false;
+      boolean switchSlots = JSONUtils.getBoolean(json, "switchslots", false);
```
KnightMiner: I suggest using `switch_slots`, more readible.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+        coolingtime = JSONUtils.getInt(json, "coolingtime");
+      }
+      return this.factory.create(recipeId, s, cast, fluidStack, item, coolingtime, consumed, switchSlots);
+    }
+
+    @Nullable
+    @Override
+    public T read(ResourceLocation recipeId, PacketBuffer buffer) {
+      // TODO: Magic number go away
+      String s = buffer.readString(32767);
```
KnightMiner: Max size of a short I think. The number is not magic, its just big. Can probably use `Short.MAX_VALUE`, but double check that value.

Also, don't name this `s`, name it `group`

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+
+  public FluidStack getFluidStack() {
+    return this.fluid;
+  }
+
+  public Fluid getFluid() {
+    return this.fluid.getFluid();
+  }
+
+  public Ingredient getCast() {
```
KnightMiner: Javadocs

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
...
+  public IRecipeSerializer<?> getSerializer() {
+    return null;
+  }
+
+  @Override
+  public IRecipeType<?> getType() {
+    return type;
+  }
+
+  public int getCoolingTime() {
```
KnightMiner: Javadoc, or use lombok to generate

src/main/java/slimeknights/tconstruct/smeltery/recipe/CastingBasinRecipe.java
```diff
@@ -0,0 +1,26 @@
...
+import javax.annotation.Nullable;
+
+public class CastingBasinRecipe extends AbstractCastingRecipe {
+
+  public CastingBasinRecipe(ResourceLocation idIn, String groupIn, @Nullable Ingredient ingredient, @Nonnull FluidStack fluidIn, ItemStack result, int coolingTime, boolean consumed, boolean switchSlots) {
+    super(TinkerSmeltery.basinRecipeType, idIn, groupIn, ingredient, fluidIn, result, coolingTime, consumed, switchSlots);
+  }
+
+  public ItemStack getIcon() {
+    return new ItemStack(TinkerSmeltery.castingBasin);
```
KnightMiner: I feel we should cache this somewhere. Is this our own method, or overridden?

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingBasinTileEntity.java
```diff
@@ -0,0 +1,17 @@
...
+
+public class CastingBasinTileEntity extends CastingTileEntity {
+
+  public CastingBasinTileEntity() {
+    this(TinkerSmeltery.basin.get(), TinkerSmeltery.basinRecipeType);
+  }
+
+  public CastingBasinTileEntity(TileEntityType<?> tileEntityTypeIn, IRecipeType<? extends AbstractCastingRecipe> recipeType) {
+    super(tileEntityTypeIn, recipeType);
+  }
```
KnightMiner: Do we need this constructor?

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTableTileEntity.java
```diff
@@ -0,0 +1,16 @@
...
+import slimeknights.tconstruct.smeltery.recipe.AbstractCastingRecipe;
+
+public class CastingTableTileEntity extends CastingTileEntity {
+
+  public CastingTableTileEntity() {
+    this(TinkerSmeltery.table.get(), TinkerSmeltery.tableRecipeType);
+  }
+  public CastingTableTileEntity(TileEntityType<?> tileEntityTypeIn, IRecipeType<? extends AbstractCastingRecipe> recipeType) {
+    super(tileEntityTypeIn, recipeType);
+  }
```
KnightMiner: Do we need this constructor?

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+
+  @Override
+  @Nonnull
+  public <T> LazyOptional<T> getCapability(@Nonnull Capability<T> capability, @Nullable Direction facing) {
+    if (capability == CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY)
+      return holder.cast();
+    return super.getCapability(capability, facing);
+  }
+
+  public void interact(PlayerEntity player) {
```
KnightMiner: Javadoc

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  @Override
+  public boolean canInsertItem(int index, ItemStack itemStackIn, @Nullable Direction direction) {
+    return index == INPUT && !isStackInSlot(OUTPUT);
+  }
+
+  @Override
+  public boolean canExtractItem(int index, ItemStack stack, Direction direction) {
+    return index == 1;
```
KnightMiner: `index == OUTPUT`

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+      else if (getWorld().rand.nextFloat() > 0.9f) {
+        world.addParticle(ParticleTypes.SMOKE, pos.getX() + getWorld().rand.nextDouble(), pos.getY() + 1.1d, pos.getZ() + getWorld().rand.nextDouble(), 0.0D, 0.0D, 0.0D);
+      }
+    }
+  }
+
+  protected AbstractCastingRecipe findRecipe(Fluid fluid) {
+
+    recipes.clear();
+    recipes = world.getRecipeManager().getRecipes((IRecipeType<AbstractCastingRecipe>) recipeType, this, world);
```
KnightMiner: This seems inefficient, if you are going to store the list, store it once per type. Alo not a fan of that cast, can we find a way to not need it?

src/main/resources/assets/tconstruct/models/block/smeltery/casting/table.json
```diff
@@ -0,0 +1,110 @@
+{
+    "parent": "block/block",
+    "textures": {
+        "top": "tconstruct:block/smeltery/casting/castingtable_top",
```
KnightMiner: You are in the casting folder, just `casting/table_top` will suffice. I hate the old 1.12 names, but did not want to make a breaking change for resource packs.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
```diff
@@ -685,6 +675,92 @@ private void addSmelteryRecipes(Consumer<IFinishedRecipe> consumer) {
...
+      .build(consumer, location("smeltery/casting/table"));
+  }
+
+  private void addCastingRecipes(Consumer<IFinishedRecipe> consumer) {
+    // Pure Fluid Recipes
+    // Slime
+    this.addSlimeCastingRecipe(consumer, TinkerFluids.blood, SlimeType.BLOOD);
+    this.addSlimeCastingRecipe(consumer, TinkerFluids.blueSlime, SlimeType.BLUE);
+    this.addSlimeCastingRecipe(consumer, TinkerFluids.purpleSlime, SlimeType.PURPLE);
+    this.addBucketCastingRecipe(consumer, TinkerFluids.searedStone);
```
KnightMiner: I don't like a separate recipe for each bucket, that will make mod compat a pain. Make a dynamic fluid container recipe instead. Will accept `Items.BUCKET` and return the bucket filled with the relevant fluid. To make it work with the recipe, either extend the casting table recipe, or have a shared base class.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
```diff
@@ -685,6 +675,92 @@ private void addSmelteryRecipes(Consumer<IFinishedRecipe> consumer) {
...
+      .addCriterion("has_item", this.hasItem(TinkerFluids.searedStone.asItem()))
+      .build(consumer, location("casting/seared_cobble"));
+
+    CastingRecipeBuilder.basinRecipe(TinkerSmeltery.searedGlass)
+      .setFluid(new FluidStack(TinkerFluids.searedStone.get(), MaterialValues.VALUE_Ingot * 4))
+      .setCast(Tags.Items.GLASS_COLORLESS, true)
+      .addCriterion("has_item", this.hasItem(TinkerFluids.searedStone.asItem()))
+      .build(consumer, location("casting/seared_glass"));
+
+    // Two recipes for now until tagged.
```
KnightMiner: Two recipes? Also, we have compound ingredients for the record, to pass in multiple input options

KnightMiner (changes requested):
Added some initial comments. Have not had a chance to test in game yet, but by the looks of this both TESR and JEI are WIP, which is fine.

src/main/java/slimeknights/tconstruct/library/registration/RecipeSerializerDeferredRegister.java
hanetzer: Ah yeah. There was no base defferred reg in tree when I wrote this. Actually,
I think there still isn't; isn't that in your partbuilder branch?

src/main/java/slimeknights/tconstruct/library/smeltery/IFaucetDepth.java
hanetzer: hm. Don't remember putting that in. Perhaps when I was splitting the faucet commits
out I left that behind

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
hanetzer: the variable or the json field?

src/main/java/slimeknights/tconstruct/smeltery/recipe/CastingBasinRecipe.java
hanetzer: From the parent. Honestly, I don't think this is needed, if we use the isDynamic=true
to keep the casting recipes out of the vanilla book I don't think its used.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
hanetzer: dang it. Missed that one.

src/main/resources/assets/tconstruct/models/block/smeltery/casting/table.json
hanetzer: Aight, since we're breaking and fixing everything already may as well
do it.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
hanetzer: Actually I just missed removing the comment. it was for vanilla and clear
glass panes before you tagged them.

src/main/java/slimeknights/tconstruct/library/registration/RecipeSerializerDeferredRegister.java
KnightMiner: There was always a base deferred register. Forge is the base deferred register. We just have wrappers around their register.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
KnightMiner: JSON field. variable is fine in camel case.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
hanetzer: Honestly unsure how that would go down. Are our buckets of molten stuff
just instances of the vanilla bucket? Or the forge bucket?

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
hanetzer: I think this can be turned to a `getRecipe()` (note the lack of an 's') with the
change to AbstractCastingRecipe you suggested on irc, which seems to not
need the cast. Testing is required and I'm trying to be very careful to avoid
breaking what already works.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
KnightMiner: Does not matter, IItemFluidHandler. Look into Forge's fluid utils to fill an item fluid container. So basically, the recipe will fill the input stack (passed in as iron bucket in JSON) with the 100b of fluid and get the result. Look at the 1.12 recipe for an example

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+
+public class CastingTileEntity extends InventoryTileEntity implements ITickableTileEntity, ISidedInventory, FluidUpdatePacket.IFluidPacketReceiver {
+  private int INPUT = 0;
+  private int OUTPUT = 1;
+  public FluidTankAnimated tank;
+  public LazyOptional<CastingFluidHandler> holder = LazyOptional.of(() -> new CastingFluidHandler(this, tank));
+  protected int timer;
+  private List<? extends AbstractCastingRecipe> recipes = Lists.newArrayList();
+  protected AbstractCastingRecipe recipe;
+  protected final IRecipeType<? extends AbstractCastingRecipe> recipeType;
```
KnightMiner: Remove the `? extends here` in light of my suggested change to `TinkerSmeltery`

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+  }
+
+  protected AbstractCastingRecipe findRecipe(Fluid fluid) {
+
+    recipes.clear();
+    recipes = world.getRecipeManager().getRecipes((IRecipeType<AbstractCastingRecipe>) recipeType, this, world);
+
+    for (AbstractCastingRecipe candidate : recipes) {
+      if (candidate.matches(fluid, this, world)) {
+        recipe = candidate;
```
KnightMiner: I do not like the instance variable here, makes read only hard. Just return the variable and let the caller worry about caching the old recipe

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+      }
+    }
+//    CastingRecipe recipe = getWorld().getRecipeManager().getRecipe(TinkerSmeltery.castingRecipeType, this, this.world).orElse(null);
+    return recipe;
+  }
+
+  public int initNewCasting(Fluid fluid, IFluidHandler.FluidAction action) {
```
KnightMiner: Javadoc, state where this is called

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+  public void updateFluidTo(FluidStack fluid) {
+    int oldAmount = tank.getFluidAmount();
+    tank.setFluid(fluid);
+
+    if (fluid.isEmpty()) {
+      reset();
+      return;
+    }
+    else if (recipe == null) {
+      findRecipe(fluid.getFluid());
```
KnightMiner: `recipe = findRecipe()`

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+      }
+    }
+//    CastingRecipe recipe = getWorld().getRecipeManager().getRecipe(TinkerSmeltery.castingRecipeType, this, this.world).orElse(null);
+    return recipe;
+  }
+
+  public int initNewCasting(Fluid fluid, IFluidHandler.FluidAction action) {
+    AbstractCastingRecipe recipe = findRecipe(fluid);
+    if (recipe != null) {
+      if (action == IFluidHandler.FluidAction.SIMULATE) {
```
KnightMiner: Simulate means don't change anything

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+          getWorld().notifyNeighborsOfStateChange(this.pos, this.getBlockState().getBlock());
+        }
+      }
+      else if (getWorld().rand.nextFloat() > 0.9f) {
+        world.addParticle(ParticleTypes.SMOKE, pos.getX() + getWorld().rand.nextDouble(), pos.getY() + 1.1d, pos.getZ() + getWorld().rand.nextDouble(), 0.0D, 0.0D, 0.0D);
+      }
+    }
+  }
+
+  protected AbstractCastingRecipe findRecipe(Fluid fluid) {
```
KnightMiner: Check if the cached recipe matches first before checking the list. Saves effort as most of the time this is called mid recipe.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,202 @@
+package slimeknights.tconstruct.smeltery.recipe;
```
KnightMiner: Why is this not in library? Recipes are part of our "public" API as other mods will likely use them (in the builder that is)

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
```diff
@@ -92,12 +104,51 @@
...
+  public static final ItemObject<Item> toughBindingCast = ITEMS.register("tough_binding_cast", SMELTERY_PROPS);
+
 
+  /*
+   * Recipe
+   */
+  public static final RecipeSerializerObject<AbstractCastingRecipe.Serializer<CastingBasinRecipe>> basinRecipeSerializer = RECIPE_SERIALIZERS.register("casting_basin", () -> new AbstractCastingRecipe.Serializer<>(CastingBasinRecipe::new));
+  public static final RecipeSerializerObject<AbstractCastingRecipe.Serializer<CastingTableRecipe>> tableRecipeSerializer = RECIPE_SERIALIZERS.register("casting_table", () -> new AbstractCastingRecipe.Serializer<>(CastingTableRecipe::new));
+  public static final IRecipeType<CastingBasinRecipe> basinRecipeType = IRecipeType.register(Util.resource("casting_basin"));
+  public static final IRecipeType<CastingTableRecipe> tableRecipeType = IRecipeType.register(Util.resource("casting_table"));
```
KnightMiner: You can make these just `IRecipeType<AbstractCastingRecipe>`. Being less specific will not hurt anything as the table and basin classes do not add anything unique. Same for the serializers. Lets you fix odd casts calling it

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,236 @@
...
+import slimeknights.tconstruct.smeltery.network.FluidUpdatePacket;
+import slimeknights.tconstruct.smeltery.recipe.AbstractCastingRecipe;
+
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+import java.util.List;
+
+public class CastingTileEntity extends InventoryTileEntity implements ITickableTileEntity, ISidedInventory, FluidUpdatePacket.IFluidPacketReceiver {
+  private int INPUT = 0;
+  private int OUTPUT = 1;
```
KnightMiner: `protected static final int`. For input, it needs to be public for my wrapper.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
hanetzer: Honestly I just null the recipe in `reset()`, but I suppose checking if `recipe.matches()`
before doing the logic is still a good idea.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
KnightMiner: Looking over this again, noticed a couple other fields in JSON like that. In general, JSON fields should be `snake_case` instead of `lowercase`

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
KnightMiner: I think it might be worth having two variables here. `currentRecipe` is set as you use recipe right now, represents the recipe we are working on and is cleared on reset. `lastRecipe` is a cache of the last recipe returned by this method, and is checks before you do the list. Can update it on simulate, so the actual result will use the cache from simulate.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
hanetzer: Maybe not; I was just following vanilla Furnace/Cooking conventions.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
hanetzer: because I just picked a spot to use so I didn't get stuck in 'what's the right place for this'
mode. Just the AbstractCastingRecipe, or its two children?

src/main/java/slimeknights/tconstruct/common/data/CastingRecipeBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+import net.minecraftforge.fluids.FluidStack;
+import net.minecraftforge.registries.ForgeRegistries;
+import slimeknights.tconstruct.library.recipe.RecipeUtil;
+import slimeknights.tconstruct.smeltery.TinkerSmeltery;
+import slimeknights.tconstruct.smeltery.recipe.AbstractCastingRecipe;
+
+import javax.annotation.Nullable;
+import java.util.function.Consumer;
+
+public class CastingRecipeBuilder {
```
KnightMiner: Look into `AbstractRecipeBuilder`, it handles the advancement builder for you. Extend it like `class CastingRecipeBuilder extends AbstractRecipeBuilder<CastingRecipeBuilder>`, and in your build, call that one funny method that returns a resource location. Let me know if you have questions, or see something else that would be good in the abstract builder.

src/main/java/slimeknights/tconstruct/common/data/CastingRecipeBuilder.java
```diff
@@ -0,0 +1,189 @@
...
+    return this.setCast(Ingredient.fromItems(itemIn), consumed);
+  }
+
+  public CastingRecipeBuilder setCast(Ingredient ingredient, boolean consumed) {
+    this.cast = ingredient;
+    this.castConsumed = consumed;
+    return this;
+  }
+
+  public CastingRecipeBuilder setSwitchSlots() {
```
KnightMiner: Might we worth Javadocs on a few of these more obscure methods especially. It is a bit redundant to the recipe class, but most people will use the builder.

src/main/java/slimeknights/tconstruct/plugin/jei/JEIPlugin.java
```diff
@@ -0,0 +1,62 @@
...
+    final IGuiHelper guiHelper = registry.getJeiHelpers().getGuiHelper();
+    castingBasinCategory = new CastingBasinCategory(guiHelper);
+    castingTableCategory = new CastingTableCategory(guiHelper);
+    registry.addRecipeCategories(castingBasinCategory);
+    registry.addRecipeCategories(castingTableCategory);
+  }
+
+  @Override
+  public void registerRecipes(IRecipeRegistration register) {
+    Collection<IRecipe<ICastingInventory>> castingBasinRecipes = Minecraft.getInstance().world.getRecipeManager().getRecipes(RecipeTypes.CASTING_BASIN).values();
```
KnightMiner: Look into `RecipeUtil.getRecipes`, it will let you apply a filter class, such as if you only want `CastingBasinRecipe` to be added. Or just get the generic proper type list if you wish (saves potential unhandled recipes in JEI):
```java
List<AbstractCastingRecipe> = RecipeUtil.getRecipes(recipeManager, RecipeTypes.CASTING_BASIN, AbstractCastingRecipe.class);
```

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
```diff
@@ -0,0 +1,86 @@
...
+import mezz.jei.api.ingredients.IIngredients;
+import mezz.jei.api.recipe.category.IRecipeCategory;
+import net.minecraft.block.Block;
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.fluids.FluidStack;
+import slimeknights.tconstruct.library.Util;
+import slimeknights.tconstruct.smeltery.recipe.AbstractCastingRecipe;
+
+public abstract class AbstractCastingCategory<T extends AbstractCastingRecipe> implements IRecipeCategory<T> {
```
KnightMiner, line 25: Is this still WIP? I see a lot of commented code and the cooling time does not seem done.

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
KnightMiner: As an update on this, maybe just comment out casts other than those in alex's PR. So broadsword (except the sword cross guard, that one is likely removed), pickaxe and shovel. I can also link you the WIP parts doc if you dont have it, will show which parts still exist

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
```diff
@@ -92,12 +102,49 @@
...
+  public static final ItemObject<Item> bowLimbCast = ITEMS.register("bow_limb_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> hammerHeadCast = ITEMS.register("hammer_head_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> excavatorHeadCast = ITEMS.register("excavator_head_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> wideGuardCast = ITEMS.register("wide_guard_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> handGuardCast = ITEMS.register("hand_guard_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> crossGuardCast = ITEMS.register("cross_guard_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> gearCast = ITEMS.register("gear_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> broadAxeHeadCast = ITEMS.register("broad_axe_head_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> shovelHeadCast = ITEMS.register("shovel_head_cast", SMELTERY_PROPS);
+  public static final ItemObject<Item> arrowHeadCast = ITEMS.register("arrow_head_cast", SMELTERY_PROPS);
```
KnightMiner: For specifically the arrow and arrow shafts, those are removed. We also removed the sword guards in favor of reusing existing parts.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
KnightMiner: Yeah, I know the packages are a little weird. Eventually I will make a proper guide of what goes in library and what does not.

For the non-abstract recipes, it a good question. Officially, mods should not need to use the class, directly but on the other hand, it makes it all in one place. My rule of thumb before we had JSON recipes was if its a special case recipe, keep it in the module, but if its a recipe other mods may use, put it in library. I guess we can stick with that for the modules.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
KnightMiner: I suppose do whatever feels cleaner to you, parameter or override.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,173 @@
...
+      boolean consumed = false;
+      boolean switchSlots = JSONUtils.getBoolean(json, "switch_slots", false);
+      if (json.has("cast")) {
+        JsonElement jsonelement = JSONUtils.getJsonObject(json, "cast");
+        cast = Ingredient.deserialize(jsonelement);
+        consumed = JSONUtils.getBoolean(json, "cast_consumed", false);
+      }
+
+      if (!json.has("fluidstack"))
+        throw new JsonSyntaxException("Missing fluid input definition!");
```
KnightMiner: This is redundant to the line below, `JSONUtils.getWhatever` methods all throw if the whatever is missing.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,173 @@
...
+      JsonObject jsonFluid = JSONUtils.getJsonObject(json, "fluidstack");
+      FluidStack fluidStack = RecipeUtil.deserializeFluidStack(jsonFluid);
+      ItemStack item = new ItemStack(JSONUtils.getItem(json, "result"));
+      int coolingtime;
+      if (!json.has("cooling_time")) {
+        int time = 24;
+        int temperature = fluidStack.getFluid().getAttributes().getTemperature() - 300;
```
KnightMiner: Double check this one when both me and diyo are online. Right now I think we store fluid units in Celsius, as its better for the player, but its inconsistent with the temperature other mods may show them from fluids.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
```diff
@@ -0,0 +1,173 @@
...
+      if (json.has("cast")) {
+        JsonElement jsonelement = JSONUtils.getJsonObject(json, "cast");
+        cast = Ingredient.deserialize(jsonelement);
+        consumed = JSONUtils.getBoolean(json, "cast_consumed", false);
+      }
+
+      if (!json.has("fluidstack"))
+        throw new JsonSyntaxException("Missing fluid input definition!");
+      JsonObject jsonFluid = JSONUtils.getJsonObject(json, "fluidstack");
+      FluidStack fluidStack = RecipeUtil.deserializeFluidStack(jsonFluid);
```
KnightMiner: Eventually, I want tags for this, but we will need a proper "fluid ingredient" for that. For this PR, single fluid matches is fine.

src/main/java/slimeknights/tconstruct/smeltery/recipe/CastingBasinRecipe.java
KnightMiner: I looked at the vanilla stuff, and say no cache is needed. Even though these are not used, might as well keep them in I guess.

src/main/java/slimeknights/tconstruct/smeltery/recipe/ICastingInventory.java
```diff
@@ -0,0 +1,108 @@
...
+import net.minecraft.fluid.Fluid;
+import net.minecraft.inventory.IInventory;
+import net.minecraft.item.Item;
+import net.minecraft.item.ItemStack;
+import net.minecraftforge.fluids.FluidStack;
+
+import java.util.Set;
+
+/* TODO: move the required methods to base interface */
+public interface ICastingInventory extends IInventory {
```
KnightMiner: Should be able to extend the new `ISingleItemInventory`, which will do all the base methods for you (solving my TODO when I sent you this interface). You will only need to add the `getFluid()` method then, as `ISingleItemInventory` gives you `getStack()`

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
```diff
@@ -0,0 +1,232 @@
...
+    if (this.recipe != null && this.recipe.matches(crafting, getWorld())) {
+      return this.recipe;
+    }
+    AbstractCastingRecipe castingRecipe = getWorld().getRecipeManager().getRecipe(this.recipeType, crafting, getWorld()).orElse(null);
+
+    return castingRecipe;
+  }
+
+  public int initNewCasting(Fluid fluid, IFluidHandler.FluidAction action) {
+    if (this.fluid == Fluids.EMPTY) {
```
KnightMiner: I would just set the fluid regardless. Its not really relevant besides in this method, and we don't want to risk a mod simulating with one fluid, then being unable to cast a second one. This is likely causing your bugs with the inconsistently failing recipes, as if you set the fluid and never find a recipe, you are stuck with that fluid forever.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
KnightMiner: Let me expand upon this one: if the `recipe` variable is used in game logic, such as the tick method, it should never be changed during simulate. Anyone can simulate and the block should not be affected, one common example is for interfaces that read fluid data in a tank. If its a "last recipe" cache only used in this method, that is fine, but then the variable should not be referenced in `tick`.

As I mentioned on Discord, best solution is two recipes, `currentRecipe` is the one that `tick` and `reset` will handle, you set it when we get an `EXECUTE`. It can block logic as needed and all that fun. `lastRecipe` is our simple cache so if we are trying to craft the same thing as last time, we save time in lookup. `lastRecipe` will only be modified here, and is modified regardless of the action. `currentRecipe` is modified in multiple places.

KnightMiner (commented):
I see a couple major things that still look WIP, mainly JEI. Rest has very good progess, save a few larger concerns in comments.

One style note, if a method overrides a parent method, make sure to annotate. IDEA has a save actions addon you can add that will add the annotation to all required methods when you hit save, so you don't have to think about it.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
hanetzer: Nope. This isn't the cause of the sporadic failures... very strange.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
hanetzer: Agreed. Would be nice to get some of the stuff you wrote into forge proper.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTileEntity.java
KnightMiner: In any case, this still is wrong, so remove the condition.

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
hanetzer: Did so.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
hanetzer: Having a bit of fun figuring out how to make the arrow do the same as cooling time. Perhaps not
such a good idea, as some things cool in 2s, which would look pretty spastic.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
hanetzer: I see... in 1.12 the casting recipe setup was an interface instead of an abstract class.
How's about this, we get this pr merged more/less as-is and I'll work out the details
for this setup in 1.16.x, I've already racked my brain quite enough over it, I think.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
hanetzer: On another note, are we going to perhaps make an api jar for folks to use?
I'm not certain as to the mechanics of that in modding context.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
```diff
@@ -1079,6 +1160,92 @@ private void addSearedStonecutter(@Nonnull Consumer<IFinishedRecipe> consumer, I
...
+    }
+  }
+
+  private void addCastCastingRecipe(@Nonnull Consumer<IFinishedRecipe> consumer, @Nonnull Tag<Item> tag, @Nonnull IItemProvider cast) {
+    CastingRecipeBuilder.tableRecipe(cast)
+      .setFluid(new FluidStack(TinkerFluids.moltenGold.get(), MaterialValues.VALUE_Ingot))
+      .setCast(tag, true)
+      .setSwitchSlots()
+      .addCriterion("has_item", this.hasItem(tag))
+      .build(consumer, location("casting/casts/" + tag.getId().getPath()));
```
KnightMiner: Typically, we name recipes based on their output.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
```diff
@@ -1079,6 +1160,92 @@ private void addSearedStonecutter(@Nonnull Consumer<IFinishedRecipe> consumer, I
...
+  }
+
+  private void addCastCastingRecipe(@Nonnull Consumer<IFinishedRecipe> consumer, @Nonnull IItemProvider item, @Nonnull IItemProvider cast) {
```
KnightMiner: Name the parameter `input`, `item` is ambiguous.

src/main/java/slimeknights/tconstruct/smeltery/recipe/AbstractCastingRecipe.java
KnightMiner: Its definitely not possible with our current library structure. Its something we can look into in the future, but not a priority.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingBasinTileEntity.java
```diff
@@ -0,0 +1,18 @@
...
+import slimeknights.tconstruct.smeltery.TinkerSmeltery;
+import slimeknights.tconstruct.smeltery.recipe.AbstractCastingRecipe;
+
+public class CastingBasinTileEntity extends AbstractCastingTileEntity {
+
+  public CastingBasinTileEntity() {
+    this(TinkerSmeltery.basin.get(), RecipeTypes.CASTING_BASIN);
+  }
+
+  public CastingBasinTileEntity(TileEntityType<?> tileEntityTypeIn, IRecipeType<AbstractCastingRecipe> recipeType) {
```
KnightMiner: Remove this constructor, if you are going to call with existing types you should call the abstract.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/CastingTableTileEntity.java
```diff
@@ -0,0 +1,17 @@
...
+import slimeknights.tconstruct.library.recipe.RecipeTypes;
+import slimeknights.tconstruct.smeltery.TinkerSmeltery;
+import slimeknights.tconstruct.smeltery.recipe.AbstractCastingRecipe;
+
+public class CastingTableTileEntity extends AbstractCastingTileEntity {
+
+  public CastingTableTileEntity() {
+    this(TinkerSmeltery.table.get(), RecipeTypes.CASTING_TABLE);
+  }
+  public CastingTableTileEntity(TileEntityType<?> tileEntityTypeIn, IRecipeType<AbstractCastingRecipe> recipeType) {
```
KnightMiner: Remove this constructor, if you are going to call with existing types you should call the abstract.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
```diff
@@ -0,0 +1,103 @@
...
+import net.minecraft.client.Minecraft;
+import net.minecraft.client.gui.FontRenderer;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.crafting.Ingredient;
+import net.minecraft.util.ResourceLocation;
+import net.minecraftforge.fluids.FluidStack;
+import slimeknights.tconstruct.library.Util;
+import slimeknights.tconstruct.smeltery.recipe.AbstractCastingRecipe;
+
+import java.awt.*;
```
KnightMiner: Not a fan of wildcard imports, makes things harder to trace without an IDE.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/AbstractCastingTileEntity.java
```diff
@@ -0,0 +1,232 @@
...
+
+public class AbstractCastingTileEntity extends InventoryTileEntity implements ITickableTileEntity, ISidedInventory, FluidUpdatePacket.IFluidPacketReceiver {
+  public static final int INPUT = 0;
+  protected static final int OUTPUT = 1;
+  public FluidTankAnimated tank;
+  public LazyOptional<CastingFluidHandler> holder = LazyOptional.of(() -> new CastingFluidHandler(this, tank));
+  private final ICastingInventory crafting;
+  protected int timer;
+  @Getter @Setter
+  private Fluid fluid;
```
KnightMiner: Might be better to put this field in the wrapper, like I did with the part builder with the material recipe. Reason being this is not technically the fluid of the TE, but the one used for finding recipes. May cause a mod using our "API" to call the wrong method.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
hanetzer: Fair enough, but this is how intellij did it 'automagically' and it matches 1.12

KnightMiner: IDEA can be configured to not use them. And it shouldn't be that way in 1.12, boni never liked wildcard imports either

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
hanetzer: Is this not? `tconstruct:casting/casts/nuggets` is a cast for nuggets?

KnightMiner: The tag ID may be similar to the output, but its not strictly the same. this is more different for the item version than the tag one, where the input may be a specific material like bricks for the ingot cast.

src/main/java/slimeknights/tconstruct/common/data/CastingRecipeBuilder.java
hanetzer: I javadoc'd this one. what would you also consider obscure here?

src/main/java/slimeknights/tconstruct/library/recipe/casting/CastingRecipeBuilder.java
```diff
@@ -0,0 +1,175 @@
...
+    if (saveKey.equals(resultKey)) {
+      throw new IllegalStateException("Recipe " + saveKey + " should remove its 'save' argument");
+    } else {
+      this.build(consumerIn, saveKey);
+    }
+  }
+
+  public void build(Consumer<IFinishedRecipe> consumerIn, ResourceLocation id) {
+    this.validate(id);
+    this.advancementBuilder.withParentId(new ResourceLocation("recipes/root")).withCriterion("has_the_recipe", new RecipeUnlockedTrigger.Instance(id)).withRewards(AdvancementRewards.Builder.recipe(id)).withRequirementsStrategy(IRequirementsStrategy.OR);
```
KnightMiner: Method in `AbstractRecipeBuilder` for this, also handles the second half of `validate` below

src/main/java/slimeknights/tconstruct/library/recipe/casting/CastingRecipeBuilder.java
```diff
@@ -0,0 +1,175 @@
...
+  public CastingRecipeBuilder setSwitchSlots() {
+    this.switchSlots = true;
+    return this;
+  }
+
+  public void build(Consumer<IFinishedRecipe> consumerIn) {
+    this.build(consumerIn, ForgeRegistries.ITEMS.getKey(this.result));
+  }
+
+  public void build(Consumer<IFinishedRecipe> consumerIn, String save) {
```
KnightMiner, line 89: Don't really need this, since we never save recipes to the `minecraft` namespace

src/main/java/slimeknights/tconstruct/library/recipe/casting/CastingRecipeBuilder.java
```diff
@@ -0,0 +1,175 @@
...
+   * Set output of recipe to be put into the input slot.
+   * Mostly used for 'casts'
+   */
+  public CastingRecipeBuilder setSwitchSlots() {
+    this.switchSlots = true;
+    return this;
+  }
+
+  public void build(Consumer<IFinishedRecipe> consumerIn) {
+    this.build(consumerIn, ForgeRegistries.ITEMS.getKey(this.result));
```
KnightMiner: `this.result.getRegistryName()`

src/main/java/slimeknights/tconstruct/plugin/jei/JEIPlugin.java
```diff
@@ -0,0 +1,65 @@
...
+
+import javax.annotation.Nullable;
+import java.util.Collection;
+import java.util.List;
+
+@JeiPlugin
+public class JEIPlugin implements IModPlugin {
+
+  @Nullable
+  private CastingBasinCategory castingBasinCategory;
```
KnightMiner: Do we use these fields anywhere? Might as well make them local if not

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
KnightMiner: Mezz just went for 1/20th of the time in JEI IIRC. Look at the furnace recipe.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
```diff
@@ -0,0 +1,103 @@
...
+    ingredients.setOutput(VanillaTypes.ITEM, recipe.getRecipeOutput());
+  }
+
+  @Override
+  public void draw(T recipe, double mouseX, double mouseY) {
+    arrow.draw(80, 25);
+
+    int coolingTime = recipe.getCoolingTime() / 20;
+    // TODO: Localizable 's'
+    String coolingString;// = Util.translate("gui.jei.casting.cooling_time", coolingTime);
```
KnightMiner, line 69: You can use `ForgeI18n` or `TranslationTextComponent` to do this. Note `ForgeI18n` uses `{0}` for the first parameter instead of `%s`.

src/main/java/slimeknights/tconstruct/smeltery/recipe/ICastingInventory.java
```diff
@@ -0,0 +1,17 @@
...
+  /**
+   * Gets the contained fluid in this inventory
+   * @return  Contained fluid
+   */
+  Fluid getFluid();
+
+  /**
+   * Sets the contained fluid in this inventory
+   */
+  void setFluid(Fluid fluid);
```
KnightMiner: Don't put `setFluid` here, the interface is intended to be read only, recipes never need to change it (and it will lead to super buggy behavior if they do change it). Its fine to leave in the wrapper.

Plus, based on the specific implementation we may not have a settable fluid. For instance, someone makes an infinite water caster.

src/main/java/slimeknights/tconstruct/smeltery/recipe/TileCastingWrapper.java
```diff
@@ -0,0 +1,29 @@
...
+public class TileCastingWrapper implements ICastingInventory {
+  private final AbstractCastingTileEntity tile;
+  private Fluid fluid;
+
+  @Override
+  public ItemStack getStack() {
+    return tile.getStackInSlot(AbstractCastingTileEntity.INPUT);
+  }
+
+  public Fluid getFluid() {
```
KnightMiner, line 21: `@Override`. Could also just toss `@Getter` on the field.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/AbstractCastingTileEntity.java
```diff
@@ -0,0 +1,234 @@
...
+  public static final int OUTPUT = 1;
+  public FluidTankAnimated tank;
+  public LazyOptional<CastingFluidHandler> holder = LazyOptional.of(() -> new CastingFluidHandler(this, tank));
+  private final ICastingInventory crafting;
+  protected int timer;
+  protected AbstractCastingRecipe recipe;
+  protected final IRecipeType<AbstractCastingRecipe> recipeType;
+
+  public AbstractCastingTileEntity(TileEntityType<?> tileEntityTypeIn, IRecipeType<AbstractCastingRecipe> recipeType) {
+    super(tileEntityTypeIn, new TranslationTextComponent("casting"), 2, 1);
```
KnightMiner: Name seems too short, that will translate it as simply `casting`. While you don't typically have a UI, it will show in the crafting station side inventory, so make it something like `gui.tconstruct.casting`.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/AbstractCastingTileEntity.java
```diff
@@ -0,0 +1,234 @@
...
+      }
+      else if (getWorld().rand.nextFloat() > 0.9f) {
+        world.addParticle(ParticleTypes.SMOKE, pos.getX() + getWorld().rand.nextDouble(), pos.getY() + 1.1d, pos.getZ() + getWorld().rand.nextDouble(), 0.0D, 0.0D, 0.0D);
+      }
+    }
+  }
+
+  protected AbstractCastingRecipe findRecipe() {
+    if (this.recipe != null && this.recipe.matches(crafting, getWorld())) {
+      return this.recipe;
```
KnightMiner: I suggest using a separate variable such as `lastRecipe` for this. `recipe` is set to `null` in `reset()`, so you will get no benefit if you cast the same item multiple times in a row.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/AbstractCastingTileEntity.java
```diff
@@ -0,0 +1,234 @@
...
+          getWorld().notifyNeighborsOfStateChange(this.pos, this.getBlockState().getBlock());
+        }
+      }
+      else if (getWorld().rand.nextFloat() > 0.9f) {
+        world.addParticle(ParticleTypes.SMOKE, pos.getX() + getWorld().rand.nextDouble(), pos.getY() + 1.1d, pos.getZ() + getWorld().rand.nextDouble(), 0.0D, 0.0D, 0.0D);
+      }
+    }
+  }
+
+  protected AbstractCastingRecipe findRecipe() {
```
KnightMiner, line 170: Mark `@Nullable` since it may return null

src/main/java/slimeknights/tconstruct/smeltery/tileentity/AbstractCastingTileEntity.java
```diff
@@ -0,0 +1,234 @@
...
+  /** Called from CastingFluidHandler.fill()
+   * @param fluid   Fluid used in casting
+   * @param action  EXECUTE or SIMULATE
+   * @return        Amount of fluid needed for recipe, used to resize the tank.
+   */
+  public int initNewCasting(Fluid fluid, IFluidHandler.FluidAction action) {
+    this.crafting.setFluid(fluid);
+
+    AbstractCastingRecipe castingRecipe = findRecipe();
+    if (castingRecipe != null && this.recipe == null) {
```
KnightMiner, line 190: If you are going to skip if we have an existing recipe, might as well skip the call to `findRecipe` entirely. Can also return the size for the existing recipe in that case.

src/main/java/slimeknights/tconstruct/library/recipe/casting/CastingTableRecipe.java
```diff
@@ -0,0 +1,27 @@
...
+import javax.annotation.Nonnull;
+import javax.annotation.Nullable;
+
+public class CastingTableRecipe extends AbstractCastingRecipe {
+
+  public CastingTableRecipe(ResourceLocation idIn, String groupIn, @Nullable Ingredient ingredient, @Nonnull FluidStack fluidIn, ItemStack result, int coolingTime, boolean consumed, boolean switchSlots) {
+    super(RecipeTypes.CASTING_TABLE, idIn, groupIn, ingredient, fluidIn, result, coolingTime, consumed, switchSlots);
+  }
+
+  public ItemStack getIcon() {
```
KnightMiner, line 21: `@Override`

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
KnightMiner: Sure, we can deal with the bucket recipe afterwards

src/main/java/slimeknights/tconstruct/common/data/CastingRecipeBuilder.java
KnightMiner: Switch slots is a little vague, since from the outside viewpoint the casting blocks only have one slot. Anything else along the same line, booleans mostly.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/AbstractCastingTileEntity.java
hanetzer: Gotcha, I just shoved something in there to have a thing.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
hanetzer: Problem is all their furnace/etc recipe stuff has a fixed time. we don't. They pass time in
via the ctor.

KnightMiner: Oh right, forgot they did that. There might be a way to tell it a time when you draw it, but its not really a high priority.

src/main/java/slimeknights/tconstruct/plugin/jei/JEIPlugin.java
hanetzer: Don't think so, just modeled after JEI's furnace stuff.

KnightMiner (approved):

PR #4138, opened 2020-06-25
forge: update mappings to 20200625 and forge to 1.15.2-31.2.27

hanetzer:
Fixes #4132


hanetzer (commented):
I'm not sure if I should do both.

PR #4146, opened 2020-07-09
Material casting

hanetzer:
Mostly seems to work. Testing welcomed


src/main/java/slimeknights/tconstruct/library/recipe/casting/PartCastingRecipe.java
```diff
@@ -0,0 +1,55 @@
...
+import net.minecraft.util.ResourceLocation;
+import net.minecraft.world.World;
+import slimeknights.tconstruct.library.materials.IMaterial;
+import slimeknights.tconstruct.library.recipe.RecipeTypes;
+import slimeknights.tconstruct.library.recipe.material.MaterialCastingRecipe;
+import slimeknights.tconstruct.library.tinkering.IMaterialItem;
+import slimeknights.tconstruct.smeltery.TinkerSmeltery;
+import slimeknights.tconstruct.smeltery.recipe.ICastingInventory;
+
+public class PartCastingRecipe extends AbstractCastingRecipe {
```
KnightMiner: Don't extend `AbstractCastingRecipe`, thats a poor fit. Make a new `ICastingRecipe` an have both classes implement. Removes the need to change `AbstractCastingRecipe` for the sake of a dynamic recipe.

Alternatively, put out the common logic between abstract and part casting into a shared parent. The key detail is I dislike that you pass in an empty fluid to meet the need of the parent (its going to be a fluid ingredient in the future anyways, so we will have to break this again)

src/main/java/slimeknights/tconstruct/smeltery/recipe/ICastingInventory.java
```diff
@@ -2,11 +2,21 @@
...
    * @return  Contained fluid
    */
   Fluid getFluid();
+
+  /**
+   * Gets the material recipe based on the current slot contents
+   * @return  Material recipe
+   */
+  @Nullable
+  MaterialCastingRecipe getMaterial();
```
KnightMiner: Not sure how I feel about adding this to the casting inventory. We needed it in the part builder as materials are a primary function of the block, while this only uses them for one specialized recipe.

How much work would it be to find the material within the `PartCastingRecipe`? I guess we may have to do it twice then, once for input, once for output.

src/main/java/slimeknights/tconstruct/smeltery/tileentity/AbstractCastingTileEntity.java
```diff
@@ -60,7 +60,7 @@ public AbstractCastingTileEntity(TileEntityType<?> tileEntityTypeIn, IRecipeType
     this.tank = new FluidTankAnimated(0, this);
     this.itemHandler = new SidedInvWrapper(this, Direction.DOWN);
     this.recipeType = recipeType;
-    this.crafting = new TileCastingWrapper(this, Fluids.EMPTY);
+    this.crafting = new TileCastingWrapper(this, Fluids.EMPTY, null);
```
KnightMiner: Might be better to make the constructor just `this`, set the other two by default.

src/main/java/slimeknights/tconstruct/library/recipe/material/MaterialCastingRecipe.java
```diff
@@ -0,0 +1,71 @@
...
+import net.minecraft.world.World;
+import slimeknights.tconstruct.library.MaterialRegistry;
+import slimeknights.tconstruct.library.materials.IMaterial;
+import slimeknights.tconstruct.library.materials.MaterialId;
+import slimeknights.tconstruct.library.recipe.RecipeTypes;
+import slimeknights.tconstruct.smeltery.TinkerSmeltery;
+import slimeknights.tconstruct.smeltery.recipe.ICastingInventory;
+
+@AllArgsConstructor
+public class MaterialCastingRecipe implements IRecipe<ICastingInventory> {
```
KnightMiner: Not a fan of a whole additional recipe type here. The material type was added because each material has multiple possible inputs of varying amounts and we need it in both part casting and tool repair. Fluids we only ever have one for each material (if you ignore tags).  Just use `IMaterial::getFluid()`

src/main/java/slimeknights/tconstruct/library/recipe/casting/AbstractCastingRecipe.java
```diff
@@ -79,4 +96,19 @@ public boolean matches(ICastingInventory inv, World worldIn) {
   public NonNullList<Ingredient> getIngredients() {
     return NonNullList.from(Ingredient.EMPTY, this.cast);
   }
+
+  /**
+   * Cooling time for the recipe.
+   * @return Cooling time specified in the json, else a dynamically calculated one.
+   */
+  public int getCoolingTime() {
```
KnightMiner: Instead of calculating this on the fly always, have a recipe `getCoolingTime(ICastingInventory)` and just have the part one calculate it

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
```diff
@@ -1210,22 +1222,31 @@ private void addMoltenMineralCastingRecipe(@Nonnull Consumer<IFinishedRecipe> co
...
+    ItemCastingRecipeBuilder.tableRecipe(cast)
+      .setFluid(new FluidStack(TinkerFluids.moltenGold.get(), MaterialValues.VALUE_Ingot))
       .setCast(input, true)
       .setSwitchSlots()
       .addCriterion("has_item", this.hasItem(input))
-      .build(consumer, location("casts/" + input.asItem().getRegistryName().getPath()));
+      .build(consumer, location("casting/casts/" + Objects.requireNonNull(input.asItem().getRegistryName()).getPath()));
   }
+
+  private void addPartCastingRecipe(@Nonnull Consumer<IFinishedRecipe> consumer, @Nonnull Supplier<ToolPartItem> part, @Nonnull IItemProvider cast, int amount) {
```
KnightMiner: As mentioned on discord, look into merging this with the method for the part builder. Can probably merge it with the method to create a part cast (or call that method), as you have parameters for both cast and part here.

KnightMiner: The `@Nonnull` in the parameter is redundant. The `package-info` file says when unspecified all parameters are `@Nonnull`. My recent commit to 1.15 cleaned them up in other methods, so just keep them out of new methods.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ICastingRecipe.java
```diff
@@ -0,0 +1,76 @@
...
+  default NonNullList<Ingredient> getIngredients() {
+    return NonNullList.from(Ingredient.EMPTY, getCast());
+  }
+
+  @Override
+  default boolean isDynamic() {
+    return true;
+  }
+
+  Ingredient getCast();
```
KnightMiner: I don't think we need this in the main interface. While we use it in JEI, that uses `ItemCastingRecipe` for display

src/main/java/slimeknights/tconstruct/library/recipe/casting/ItemCastingRecipe.java
```diff
@@ -0,0 +1,122 @@
...
+    }
+
+    @Override
+    public IRecipeSerializer<?> getSerializer() {
+      return TinkerSmeltery.basinRecipeSerializer.get();
+    }
+
+    @Override
+    public ItemStack getIcon() {
+      return new ItemStack(TinkerSmeltery.castingBasin);
```
KnightMiner: For these icons methods, I suggest the following code in `ICastingRecipe` do we don't have to override this everywhere (here and in material):

```java
default ItemStack getIcon() {
  if (getType() == RecipeTypes.CASTING_TABLE) {
    return new ItemStack(TinkerSmeltery.castingTable);
  }
  return new ItemStack(TinkerSmeltry.castingBasin);
```

src/main/java/slimeknights/tconstruct/library/recipe/casting/ItemCastingRecipeBuilder.java
```diff
@@ -20,71 +21,80 @@
...
   }
 
   /**
-   * Sets the fluid for this recipe, and sets the cooling time based on that fluid
+   * Sets the fluid for this recipe
    * @param fluidStack  Fluid input
    * @return  Builder instance
    */
-  public CastingRecipeBuilder setFluidAndTime(FluidStack fluidStack) {
+  public ItemCastingRecipeBuilder setFluid(FluidStack fluidStack) {
```
KnightMiner, line 94: Not sure why you renamed this and removed the JavaDoc. This method does set the cooling time, should make that clear to the caller. JavaDoc at minimum, method name I am less strict on but it helps.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipe.java
```diff
@@ -0,0 +1,105 @@
...
+  protected final boolean switchSlots;
+
+  @Override
+  public boolean switchSlots() {
+    return this.switchSlots;
+  }
+
+  @Override
+  public int getCoolingTime(ICastingInventory inv) {
+    return ICastingRecipe.calcCoolingTime(inv.getFluid(), this.fluidAmount);
```
KnightMiner: Use `MaterialRegistry.getMaterial(inv.getFluid).getTemperature()` instead of the fluid. Otherwise when we get around to tag support for inputs, another mods iron may change the cooling times. You already know the fluid is valid since we passed `matches`, so no extra checks needed.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipeBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+  private MaterialCastingRecipeBuilder(IMaterialItem result, MaterialCastingRecipeSerializer<?> recipeSerializer) {
+    this.result = result.asItem();
+    this.recipeSerializer = recipeSerializer;
+  }
+
+  public static MaterialCastingRecipeBuilder castingRecipe(IMaterialItem result, MaterialCastingRecipeSerializer<?> recipeSerializer) {
+    return new MaterialCastingRecipeBuilder(result, recipeSerializer);
+  }
+
+  public static MaterialCastingRecipeBuilder tableRecipe(IMaterialItem result) {
```
KnightMiner, line 45: We have the basin serializer registered, might as well add a builder method for it.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipeBuilder.java
```diff
@@ -0,0 +1,140 @@
...
+  }
+
+  /**
+   * Makes sure that this is valid
+   */
+  private void validate(ResourceLocation id) {
+    if (this.fluidAmount <= 0) {
+      throw new IllegalStateException("Material casting recipes require a positive amount of fluid");
+    }
+    if (!(this.result instanceof IMaterialItem)) {
```
KnightMiner: Make this impossible. Make the field type `IMaterialItem` or only accept one as the parameter. Can call `asItem()` in `Result` later if needed.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipeSerializer.java
```diff
@@ -0,0 +1,90 @@
...
+    Ingredient cast = Ingredient.EMPTY;
+    String group = JSONUtils.getString(json, "group", "");
+    boolean consumed = false;
+    boolean switchSlots = JSONUtils.getBoolean(json, "switch_slots", false);
+    if (json.has("cast")) {
+      JsonElement jsonElement = JSONUtils.getJsonObject(json, "cast");
+      cast = Ingredient.deserialize(jsonElement);
+      consumed = JSONUtils.getBoolean(json, "cast_consumed", false);
+    }
+    int fluidAmount = JSONUtils.getInt(json, "cost");
```
KnightMiner: Not a fan of calling this `cost`, since `cost` is used by parts to mean a material amount (1 iron ingot is 1 cost, but 144 fluid amount). I suggest `amount` or `fluid_amount`.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
```diff
@@ -97,8 +98,8 @@ public void setRecipe(IRecipeLayout recipeLayout, T recipe, IIngredients ingredi
 
     IGuiFluidStackGroup fluidStacks = recipeLayout.getFluidStacks();
     int capacity = MaterialValues.VALUE_Block;
-    if (recipe.getType() == RecipeTypes.CASTING_TABLE) {
-      capacity /= 2;
+    if (recipe.getType() == RecipeTypes.CASTING_BASIN) {
+      capacity += MaterialValues.VALUE_BrickBlock;
```
KnightMiner: I think at that point I would just do `capacity = MaterialValues.VALUE_Ingot * 13`, we start to lose the meaning of those values when we sum them together. Maybe 12 ingots is better for basins, or 16. What is the most we use? Blocks are 9, anything bigger you can think of?

KnightMiner (commented):
I think just 1 or 2 things that need fixing, then we should be good to merge. We also have some conflicts, likely related to the commit I made that moved casting recipes into the casting folder.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ICastingRecipe.java
hanetzer: A fair assessment.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ItemCastingRecipe.java
hanetzer: Yeah. done locally.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ItemCastingRecipeBuilder.java
hanetzer: honestly I didn't, but git figured both files are the same enough to call it a `git mv`. Meant to copy that over,
more or less.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipe.java
hanetzer: Done locally.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipeBuilder.java
hanetzer: Oh yeah. I did it as a 'may as well' and missed that part, will do.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
hanetzer: Not offhand, no.

src/main/java/slimeknights/tconstruct/common/data/TConstructRecipeProvider.java
hanetzer: True enough.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ICastingRecipe.java
hanetzer: Well... we do use getCast() in the MaterialCastingRecipe, no? Seems useful.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipeBuilder.java
hanetzer: It is, but I figured better safe than sorry.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipeSerializer.java
hanetzer: went with `fluid_amount`

src/main/java/slimeknights/tconstruct/library/recipe/casting/ICastingRecipe.java
KnightMiner: The casting table never asks for the cast ingredient during crafting. Yes, the material recipe also uses cast ingredients, but we don't want that to be a strict requirement if we are not going to use it. Keep the base interfaces to just the minimum you need for game logic.

Also, whenever the full bucket recipe gets ported, it won't have a proper cast. So it will not be able to implement that method anyways. Better to only have methods we need.

src/main/java/slimeknights/tconstruct/library/recipe/casting/MaterialCastingRecipeBuilder.java
KnightMiner: My comment was to make it safer. If the field type is IMaterialItem, it can never fail, which is better than an instance of check.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
KnightMiner: Lets leave it at 9 ingots for both for now. No sense making all the fluids smaller unless there is a case we need it bigger.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/MaterialCastingRecipeMaker.java
```diff
@@ -0,0 +1,62 @@
...
+    String group = "jei.material.casting.";
+    for (IMaterial material : MaterialRegistry.getInstance().getMaterials()) {
+      if (material.getFluid() == Fluids.EMPTY) continue;
+      for (MaterialCastingRecipe.Table table : materialCastingRecipes) {
+        IMaterialItem result = table.getResult();
+        ItemStack partStack = result.getItemstackWithMaterial(material);
+        String path = String.format("jei.material.casting.%s.%s", Objects.requireNonNull(result.asItem().getRegistryName()).getPath(), Objects.requireNonNull(material.getFluid().getRegistryName()).getPath());
+        ResourceLocation id = new ResourceLocation(TConstruct.modID, path);
+        ItemCastingRecipe.Table tableRecipe = new ItemCastingRecipe.Table(id, group + Objects.requireNonNull(result.asItem().getRegistryName()).getPath(), table.getCast(), FluidIngredient.of(material.getFluid(), table.getFluidAmount()), partStack, ICastingRecipe.calcCoolingTime(material.getTemperature(), table.getFluidAmount()), table.isConsumed(), table.switchSlots());
```
KnightMiner: Leave the group alone, those are supposed to be the same for similar recipes.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ItemCastingRecipe.java
```diff
@@ -0,0 +1,113 @@
...
+import net.minecraft.world.World;
+import net.minecraftforge.fluids.FluidStack;
+import slimeknights.tconstruct.library.recipe.FluidIngredient;
+import slimeknights.tconstruct.library.recipe.RecipeTypes;
+import slimeknights.tconstruct.smeltery.TinkerSmeltery;
+import slimeknights.tconstruct.smeltery.recipe.ICastingInventory;
+
+import java.util.List;
+
+public abstract class ItemCastingRecipe implements ICastingRecipe {
```
KnightMiner, line 21: Consider `@AllArgsConstructor` or `@RequiredArgsConstructor`

src/main/java/slimeknights/tconstruct/library/recipe/casting/ICastingRecipe.java
hanetzer: True, buckets are done btw. May need to adjust the SearedTank item to be able to work with it.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/MaterialCastingRecipeMaker.java
hanetzer: so group as `""` then? Mostly was just adapting JEIs setup.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ContainerFillingRecipe.java
```diff
@@ -0,0 +1,99 @@
...
+  @Getter
+  protected final IRecipeType<?> type;
+  @Getter
+  protected final ResourceLocation id;
+  @Getter
+  protected final String group;
+  @Getter
+  protected final int fluidAmount;
+  @Getter
+  protected final IItemProvider result;
```
KnightMiner: Rename this to container as well, don't like comparing item to to a field named `result` when its the input.

Could also use `Item` here instead of `IItemProvider`, fie, so there is no need to call `asItem()` each time.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ContainerFillingRecipe.java
```diff
@@ -0,0 +1,99 @@
...
+
+  @Override
+  public boolean matches(ICastingInventory inv, World worldIn) {
+    return inv.getStack().getItem() == this.result.asItem();
+  }
+
+  @Override
+  public ItemStack getRecipeOutput() {
+    return ItemStack.EMPTY;
+  }
```
KnightMiner: I suggest returning a stack of the empty container, so at least it has something. Probably also mark deprecated so people know this won't really work right.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ContainerFillingRecipeBuilder.java
```diff
@@ -0,0 +1,102 @@
...
+  public static ContainerFillingRecipeBuilder tableRecipe(IItemProvider result) {
+    return castingRecipe(result, TinkerSmeltery.tableFillingRecipeSerializer.get());
+  }
+
+  public ContainerFillingRecipeBuilder setGroup(String group) {
+    this.group = group;
+    return this;
+  }
+
+  public ContainerFillingRecipeBuilder setFluidAmount(int fluidAmount) {
```
KnightMiner: Could probably put the fluid amount in the main constructor. Furnace recipes have four arguments in the main constructor, excluding the serializer.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ICastingRecipe.java
```diff
@@ -0,0 +1,74 @@
...
+  }
+
+  /**
+   * Gets the amount of fluid required for this recipe
+   * @param inv  Inventory instance
+   * @return  Fluid amount when using the fluid in the inventory
+   */
+  int getFluidAmount(ICastingInventory inv);
+
+  boolean isConsumed();
```
KnightMiner, line 41: Java docs on the unique methods is useful. This one says something like `@returns true if the cast item is consumed when crafting`

src/main/java/slimeknights/tconstruct/library/recipe/casting/ICastingRecipe.java
```diff
@@ -0,0 +1,74 @@
...
+
+  boolean switchSlots();
+
+  int getCoolingTime(ICastingInventory inv);
+
+  int getCoolingTime();
```
KnightMiner: Don't think we need both the inventory sensitive cooling time and the normal one here. Normal is only used for JEI display, which is already set up for specific recipe types.

src/main/java/slimeknights/tconstruct/library/recipe/casting/ItemCastingRecipe.java
```diff
@@ -27,21 +28,27 @@
...
   protected final boolean switchSlots;
 
   @Override
-  public ItemStack getCraftingResult(ICastingInventory inv) {
-    return this.result.copy();
+  public boolean matches(ICastingInventory inv, World worldIn) {
+    return this.getCast().test(inv.getStack()) && this.getFluid(inv).test(inv.getFluid());
+  }
+
+  public FluidIngredient getFluid(ICastingInventory inv) {
```
KnightMiner: What is the reason for this method? Its not in the interface, so we might as well in matches use `this.fluid`

src/main/java/slimeknights/tconstruct/library/recipe/casting/ContainerFillingRecipe.java
hanetzer: How does marking as deprecated work again?

src/main/java/slimeknights/tconstruct/library/recipe/casting/ContainerFillingRecipeBuilder.java
hanetzer: True enough. So I could prolly nix this method entirely?

src/main/java/slimeknights/tconstruct/library/recipe/casting/ContainerFillingRecipe.java
KnightMiner: `@Deprecated` annotation. Can also throw in a java doc, `/** @deprecated use {@link whatever} */` or `/** @deprecated unused */`

src/main/java/slimeknights/tconstruct/library/recipe/casting/ContainerFillingRecipeBuilder.java
KnightMiner: I think it would be cleaner to do so. Means no chance the user forgets a fluid amount.

PR #4148, opened 2020-07-12
Revert "Update creative tabs to a system more in line with deferred r…

hanetzer:
…egisters"

Causes NPE when doing runData.

This reverts commit 83c4919aac2ade080c5b2761f13d36476a4c259a.


KnightMiner:
Don't revert the commit, fix what it broke.

PR #4152, opened 2020-07-14
1.15 - creation of the Titanium material data

ColonelGerdauf:
I hope that you can make use of this.


KnightMiner:
Closing as titanium is not on our current material plans

PR #4144, opened 2020-07-07
Fix gradle/maven issue in 1.12 branch

ghost:
Pretty self explanatory really, just corrects a problem as outlined here: https://github.com/TehNut/HWYLA/issues/253

PR #4153, opened 2020-07-17
The original maven repos are not functional

ColonelGerdauf:
These substitutions need to suffice, as people forking this project will smash their heads against the wall when they are constantly spammed with resolution errors

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -52,9 +52,17 @@ repositories {
         name 'K-4u maven'
         url "https://maven.k-4u.nl/"
     }
-    maven {
+    maven { // backup maven location
         name "Tehnut Maven FS"
-        url "https://tehnut.info/maven"
+        url "https://github.com/TehNut/temporary-maven-thing/tree/master/maven/"
+    }
+    maven { // location of the maven that hosts JEI files
+        name = "Progwml6 maven"
+        url = "https://dvs1.progwml6.com/files/maven/"
+    }
+    maven { // location of a maven mirror for JEI files, as a fallback
+        name = "ModMaven"
+        url = "https://modmaven.k-4u.nl"
     }
     maven {
         name = "CoFH Maven"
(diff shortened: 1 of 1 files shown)


build.gradle
```diff
@@ -52,9 +52,17 @@ repositories {
...
-    maven {
+    maven { // backup maven location
         name "Tehnut Maven FS"
-        url "https://tehnut.info/maven"
+        url "https://github.com/TehNut/temporary-maven-thing/tree/master/maven/"
+    }
+    maven { // location of the maven that hosts JEI files
+        name = "Progwml6 maven"
+        url = "https://dvs1.progwml6.com/files/maven/"
+    }
```
alexbegt, line 58: Progwml6's maven is already in the build.gradle as 'DVS1 Maven FS', please remove the duplicate.

alexbegt (changes requested):

build.gradle
ColonelGerdauf: Done.

ShadowBoxx-LLC:
~~ModMaven is already a repository in the repo list as well:
https://github.com/SlimeKnights/TinkersConstruct/blob/1.12/build.gradle#L41-L44~~
Disregard, that was incorrect. It's not modmaven, it's actually just the the normal maven repository for k-4u.nl

Still, the only issue here is most likely resolved by setting the HWYLA (TeHNuts' repository) to https://maven.tehnut.info, see this PR for an example.
https://github.com/SlimeKnights/TinkersConstruct/pull/4144

This is the correct repo for artifacts for all versions as can be demonstrated:
https://maven.tehnut.info/mcp/mobius/waila/Hwyla/Hwyla-1.10.8-B72_1.15.2/Hwyla-1.10.8-B72_1.15.2.pom
https://maven.tehnut.info/mcp/mobius/waila/Hwyla/Hwyla-1.10.8-B72_1.15.2/Hwyla-1.10.8-B72_1.15.2.jar

PR #4154, opened 2020-07-18
1.12 add events to workbenches

ghost:
This adds events for when recipes are registered to the tool station, tool forge and the stencil table.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/TinkerRegistry.java
+++ b/src/main/java/slimeknights/tconstruct/library/TinkerRegistry.java
@@ -385,7 +385,9 @@ public final class TinkerRegistry {
 
   /** Adds a tool to the Crafting UI of the Tool Station */
   public static void registerToolStationCrafting(ToolCore tool) {
-    toolStationCrafting.add(tool);
+    if(new TinkerRegisterEvent.ToolStationCraftingRegisterEvent(tool).fire()) {
+      toolStationCrafting.add(tool);
+    }
   }
 
   public static Set<ToolCore> getToolStationCrafting() {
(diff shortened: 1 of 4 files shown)


src/main/java/slimeknights/tconstruct/library/TinkerRegistry.java
```diff
@@ -385,7 +385,9 @@ public static void registerToolCrafting(ToolCore tool) {
 
   /** Adds a tool to the Crafting UI of the Tool Station */
   public static void registerToolStationCrafting(ToolCore tool) {
-    toolStationCrafting.add(tool);
+    if(new TinkerRegisterEvent.ToolStationCraftingRegisterEvent(tool).fire()) {
+      toolStationCrafting.add(tool);
+    }
```
KnightMiner: Missing the debug trace if registering fails. Literally do exactly what we do on other events.

ShadowBoxx-LLC:
Last commit was just bringing the branch in line with upstream/1.12 👍

KnightMiner:
Sorry, got distracted and forgot about this PR. Is this still something you need?

Been pretty focused on 1.16 lately, and still not sure we will release another 1.12 build, but worth discussing this if you were waiting on these changes.

ShadowBoxx-LLC:
Ugh... just now found the notification here in my inbox for this... sorry I didn't respond right away.

It's ultimately up to you at this point. I don't know if the mod pack I wrote will ever actually be finished with what little time I have now. So the only thing 'dependent' on these features will likely never be released. If it is, 1.12 is so old at this point, that I may just point people to a custom fork of the mod I have compiled myself.

In the end, it's probably not worth your time messing with it. :)

KnightMiner:
Closing this as we decided to no longer release more 1.12 builds. It has been added to the milestone so if we ever change that decision it will be reviewed again.

PR #4162, opened 2020-08-05
Adds slime textures

TinyAtomss:
Adds Slime related Textures that Match the recent texture update.


mDiyo:
Slime textures have already been updated. These look like they'd go well in a resource pack and the art quality is decent.

PR #4178, opened 2020-08-14
Get the starting tools working for 1.16!

alexbegt:


src/main/java/slimeknights/tconstruct/library/events/TinkerEvent.java
```diff
@@ -0,0 +1,9 @@
+package slimeknights.tconstruct.library.events;
+
+import net.minecraftforge.eventbus.api.Event;
+
+/**
+ * Base class for all tinkers events
+ */
+public abstract class TinkerEvent extends Event {
```
KnightMiner: Do we need a base class for all events? I cannot see someone running an event bus subscriber for any generic Tinkers Construct event.

src/main/java/slimeknights/tconstruct/library/materials/stats/IMaterialStats.java
```diff
@@ -29,7 +29,7 @@
    * </br>
    * This is used to display properties of materials to the user.
    */
-  List<ITextComponent> getLocalizedInfo();
+  List<IFormattableTextComponent> getLocalizedInfo();
```
KnightMiner: Vanilla does return `ITextComponent` in a lot of places. Do we specifically need formattable to get the info with formatting?

src/main/java/slimeknights/tconstruct/library/tinkering/ITinkerable.java
```diff
@@ -13,5 +14,15 @@
   /**
    * The default tooltip for the item
    */
-  void getTooltip(ItemStack stack, List<String> tooltips);
+  void getTooltip(ItemStack stack, List<ITextComponent> tooltips);
```
KnightMiner: How much is the tooltip relevant to all tinkerable items? I almost feel this should be in display. Probably want to talk on discord about a few of these, I just imagine ITinkerable being a pretty minimal interface.

src/main/java/slimeknights/tconstruct/library/tinkering/ITinkerable.java
```diff
@@ -13,5 +14,15 @@
...
+
+  /**
+   * Detailed info about the tool. Displayed when Shift is held]
+   */
+  void getTooltipDetailed(ItemStack stack, List<ITextComponent> tooltips);
```
KnightMiner: Should look into what JEI does, instead of having three methods, have one method and pass an enum para for control and shift.

src/main/java/slimeknights/tconstruct/library/tinkering/PartMaterialRequirement.java
```diff
@@ -50,15 +51,20 @@ public boolean isValidMaterial(IMaterial material) {
...
       if (type.equals(statID)) {
         return true;
       }
     }
 
     return false;
   }
-//
+
+  public Supplier<? extends Item> getPart() {
```
KnightMiner: Does this need to return a supplier, that seems pretty gross to use in my opinion, calling `PartMaterialRequirement.getPart().get()`. just return item and call `.get` in the method.

src/main/java/slimeknights/tconstruct/library/tools/AoeToolCore.java
```diff
@@ -0,0 +1,33 @@
...
+  }
+
+  @Override
+  public ImmutableList<BlockPos> getAOEBlocks(@Nonnull ItemStack stack, World world, PlayerEntity player, BlockPos origin) {
+    return AoeToolInteractionUtil.calculateAOEBlocks(stack, world, player, origin, 1, 1, 1);
+  }
+
+  @Override
+  public boolean isAoeHarvestTool() {
+    return true;
```
KnightMiner: This method is a little weird to me. I know in 1.12 we had it, do you know why? As in, what is the use case for implementing the interface but not being AOE?

src/main/java/slimeknights/tconstruct/library/tools/events/TinkerToolEvent.java
```diff
@@ -0,0 +1,49 @@
...
+  @Cancelable
+  public static class ExtraBlockBreak extends TinkerToolEvent {
+
+    public final PlayerEntity player;
+    public final BlockState state;
+
+    public int width;
+    public int height;
+    public int depth;
+    public int distance;
```
KnightMiner: Make these private with getters and setters as needed

src/main/java/slimeknights/tconstruct/library/tools/events/TinkerToolEvent.java
```diff
@@ -0,0 +1,49 @@
...
+
+    public ExtraBlockBreak(ItemStack itemStack, PlayerEntity player, BlockState state) {
```
KnightMiner: width, height, depth, and distance are not final, but there is no sensible default. Why are they not parameters?

src/main/java/slimeknights/tconstruct/library/utils/TooltipBuilder.java
```diff
@@ -39,7 +39,7 @@ public TooltipBuilder add(String text) {
     return this;
   }
 
-  public TooltipBuilder addMiningSpeed() {
+  public void addMiningSpeed() {
```
KnightMiner: Why switch these all to void? It was set up as a builder, which should be chainable.

src/main/java/slimeknights/tconstruct/plugin/jei/JEIPlugin.java
```diff
@@ -60,4 +67,61 @@ public void registerRecipeCatalysts(IRecipeCatalystRegistration registry) {
     registry.addRecipeCatalyst(new ItemStack(TinkerSmeltery.castingTable), TConstructRecipeCategoryUid.castingTable);
     registry.addRecipeCatalyst(new ItemStack(TinkerSmeltery.searedMelter), TConstructRecipeCategoryUid.melting);
   }
+
+  @Override
+  public void registerItemSubtypes(ISubtypeRegistration registry) {
+    ISubtypeInterpreter toolPartInterpreter = itemStack -> {
+      String meta = itemStack.getDamage() + ":";
```
KnightMiner: Meta is dead. Kill it. In other words, tool part items have no damage value

src/main/java/slimeknights/tconstruct/plugin/jei/JEIPlugin.java
```diff
@@ -60,4 +67,61 @@ public void registerRecipeCatalysts(IRecipeCatalystRegistration registry) {
...
+    registry.registerSubtypeInterpreter(TinkerToolParts.smallBinding.get(), toolPartInterpreter);
+    registry.registerSubtypeInterpreter(TinkerToolParts.toughBinding.get(), toolPartInterpreter);
+    registry.registerSubtypeInterpreter(TinkerToolParts.wideGuard.get(), toolPartInterpreter);
+    registry.registerSubtypeInterpreter(TinkerToolParts.largePlate.get(), toolPartInterpreter);
+    registry.registerSubtypeInterpreter(TinkerToolParts.toolRod.get(), toolPartInterpreter);
+    registry.registerSubtypeInterpreter(TinkerToolParts.toughToolRod.get(), toolPartInterpreter);
+
+    ISubtypeInterpreter toolInterpreter = itemStack -> {
+      StringBuilder builder = new StringBuilder();
+      builder.append(itemStack.getDamage());
```
KnightMiner: We do not need to distinguish tools by damage any more than we do by modifiers I think.

src/main/java/slimeknights/tconstruct/shared/CommonsClientEvents.java
```diff
@@ -1,6 +1,7 @@
 package slimeknights.tconstruct.shared;
 
 import net.minecraft.block.Block;
+import net.minecraft.client.Minecraft;
```
KnightMiner: Why did you just randomly add an import here?

src/main/java/slimeknights/tconstruct/tables/client/inventory/table/PartBuilderScreen.java
```diff
@@ -135,7 +135,7 @@ public void updateDisplay() {
       this.setDisplayForMaterial(materialRecipe.getMaterial());
     } else {
       // default text
-      this.infoPanelScreen.setCaption(this.getTitle());
+      this.infoPanelScreen.setCaption(this.getTitle().copyRaw());
```
KnightMiner: I'd rather copyRaw before you modify it (merge style) than copy raw when you set the caption. Saves effort if we never modify it.

src/main/java/slimeknights/tconstruct/tools/TinkerTools.java
```diff
@@ -30,24 +36,38 @@
...
-    return ToolBuildHandler.buildItemFromMaterials(TinkerTools.pickaxe.get(), toolMats);
+
+    return TinkerTools.pickaxe.get().buildToolForRendering();
   });
 
   /*
    * Items
    */
   public static final ItemObject<PickaxeTool> pickaxe = ITEMS.register("pickaxe", () -> new PickaxeTool(new Item.Properties().group(TAB_TOOLS), ToolDefinitions.PICKAXE));
+  public static final ItemObject<HammerTool> hammer = ITEMS.register("hammer", () -> new HammerTool(new Item.Properties().group(TAB_TOOLS), ToolDefinitions.HAMMER));
```
KnightMiner: We reuse the same item properties, make a private static final variable

src/main/java/slimeknights/tconstruct/library/tools/ToolCore.java
```diff
@@ -142,290 +199,576 @@ public int getHarvestLevel(ItemStack stack, ToolType toolClass, @Nullable Player
...
+    for (int i = 0; i < components.size(); i++) {
+      PartMaterialRequirement requirement = components.get(i);
+      IMaterial material = materials.get(i);
+
+      Item toolPart = requirement.getPart().get();
+
+      if (toolPart instanceof IMaterialItem) {
+        ItemStack partStack = ((IMaterialItem) toolPart).getItemstackWithMaterial(material);
+
+        tooltips.add((new StringTextComponent("")).append(partStack.getDisplayName()).mergeStyle(TextFormatting.UNDERLINE).modifyStyle(style -> style.setColor(material.getColor())));
```
KnightMiner: Ditch the `new StringTextComponent` thing here, switch to `copyRaw`. I did not fully understand the formattable stuff at first. Same below

src/main/java/slimeknights/tconstruct/library/tools/ToolCore.java
```diff
@@ -142,290 +199,576 @@ public int getHarvestLevel(ItemStack stack, ToolType toolClass, @Nullable Player
...
+  }
+
+  /* Information */
 
   @Override
+  @OnlyIn(Dist.CLIENT)
   public void addInformation(ItemStack stack, @Nullable World worldIn, List<ITextComponent> tooltip, ITooltipFlag flagIn) {
-    // todo: evaluate if we want to keep the material info
+    boolean shift = Util.isShiftKeyDown();
+    boolean ctrl = Util.isCtrlKeyDown();
```
KnightMiner, line 450: ITooltipFlag do anything for us? Or is that advanced tooltips?

src/main/java/slimeknights/tconstruct/library/tools/ToolCore.java
```diff
@@ -142,290 +199,576 @@ public int getHarvestLevel(ItemStack stack, ToolType toolClass, @Nullable Player
...
+  /* Trait interactions */
+
+  @Override
+  public void inventoryTick(ItemStack stack, World worldIn, Entity entityIn, int itemSlot, boolean isSelected) {
+    super.inventoryTick(stack, worldIn, entityIn, itemSlot, isSelected);
+
+    this.onUpdateTraits(stack, worldIn, entityIn, itemSlot, isSelected);
+  }
 
+  protected void onUpdateTraits(ItemStack stack, World worldIn, Entity entityIn, int itemSlot, boolean isSelected) {
```
KnightMiner: Why make a separate function for this? Do we expect to override it?

src/main/java/slimeknights/tconstruct/library/tools/ToolCore.java
```diff
@@ -142,290 +199,576 @@ public int getHarvestLevel(ItemStack stack, ToolType toolClass, @Nullable Player
...
+    }
+
+    return super.hitEntity(stack, target, attacker);
+  }
+
+  @Override
+  public Multimap<Attribute, AttributeModifier> getAttributeModifiers(EquipmentSlotType slot, ItemStack stack) {
+    ImmutableMultimap.Builder<Attribute, AttributeModifier> builder = ImmutableMultimap.builder();
+
+    float speed = ToolData.from(stack).getStats().attackSpeedMultiplier;
```
KnightMiner: Create the `ToolData` once, saves the amount of NBT we have to parse. Then call methods from it later

KnightMiner (commented):
Some preliminary comments. As far as general comments:
* More javadoc. Anything that is not private or marked as `@Override` needs javadocs. Feel free to yell at me if I violate this rule
* If you are going to call the same method multiple times in a single method, and its stateless, cache the result. No reason to calculate it multiple times. Big offender here is `ToolData.from`, as that has to parse the whole tool NBT.

src/main/java/slimeknights/tconstruct/library/tinkering/IAoeTool.java
```diff
@@ -22,13 +23,18 @@
...
+  }
 
 
   /**
    * Checks if the AOE tool mines blocks, used for the extra block breaking rendering.
    *
    * @return if the tool is a AOE harvest tool
    */
-  boolean isAoeHarvestTool();
+  default boolean isAoeHarvestTool() {
```
KnightMiner: I still think we should get rid of this method entirely for now. If we do keep this or something similar, we can add it later using a parameter saying what you want, so something can be AOE harvest, AOE interact, etc. I think we have tool types somewhere that would help with that

But anyways, for now lets ditch this method.

src/main/java/slimeknights/tconstruct/library/tools/ToolCore.java
alexbegt: ITooltipFlag is for advanced tooltips

src/main/java/slimeknights/tconstruct/library/tinkering/IAoeTool.java
alexbegt: Has been removed 👍

src/main/java/slimeknights/tconstruct/library/modifiers/IModifier.java
```diff
@@ -54,22 +47,22 @@
...
    * Color tags are not necessary.
    */
-//  String getTooltip(NBTTagCompound modifierTag, boolean detailed);
+  String getTooltip(CompoundNBT modifierTag, boolean detailed);
 
   /**
    * Used for specific modifiers that need a texture variant for each material
    */
-//  @SideOnly(Side.CLIENT)
+  @OnlyIn(Dist.CLIENT)
```
KnightMiner: Should never use `OnlyIn` outside of vanilla extensions.

PR #4184, opened 2020-08-25
initial 1.16.2 port

hanetzer:
There's some features in the Tags classes in 1.16.2 which could be useful for your
desired tag-based item output.


src/main/java/slimeknights/tconstruct/library/Util.java
```diff
@@ -128,7 +127,9 @@ public static String translateFormatted(String key, Object... pars) {
    * Returns a fixed size DEEP copy of the list
    */
   public static NonNullList<ItemStack> deepCopyFixedNonNullList(NonNullList<ItemStack> in) {
-    return RecipeMatchRegistry.copyItemStackArray(in);
+    // TODO: Not sure how to migrate. Not used aside from copyItemStackArray
+    // return RecipeMatchRegistry.copyItemStackArray(in);
+    return in;
```
KnightMiner: Was not aware this was used here. Check to see if its used anywhere in TiC. If it is, copy the code from Mantle before I removed that function. If not, delete this function,

src/main/java/slimeknights/tconstruct/library/tools/helper/AoeToolInteractionUtil.java
```diff
@@ -133,7 +134,7 @@ public static void breakExtraBlock(ItemStack tool, World world, PlayerEntity pla
       if (block.removedByPlayer(state, world, pos, player, true, fluidState)) { // boolean is if block can be harvested, checked above
         block.onPlayerDestroy(world, pos, state);
         block.harvestBlock(world, player, pos, state, tileEntity, tool);
-        block.dropXpOnBlockBreak(world, pos, xp);
+        block.dropXpOnBlockBreak((ServerWorld) world, pos, xp);
```
KnightMiner: I don't like this unchecked cast. On the odd chance some non-server world returns false for isRemote, I think its a good idea to do an instanceof check here. Really should have been done above for the player entity too.

src/main/java/slimeknights/tconstruct/library/traits/ITrait.java
```diff
@@ -60,7 +60,8 @@
    * Called after a block has been broken. See {@link net.minecraftforge.event.world.BlockEvent.HarvestDropsEvent}
    * Note that, as opposed to the original event, this only gets called with a player.
    */
-  void blockHarvestDrops(ItemStack tool, BlockEvent.HarvestDropsEvent event);
+  // TODO: Event is gone?
+  // void blockHarvestDrops(ItemStack tool, BlockEvent.HarvestDropsEvent event);
```
KnightMiner, line 64: Yeah, event has been deprecated since 1.14. We are going to have to use global loot tables for traits I think. It will be interesting, but at least easier for mod compat

Will have to see if an event exists that lets us cancel the drops to redirect them, may be worth adding that here instead

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/InfoPanelScreen.java
```diff
@@ -212,7 +213,8 @@ protected void updateSliderParameters() {
         continue;
       }
 
-      lines.addAll(this.font.func_238425_b_(textComponent, w));
+      // TODO: Not sure
+      // lines.addAll(this.font.func_238425_b_(textComponent, w));
```
KnightMiner: Lines might need to be a list of IReorderingProcessor, pretty sure a function exists to return those. Vanilla UI has some stuff

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsScreen.java
```diff
@@ -30,11 +30,13 @@ public void addSideButton(Button button) {
     int rows = (this.buttonCount - 1) / this.columns + 1;
 
     this.xSize = button.getWidth() * this.columns + this.spacing * (this.columns - 1);
-    this.ySize = button.getHeight() * rows + this.spacing * (rows - 1);
+    // TODO: getWidth_CLASH->getHeight()
+    this.ySize = button.getWidth_CLASH() * rows + this.spacing * (rows - 1);
```
KnightMiner, line 34: `getHeight` still exists. For some dumb reason Mojang has two getters for height, one of which had a mapping conflict.

src/main/java/slimeknights/tconstruct/library/Util.java
hanetzer: Looks like this whole block is unused atm. Nixed.

src/main/java/slimeknights/tconstruct/library/tools/helper/AoeToolInteractionUtil.java
hanetzer: wrapping it in an instanceof check seems to work. Need to test on a real server tho.

src/main/java/slimeknights/tconstruct/library/traits/ITrait.java
hanetzer: So what should I do here? Just leave it commented out for later stuff?

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/InfoPanelScreen.java
hanetzer: Yeah, I tried some variations of that and ended up with rendered tooltips but no
actual normal text. Investigating more.

build.gradle
```diff
@@ -216,7 +216,14 @@ processResources {
     from(sourceSets.main.resources.srcDirs) {
         include 'META-INF/mods.toml'
 
-        expand 'version': artifact_version, 'MCversion': minecraft_version, 'forge_version': forge_version_min
+        expand 'version': artifact_version,
+            'minecraft_version_min': minecraft_version_min,
```
KnightMiner: We could put two of these per line (min and max). I just don't want a giant line of 8 variables

src/main/java/slimeknights/tconstruct/common/TinkerTags.java
```diff
@@ -126,11 +126,11 @@
     public static final INamedTag<EntityType<?>> SLIMES = forgeTag("slimes");
 
     private static INamedTag<EntityType<?>> tag(String name) {
-      return EntityTypeTags.func_232896_a_(Util.resource(name));
+      return EntityTypeTags.getTagById(Util.resource(name));
```
KnightMiner: We should switch to `createOptionalTag` for all our tags. New Forge feature that solves tag conflicts, see the `Tags` class for an example.

src/main/java/slimeknights/tconstruct/tables/client/inventory/library/slots/SlotPosition.java
```diff
@@ -37,7 +37,7 @@
 
     // array: multiple cubes
     if (json.isJsonArray()) {
-      return JsonHelper.parseList(json.getAsJsonArray(), key, JSONUtils::getJsonObject, SlotPosition::fromJson);
+      return JsonHelper.parseList(json.getAsJsonArray(), key, (j) -> SlotPosition.fromJson(j));
```
KnightMiner: The `SlotPosition` qualifier is unneeded, since we are `SlotPosition`

src/main/java/slimeknights/tconstruct/common/TinkerTags.java
hanetzer: Not finding a `Tags` class

hanetzer: nvm, found it.

KnightMiner: Its in Forge

src/main/java/slimeknights/tconstruct/library/traits/ITrait.java
KnightMiner: Yeah, lets leave it commented out for now. We can bring it back later with a better plan

PR #4185, opened 2020-08-25
Begin working on the tinker station

alexbegt:
Currently only supports building tools and nothing else


src/main/java/slimeknights/tconstruct/common/data/TConstructBlockLootTables.java
```diff
@@ -96,6 +96,12 @@ private void addDecorative() {
 
   private void addTools() {
     this.registerDropSelfLootTable(TinkerTables.craftingStation.get());
+    for (Block block : new Block[] {TinkerTables.tinkerStation.get()}) {
```
KnightMiner: Isn't this the same as the part builder below?

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
```diff
@@ -0,0 +1,46 @@
...
+import net.minecraft.inventory.IInventory;
+import net.minecraft.item.ItemStack;
+import net.minecraft.item.crafting.IRecipe;
+import net.minecraft.item.crafting.IRecipeType;
+import net.minecraft.world.World;
+import slimeknights.tconstruct.library.recipe.RecipeTypes;
+import slimeknights.tconstruct.library.recipe.ValidationResult;
+import slimeknights.tconstruct.tables.tileentity.table.tinkerstation.ITinkerStationInventory;
+
+public interface ITinkerStationRecipe extends IRecipe<ITinkerStationInventory> {
```
KnightMiner: Extend `ICommonRecipe` from Mantle, it removes the need to implement some of the useless mojang methods by defaulting them.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
```diff
@@ -0,0 +1,46 @@
...
+
+  /** @deprecated unused */
+  @Deprecated
+  @Override
+  default boolean canFit(int width, int height) {
+    return true;
+  }
+
+  /** If true, this recipe matches the given inputs, ignoring current tool state */
+  boolean matches(ITinkerStationInventory inv, World world);
```
KnightMiner, line 25: Should have an `@Override` I think, as this overrides the original matches to add the new javadoc. If its not overriding, fix it to override

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -0,0 +1,53 @@
...
+import slimeknights.tconstruct.library.tools.ToolBuildHandler;
+import slimeknights.tconstruct.library.tools.ToolCore;
+import slimeknights.tconstruct.tables.TinkerTables;
+import slimeknights.tconstruct.tables.tileentity.table.tinkerstation.ITinkerStationInventory;
+
+import java.util.List;
+import java.util.stream.Collectors;
+
+@AllArgsConstructor
+public class ToolBuildingRecipe implements ITinkerStationRecipe {
```
KnightMiner, line 25: Since we are going to have a lot of these tinker station recipes, might be good to add a javadoc on the top saying what this specific recipe does. "Recipe used for crafting a set of parts into a tool", or "Dynamic recipe used to repair tools"

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipeSerializer.java
```diff
@@ -0,0 +1,53 @@
...
+    ToolCore item = RecipeHelper.deserializeItem(JSONUtils.getString(json, "output"), "output", ToolCore.class);
+
+    return new ToolBuildingRecipe(recipeId, group, item);
+  }
+
+  @Nullable
+  @Override
+  public ToolBuildingRecipe read(ResourceLocation recipeId, PacketBuffer buffer) {
+    try {
+      String group = buffer.readString(32767);
```
KnightMiner: Use `Short.MAX_VALUE`, little better for the in code docs

src/main/java/slimeknights/tconstruct/library/tinkering/ITinkerable.java
```diff
@@ -1,6 +1,7 @@
 package slimeknights.tconstruct.library.tinkering;
 
 import net.minecraft.item.ItemStack;
+import net.minecraft.util.IItemProvider;
```
KnightMiner: Why add this import?

src/main/java/slimeknights/tconstruct/tables/client/SlotInformationLoader.java
```diff
@@ -0,0 +1,75 @@
...
+  /** GSON instance for this */
+  private static final Gson GSON = new GsonBuilder()
+    .setPrettyPrinting()
+    .disableHtmlEscaping()
+    .create();
+
+  /** Singleton instance */
+  public static final SlotInformationLoader INSTANCE = new SlotInformationLoader();
+
+  public static final ResourceLocation REPAIR_NAME = Util.getResource("repair");
```
KnightMiner: I would move this constant to the screen, as nothing about the info loader special cases repair

src/main/java/slimeknights/tconstruct/tables/client/inventory/library/slots/SlotPosition.java
```diff
@@ -0,0 +1,61 @@
...
+
+    return new SlotPosition(vec[0], vec[1]);
+  }
+
+  public static SlotPosition fromJson(JsonObject json) {
+    return arrayToPosition(json, "position");
+  }
+
+  public boolean isHidden() {
+    return this.x == 0 && this.y == 0;
```
KnightMiner: Switch this to -1, -1 I think. I could see 0,0 technically being wanted, but -1, -1 is outside the UI

src/main/java/slimeknights/tconstruct/tables/client/inventory/library/slots/SlotPosition.java
```diff
@@ -0,0 +1,61 @@
...
+    int[] vec = new int[2];
+    for (int i = 0; i < vec.length; ++i) {
+      vec[i] = JSONUtils.getInt(array.get(i), name + "[" + i + "]");
+    }
+
+    return new SlotPosition(vec[0], vec[1]);
+  }
+
+  public static SlotPosition fromJson(JsonObject json) {
+    return arrayToPosition(json, "position");
```
KnightMiner: Something feels slightly off about this. Is a slot position a pair of two numbers (arrayToPosition), or is it the object containing the position key? If this method is just a helper to set the default name, I suggest using the same name for both methods

src/main/java/slimeknights/tconstruct/tables/client/inventory/library/slots/SlotPosition.java
```diff
@@ -0,0 +1,61 @@
...
+import java.util.List;
+
+@AllArgsConstructor
+@Getter
+public class SlotPosition {
+
+  private final int x;
+  private final int y;
+
+  public static List<SlotPosition> listFromJson(JsonObject parent, String key) {
```
KnightMiner, line 30: Javadocs for public methods here

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/TinkerStationButtonsScreen.java
```diff
@@ -0,0 +1,116 @@
...
+    super.updatePosition(parentX, parentY, parentSizeX, parentSizeY);
+  }
+
+  public void addInfoButton(ButtonItem<SlotInformation> buttonItem) {
+    this.shiftButton(buttonItem, 0, -18 * this.style);
+    this.addSideButton(buttonItem);
+  }
+
+  @SuppressWarnings("unchecked")
+  public void wood() {
```
KnightMiner: This method and the one below seem a bit redundant. Why not "shiftButtons(int style)" with some constants for the styles?

src/main/java/slimeknights/tconstruct/tables/inventory/BaseStationContainer.java
```diff
@@ -92,51 +93,45 @@ public void detectStationParts(World world, BlockPos start) {
...
    * Tells the client to display the LOCALIZED error message
    */
   public void error(final IFormattableTextComponent message) {
-    if (this.tile != null) {
-      if (this.tile.getWorld() != null) {
-        if (tile.getWorld().isRemote) {
-          Minecraft.getInstance().execute(() -> TinkerStationContainer.clientError(message));
-        }
-      }
+    assert this.tile != null;
```
KnightMiner: Not sure asserts are better here. A null check says "if this happens, ignore it". An assert says "if this happens, crash". I think we can safely ignore the error message instead of crashing, even if its unlikely here

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/TinkerStationInventoryWrapper.java
```diff
@@ -0,0 +1,57 @@
...
+  private NonNullList<ItemStack> inputs;
+
+  public TinkerStationInventoryWrapper(IInventory station) {
+    this.station = station;
+  }
+
+  @Override
+  public NonNullList<ItemStack> getAllInputStacks() {
+    if (this.inputs == null) {
+      this.inputs = NonNullList.from(ItemStack.EMPTY, IntStream.range(0, 6).mapToObj(this.station::getStackInSlot).filter(itemStack -> !itemStack.isEmpty()).toArray(ItemStack[]::new));
```
KnightMiner: Split this over a few lines to make it more readible. As in,
```java
this.inputs = NonNullList.from(ItemStack.EMPTY, IntStream.range(0, 6)
        .mapToObj(this.station::getStackInSlot)
        .filter(itemStack -> !itemStack.isEmpty())
        .toArray(ItemStack[]::new));
```

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/ITinkerStationInventory.java
```diff
@@ -0,0 +1,14 @@
+package slimeknights.tconstruct.tables.tileentity.table.tinkerstation;
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.NonNullList;
+import slimeknights.mantle.recipe.inventory.IReadOnlyInventory;
+
+public interface ITinkerStationInventory extends IReadOnlyInventory {
+
+  NonNullList<ItemStack> getAllInputStacks();
```
KnightMiner, line 15: Java doc here saying what the method returns is important

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/ITinkerStationInventory.java
```diff
@@ -0,0 +1,14 @@
...
+
+  void clearInputs();
```
KnightMiner: Should never be called by recipes. Keep in the wrapper, but remove from the interface

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/ITinkerStationInventory.java
```diff
@@ -0,0 +1,14 @@
+package slimeknights.tconstruct.tables.tileentity.table.tinkerstation;
+
+import net.minecraft.item.ItemStack;
+import net.minecraft.util.NonNullList;
+import slimeknights.mantle.recipe.inventory.IReadOnlyInventory;
+
+public interface ITinkerStationInventory extends IReadOnlyInventory {
```
KnightMiner: Since we are using primarily `getAllInputStacks` and `getTinkerableStack`, I think deprecating and defaulting the vanilla methods is good. Notably `getStackInSlot`.

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/TinkerStationTileEntity.java
```diff
@@ -0,0 +1,215 @@
...
+      // first, try the cached recipe
+      ITinkerStationRecipe recipe = lastRecipe;
+      // if it does not match, find a new recipe
+      if (recipe == null || !recipe.matches(this.inventoryWrapper, this.world)) {
+        recipe = manager.getRecipe(RecipeTypes.TINKER_STATION, this.inventoryWrapper, this.world).orElse(null);
+      }
+
+      // if we have a recipe, fetch its result
+      if (recipe != null) {
+        result = recipe.getCraftingResult(this.inventoryWrapper);
```
KnightMiner: We need to call validate at some point to prevent the result if the recipe does not validate. Maybe here? We really need the result of validate in the screen, could always send it in a packet from this method, cache it for new clients to open the container

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/TinkerStationTileEntity.java
```diff
@@ -0,0 +1,215 @@
...
+    // remove remaining items
+    NonNullList<ItemStack> remaining = this.lastRecipe.getRemainingItems(this.inventoryWrapper);
+
+    for (int i = 0; i < remaining.size(); ++i) {
+      ItemStack original = getStackInSlot(i);
+      ItemStack newStack = remaining.get(i);
+
+      // if the slot contains a stack, decrease by 1
+      if (!original.isEmpty()) {
+        original.shrink(1);
```
KnightMiner: Going to need a way for the recipe to specify a higher number eventually

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/TinkerStationTileEntity.java
```diff
@@ -0,0 +1,215 @@
...
+    }
+    else if (this.lastRecipe != null && this.lastRecipe.matches(this.inventoryWrapper, world)) {
+      result = this.lastRecipe.getCraftingResult(this.inventoryWrapper);
+    }
+
+    return result;
+  }
+
+  @Override
+  public ItemStack onCraft(PlayerEntity player, ItemStack result, int amount) {
```
KnightMiner, line 130: A lot of this is pretty similar to the crafting station. I wonder if we could not move the code to a base class somehow, or put it in the lazy result inventory logic. Can always deal with that after merging if you prefer

alexbegt: I feel we should figure this out after merging the base table in

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/TinkerStationTileEntity.java
alexbegt: Yes, I think we should do that once we start working on the repair recipe and such

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/TinkerStationButtonsScreen.java
```diff
@@ -0,0 +1,110 @@
...
+  public static final int WOOD_STYLE = 2;
+  public static final int METAL_STYLE = 1;
+
+  public TinkerStationButtonsScreen(TinkerStationScreen parent, Container container, PlayerInventory playerInventory, ITextComponent title) {
+    super(parent, container, playerInventory, title, TinkerStationScreen.COLUMN_COUNT, false);
+
+    this.parent = parent;
+  }
+
+  @SuppressWarnings("unchecked")
```
KnightMiner: Any reason this is unchecked? Only cases I see could also be written as `ButtonItem<?>`

src/main/java/slimeknights/tconstruct/tables/inventory/table/tinkerstation/TinkerStationContainer.java
```diff
@@ -0,0 +1,109 @@
...
+
+  public void setToolSelection(int activeSlots, boolean mainSlotHidden) {
+    assert this.tile != null;
+
+    if (activeSlots > this.tile.getSizeInventory()) {
+      activeSlots = this.tile.getSizeInventory();
+    }
+
+    for (int i = 0; i < this.tile.getSizeInventory(); i++) {
+      Slot slot = this.inventorySlots.get(i);
```
KnightMiner, line 81: Wouldn't it be easier to reference variables rather than doing a ton of instanceof checks? We have a tinker slot variable, and can make one for input slots, both with the proper generics.

src/main/java/slimeknights/tconstruct/tables/network/UpdateStationScreenPacket.java
```diff
@@ -0,0 +1,71 @@
...
+
+  public UpdateStationScreenPacket(PacketBuffer buffer) {
+    this.type = PacketType.valueOf(buffer.readString());
+    this.message = buffer.readTextComponent();
+  }
+
+  @Override
+  public void encode(PacketBuffer packetBuffer) {
+    packetBuffer.writeString(this.type.toString());
+    packetBuffer.writeTextComponent(this.message);
```
KnightMiner: I wonder if its worth saving some packet space by not syncing the message if its `StringTextComponent.EMPTY`. You can do that with the following:

```java
if (message == StringTextComponent.EMPTY) {
  buffer.writeBoolean(false);
} else {
  buffer.writeBoolean(true);
  buffer.writeTextComponent(this.message);
}
```

Then read with:
```java
if (buffer.readBoolean()) {
  this.message = buffer.readTextComponent();
} else {
  this.message = StringTextComponent.EMPTY;
}
```

src/main/java/slimeknights/tconstruct/tables/network/UpdateStationScreenPacket.java
```diff
@@ -0,0 +1,71 @@
...
+  }
+
+  public UpdateStationScreenPacket(PacketBuffer buffer) {
+    this.type = PacketType.valueOf(buffer.readString());
+    this.message = buffer.readTextComponent();
+  }
+
+  @Override
+  public void encode(PacketBuffer packetBuffer) {
+    packetBuffer.writeString(this.type.toString());
```
KnightMiner: `buffer::writeEnum`

src/main/java/slimeknights/tconstruct/tables/tileentity/table/tinkerstation/TinkerStationTileEntity.java
KnightMiner: One idea is making a method in `ITinkerStationRecipe`:

```java
default int getAmountUsed(int index) {
  return 1;
}
```
We then call the method here by incrementing the index only when we see a non empty slot that is not the tool slot.

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/TinkerStationButtonsScreen.java
alexbegt: I can remove the suppress warning however, `ButtonItem<SlotInformation> buildScreenInfoButtonItem = (ButtonItem<SlotInformation>) button;` is unchecked as it's an unchecked cast for Button to `ButtonItem<SlotInformation>`

KnightMiner: Any way we can maintain a list of `ButtonItem<SlotInformation>` somewhere instead?

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
```diff
@@ -30,7 +34,17 @@ default ItemStack getCraftingResult(ITinkerStationInventory inv) {
...
-    return 1;
+  /**
+   * Handles subtracting the inputs used by the recipe and passes any extra items (container items) to the consumer
+   *
+   * @param inventory the actual items in the inventory
+   * @param extraStackConsumer handles the extra stacks if they are unable to be added to the inventory
+   */
+  void consumeInputs(List<ItemStack> inventory, Consumer<ItemStack> extraStackConsumer);
+
+  /** Unused, please call consumeInputs */
```
KnightMiner: I typically would write this as `@deprecated use {@link #consumeInputs(List,Consumer)}` so you get the fancy javadoc formatting

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -37,7 +37,7 @@
 
   @Override
   public boolean matches(ITinkerStationInventory inv, World worldIn) {
-    return ToolBuildHandler.canToolBeBuilt(inv.getAllInputStacks(), this.output);
+    return ToolBuildHandler.canToolBeBuilt(inv.getAllInputStacks(), this.output) && inv.getTinkerableStack().isEmpty();
```
KnightMiner: Check the tinkerable stack first, that will run a little faster than the "can tool be built" logic

PR #4201, opened 2020-10-08
Updated versions and fixed build for 1.16.3

Tankonor:
Fixed the build issues and updated dependency versions.

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -3,7 +3,7 @@
 org.gradle.jvmargs=-Xmx3G
 org.gradle.daemon=false
 
-mod_version=3.0.1
+mod_version=3.0.2
 
 # Minecraft Version Information
 minecraft_base_version=1.16
@@ -14,8 +14,8 @@ minecraft_version_max=1.17
 # Forge Version Information
 javafml_min=33.0
 javafml_max=35.0
-forge_version_min=33.0.0
-forge_version=34.0.13
+forge_version_min=34.0.0
+forge_version=34.1.12
 forge_version_max=35.0.0
 
 # Mappings Information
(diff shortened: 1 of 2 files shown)


Tankonor:
Should clarify, this is for 1.16.3

KnightMiner:
Can you explain the reason for this PR? Yes, we can update versions, but why? It builds just fine with the current versions, and being built against the current versions still allows it to run with the latest Forge.

My general policy on version changes in PRs is only do it if needed for another feature (e.g. a PR needs a feature from the latest Forge) or if part of a substantial update (e.g., porting to 1.16.4).

gradle.properties
```diff
@@ -14,15 +14,15 @@ minecraft_version_max=1.17
...
 
 # Mappings Information
 mappings_version=20200916-1.16.2
 
 # Build Dependencies
 mantle_version=1.6.38
-jei_version=7.1.+
-hwyla_version=1.10.10-B77
-probe_version=3.0.2-beta-5
+jei_version=7.5.0.43
```
AlphonseSantoro: Restricting versions to a specific one is never a good idea for compatibility issues, unless you have a very good reason then this should be `7.5.+`

Tankonor:
> Can you explain the reason for this PR? Yes, we can update versions, but why? It builds just fine with the current versions, and being built against the current versions still allows it to run with the latest Forge.
> 
> My general policy on version changes in PRs is only do it if needed for another feature (e.g. a PR needs a feature from the latest Forge) or if part of a substantial update (e.g., porting to 1.16.4).

It seems the last PR into master broke the build (project is not currently building) so I wanted to fix it.

Build ref:
https://dvs1.progwml6.com/jenkins/blue/organizations/jenkins/SlimeKnights%2FTinkersConstruct/detail/1.16/19/pipeline

gradle.properties
Tankonor: No problem, will correct this!

KnightMiner:
Build did not fail because of build.gradle, it randomly failed to connect to all Maven repos, versions had nothing to do with it.

Restarted the build and 20 worked: https://dvs1.progwml6.com/jenkins/blue/organizations/jenkins/SlimeKnights%2FTinkersConstruct/detail/1.16/20/pipeline

Since the build worked fine, I'm going to close this. As I said, in general we don't take PRs that just update build numbers, the time it takes to review it is much more than it would take to just do it ourselves, so only update build numbers in a PR if its relevant to the actual PR content.

PR #4208, opened 2020-10-19
1.16 tool repair

KnightMiner:
This branch both fixes tools not being repairable, and not properly breaking (including wrong breaking particles)

----

I do still need to fix the part models so they properly handle the broken state. I have an idea of how to do it, but want to wait until after these changes are merged as more people are waiting on these.


alexbegt (approved):
Looks good 👍

PR #4244, opened 2021-01-14
Add shurikens

ghost:
Shurikens re-implemented through Quartz and Flint variations. Knockback may need to be tested and modified.

Note: Rendering is not entirely finished. This is another thing that needs to be tested, and if we need to write more compatibility, we can, but a few lines from the previous implementation of Shurikens seem to add a decently realistic rendering. Original rendering used G11, and has been converted to MatrixStacks and Quaternions. See RenderShuriken class.


src/main/java/slimeknights/tconstruct/gadgets/data/GadgetRecipeProvider.java
```diff
@@ -161,6 +161,20 @@ protected void registerRecipes(Consumer<IFinishedRecipe> consumer) {
                        .addCriterion("has_item", hasItem(Tags.Items.DUSTS_GLOWSTONE))
                        .build(consumer, prefix(TinkerGadgets.glowBall, "gadgets/throwball/"));
 
+    // Shurikens
+    ShapedRecipeBuilder.shapedRecipe(TinkerGadgets.flintShuriken.get(), 4)
+                        .key('X', Items.FLINT)
+                        .patternLine(" X ")
+                        .patternLine("X X")
+                        .patternLine(" X ")
+                        .build(consumer, prefix(TinkerGadgets.flintShuriken, "gadgets/flint_shuriken/"));
```
KnightMiner: this filename mismatches with the committed file above. prefix should automatically add `flint_shuriken` for you, so I'd expect if you rerun datagen you'd get `gadgets/flint_shuriken/flint_shuriken` as the full recipe name.

src/main/java/slimeknights/tconstruct/gadgets/entity/ThrowableItem.java
```diff
@@ -0,0 +1,28 @@
...
+import slimeknights.mantle.util.TranslationHelper;
+import slimeknights.tconstruct.gadgets.TinkerGadgets;
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+public abstract class ThrowableItem extends SnowballItem {
+
+  public ThrowableItem() {
+    super((new Properties()).maxStackSize(16).group(TinkerGadgets.TAB_GADGETS));
```
KnightMiner: Don't hardcode item properties, pass them in in `TinkersGadgets`

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/ShurikenEntityBase.java
```diff
@@ -0,0 +1,62 @@
...
+
+  public ShurikenEntityBase(EntityType<? extends ShurikenEntityBase> type, double x, double y, double z, World worldIn) {
+    super(type, x, y, z, worldIn);
+  }
+
+  public ShurikenEntityBase(EntityType<? extends ShurikenEntityBase> type, LivingEntity livingEntityIn, World worldIn) {
+    super(type, livingEntityIn, worldIn);
+  }
+
+  public abstract float getDamage();
```
KnightMiner, line 42: I tend to like javadoc comments on public methods. Getters are a bit obvious, but I'd rather be consistent than make exceptions. Its a bit less obvious for knockback, helps to be clear the return value is in blocks or alike

src/main/java/slimeknights/tconstruct/gadgets/item/FlintShurikenItem.java
```diff
@@ -0,0 +1,34 @@
...
+  public ActionResult<ItemStack> onItemRightClick(World worldIn, PlayerEntity playerIn, Hand handIn) {
+    ItemStack itemStack = playerIn.getHeldItem(handIn);
+    if (!playerIn.abilities.isCreativeMode) {
+      itemStack.shrink(1);
+    }
+
+    playerIn.getCooldownTracker().setCooldown(itemStack.getItem(), 4);
+
+    if(!worldIn.isRemote) {
+      FlintShurikenEntity entity = new FlintShurikenEntity(worldIn, playerIn);
```
KnightMiner: This line is the only difference between the two classes. I suggest making one `ShurikenItem` class with a `BiFunction<ShurikenEntity,World,PlayerEntity> entity` parameter. Will let you pass `FlintShurikenEntity::new` into the constructor and call the function here.

src/main/java/slimeknights/tconstruct/gadgets/entity/ThrowableItem.java
```diff
@@ -0,0 +1,28 @@
...
+import net.minecraft.world.World;
+import net.minecraftforge.api.distmarker.Dist;
+import net.minecraftforge.api.distmarker.OnlyIn;
+import slimeknights.mantle.util.TranslationHelper;
+import slimeknights.tconstruct.gadgets.TinkerGadgets;
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+public abstract class ThrowableItem extends SnowballItem {
```
KnightMiner: I feel throwballs should also extend this class, or its a bit pointless abstraction. Especially since you did not make use of the optional tooltip. May be unneeded with the comment on FlintShurikenItem above

Additionally, wrong package for an item

KnightMiner (changes requested):

src/main/java/slimeknights/tconstruct/gadgets/data/GadgetRecipeProvider.java
ghost: Got it. I actually think it should be `gadgets/shuriken` from looking at how the other item recipes are registered and how the current folder structures look.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/ShurikenEntityBase.java
ghost: Done! I don't know exactly what the value is in blocks, but I know some typical values that Minecraft used, so I'll try to get some useful info in there.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/FlintShurikenEntity.java
```diff
@@ -0,0 +1,37 @@
...
+
+  public FlintShurikenEntity(EntityType<? extends FlintShurikenEntity> p_i50159_1_, World p_i50159_2_) {
+    super(p_i50159_1_, p_i50159_2_);
+  }
+
+  public FlintShurikenEntity(World worldIn, LivingEntity throwerIn) {
+    super(TinkerGadgets.flintShurikenEntity.get(), throwerIn, worldIn);
+  }
+
+  public FlintShurikenEntity(EntityType<? extends QuartzShurikenEntity> type, LivingEntity livingEntityIn, World worldIn) {
```
KnightMiner: Looks like the wrong entity type there, should be Flint. Though, are these two constructors taking an entity type even used? Can probably ditch them, they are mostly for inheritance (and extensions should use `ShurikenEntityBase`)

src/main/java/slimeknights/tconstruct/gadgets/TinkerGadgets.java
```diff
@@ -115,6 +118,13 @@
   public static final ItemObject<Item> soggySpaghetti = ITEMS.register("soggy_spaghetti", SPAGET_PROPS);
   public static final ItemObject<Item> coldSpaghetti = ITEMS.register("cold_spaghetti", SPAGET_PROPS);
 
+  // Shuriken Properties -- TODO potentially make "THROWABLE_PROPS" or similar for other throwballs
```
KnightMiner: Can probably just name it throwable props and leave out the TODO

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/RenderShuriken.java
```diff
@@ -0,0 +1,45 @@
+package slimeknights.tconstruct.gadgets.entity.shuriken;
```
KnightMiner: Wrong package, we have a client package for things like this

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/RenderShuriken.java
```diff
@@ -0,0 +1,45 @@
...
+    // TODO Figure out rotation stuff if the current isn't good enough
+    // original credits to Boni
+    matrixStackIn.rotate(new Quaternion(1.0F, entity.rotationYaw, 0F, true));
+    matrixStackIn.rotate(new Quaternion(-entity.rotationPitch, 1.0F, 0F, true));
+
+    // add some diversity
+    matrixStackIn.rotate(new Quaternion(entity.rotationPitch, 0F, 1.0F, true));
+
+    // rotate it into a horizontal position
+    matrixStackIn.rotate(new Quaternion(90.0F, 0.0F, 0.0F, true));
```
KnightMiner: Something is off with the rotation, I get a vertical sideview when throwing it east or west and a horizontal sideview when throwing it north or south. Look at the base class for sprite renderer, maybe you can just undo those rotations by applying the opposite. Or reimplement that class instead of extending it to replace the rotations

I did not notice a lack of spin at all, moved too fast, so we might be able to get away without it.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/ShurikenEntityBase.java
```diff
@@ -0,0 +1,73 @@
...
+  /**
+   * Get knockback dealt by Shuriken
+   * Should be <= 1.0F, Minecraft
+   * typically uses values from 0.2F-0.6F
+   * @return float knockback
+   */
+  public abstract float getKnockback();
+
+  @Override
+  protected void onEntityHit(EntityRayTraceResult result) {
```
KnightMiner, line 70: Entity does not die on first impact. You want to override `onImpact` like `SnowballEntity` did. I suggest also overriding `onBlockHit` to drop this as an item

src/main/resources/assets/tconstruct/lang/en_us.json
```diff
@@ -341,6 +341,9 @@
     "item.tconstruct.efln_ball": "EFLN",
     "item.tconstruct.efln_ball.tooltip": "Explodes! Safe for mining",
 
+    "item.tconstruct.quartz_shuriken": "Quartz Shuriken",
+    "item.tconstruct.flint_shuriken": "Flint Shuriken",
```
KnightMiner, line 346: Maybe want tooltips? Could do something like "Throwable weapon with high knockback" for flint and "Throwable weapon with higher damage" for quartz. To make the differences more visible.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/FlintShurikenEntity.java
ghost: Fixed. That constructor wasn't even needed, may bad.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/RenderShuriken.java
ghost: Couldn't seem to figure this one out, I've never worked with Quaternions. If you could take a look, that would be great.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/ShurikenEntityBase.java
ghost: Fixed it to drop as an item now - and in all of my testing, the entity did die on first impact, so let me know if that was fixed as well.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/ShurikenEntityBase.java
```diff
@@ -0,0 +1,103 @@
...
+   */
+  public abstract ShurikenEntityBase getShurikenEntity();
+
+  @Override
+  protected void onImpact(RayTraceResult result) {
+    super.onImpact(result);
+
+    if (!this.world.isRemote) {
+      if (result.getType() == RayTraceResult.Type.BLOCK) {
+        if(getShurikenEntity() instanceof FlintShurikenEntity) {
```
KnightMiner: Remember `getDefaultItem()`?

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/ShurikenEntityBase.java
```diff
@@ -0,0 +1,103 @@
...
+   * @return ShurikenEntityBase entity
+   */
+  public abstract ShurikenEntityBase getShurikenEntity();
+
+  @Override
+  protected void onImpact(RayTraceResult result) {
+    super.onImpact(result);
+
+    if (!this.world.isRemote) {
+      if (result.getType() == RayTraceResult.Type.BLOCK) {
```
KnightMiner: Handled in super class. Code is as follows:

```java
      RayTraceResult.Type raytraceresult$type = result.getType();
      if (raytraceresult$type == RayTraceResult.Type.ENTITY) {
         this.onEntityHit((EntityRayTraceResult)result);
      } else if (raytraceresult$type == RayTraceResult.Type.BLOCK) {
         this.func_230299_a_((BlockRayTraceResult)result);
      }
```

So looks like `func_230299_a_` is the method you want to override for the block hit.

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/QuartzShurikenEntity.java
```diff
@@ -0,0 +1,38 @@
...
+    return 5.0F;
+  }
+
+  @Override
+  public float getKnockback() {
+    return 0.4F;
+  }
+
+  @Override
+  public ShurikenEntityBase getShurikenEntity() {
```
KnightMiner: I really don't get this method at all. `QuartzShurikenEntity` is already of type `ShurikenEntityBase`, so why do you need to implement a method to return it?

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/ShurikenEntityBase.java
ghost: Oops..!

src/main/java/slimeknights/tconstruct/gadgets/entity/shuriken/QuartzShurikenEntity.java
ghost: This was partly with the getDefaultItem() thing; will be removed.

KnightMiner (approved):

PR #4249, opened 2021-01-20
1.12 Fixed 'slimeIslandsOnlyGenerateInSurfaceWorlds' config option

jpark651:
The 'slimeIslandsOnlyGenerateInSurfaceWorlds' config option was inverted in the code: to make slime islands only generate in surface worlds, 'slimeIslandsOnlyGenerateInSurfaceWorlds' had to be set to false.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/common/config/Config.java
+++ b/src/main/java/slimeknights/tconstruct/common/config/Config.java
@@ -277,7 +277,7 @@ public final class Config {
       propOrder.add(prop.getName());
 
       prop = configFile.get(cat, "slimeIslandsOnlyGenerateInSurfaceWorlds", slimeIslandsOnlyGenerateInSurfaceWorlds);
-      prop.setComment("If true, slime islands wont generate in dimensions which aren't of type surface. This means they wont generate in modded cave dimensions like the deep dark.");
+      prop.setComment("If true, slime islands only generate in dimensions which are of type surface. This means they won't generate in modded cave dimensions like the deep dark.");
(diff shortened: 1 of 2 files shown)


KnightMiner:
This is a compatibility breaking change, and for a version we are basically done supporting. As a result, you will basically just break a bunch of modpacks with this. If you want to correct the comment to the actual behavior, that could possibly be merged, but we won't be inverting a config option in a new release for a version we are nearly done supporting.

jpark651:
I see. The default value could also be inverted, so a fresh install would behave the same way it did before. Would that address this concern, or would you still prefer to just have a comment that describes the config option behaving the opposite of its name?

KnightMiner:
The concern is packs with existing configs, so even if you change the default value, they already set the value. Easiest thing is have the comment clearly describe what it does now, and if needed mention the property name is inverted. I'm expecting most packs either never set this property (so are using the original default, not something we can change now), or set it and in testing realized its backwards (so they have it set to the one that works contrary to the description). If someone set it and did not realize it was backwards, well, I question that modpack's testing quality.

So yeah, just make the description accurate to what it does without changing game logic.

jpark651:
Makes sense, and sounds good. I've updated the branch to only change the description of the property.

jpark651:
Just to verify, does it look good now?

KnightMiner:
Yeah, it looks fine. I can merge this I suppose, though I cannot guarantee another 1.12 build as 1.16 is a much bigger priority (not even sure my 1.12 workspace still works). Its a case of if I wanted another 1.12 build, there are a few other bug fixes I'd want in there, but most of them are minor bugs.

PR #4253, opened 2021-01-24
Added 1.16 full Chinese translation

VeroFess:
A complete translation was carried out according to the English language document of 1.16, and it did not completely follow the partial uncontextual translation of the old version.


KnightMiner:
I'm a little concerned about accepting a translation at this time as content is going to change quite quickly, though if you are willing to update the translation later down the line when an alpha is ready I can look into merging it.

I should point out that I am working on one of the major content updates over on the `1.16-materials-update` branch. I would suggest at basing your translation on [the lang file there(https://github.com/SlimeKnights/TinkersConstruct/blob/1.16-materials-update/src/main/resources/assets/tconstruct/lang/en_us.json). Its mostly similar to the 1.16 branch, a few entries got language keys updated and a lot of new items got added. It was pretty late when I pushed that commit though, so I may have missed a few material names in the lang file

VeroFess:
Basically, I completely translated the new language files based on the 1.16 version. On the other hand, the TinkersConstruct module is already running on my 1.16.5 server (yes, this is crazy), so I will Update translating file in TinkersConstruct change it as soon as possible.

VeroFess:
请不要合并本次请求，我将基于1.16-materials-update重新进行翻译

VeroFess:
Closed, now working on 1.16-materials-update。 New PR here -> https://github.com/SlimeKnights/TinkersConstruct/pull/4254

PR #4262, opened 2021-02-15
1.16 modifiers update

hamza123654-AL:
make this mod compatible with other mod like allthemodiums and gobber


KnightMiner:
First off, a pull request is not where you make a feature request. A pull request is where you contribute code. The reason there was no suggestion template for 1.16 is we are not taking suggestions at this time

Second off, we don't add support for materials unique to one mod, on them to add it.

PR #4266, opened 2021-02-27
1.16 modifiers implementation

KnightMiner:
Modifier branch merging time! Its a ton of commits at this point so its a bit overdue, but this contains support for traits, upgrades, and abilities. Only missing incremental modifiers which we will add to the main branch later.


alexbegt (approved):
Looks great and runs great!

PR #4273, opened 2021-03-13
Slimesling

Stick-2357:


Stick-2357:
Not entirely certain why when I merged 1.16, it added all those changes to my PR. Sorry, I'm not the most adapt with Git.

src/main/java/slimeknights/tconstruct/gadgets/TinkerGadgets.java
```diff
@@ -93,8 +93,8 @@
   public static final ItemObject<PiggyBackPackItem> piggyBackpack = ITEMS.register("piggy_backpack", PiggyBackPackItem::new);
   public static final EnumObject<FrameType,FancyItemFrameItem> itemFrame = ITEMS.registerEnum(FrameType.values(), "item_frame", (type) -> new FancyItemFrameItem(((world, pos, dir) -> new FancyItemFrameEntity(world, pos, dir, type.getId()))));
   // slime tools
-  public static final EnumObject<SlimeType,SlimeSlingItem> slimeSling = ITEMS.registerEnum(SlimeType.values(), "slime_sling", (type) -> new SlimeSlingItem(UNSTACKABLE_PROPS));
```
Stick-2357, line 96: Important Changes

src/main/java/slimeknights/tconstruct/gadgets/item/SlimeSlingItem.java
```diff
@@ -1,5 +1,6 @@
 package slimeknights.tconstruct.gadgets.item;
```
Stick-2357: Important Changes

KnightMiner:
Try rebasing instead of merging. Might fix the unrelated commits.

Stick-2357:
There we go. Fixed those issues

src/main/java/slimeknights/tconstruct/gadgets/item/SlimeSlingItem.java
```diff
@@ -59,22 +68,92 @@ public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity en
...
-        vec.z * -f);
-
-      if (player instanceof ServerPlayerEntity) {
-        ServerPlayerEntity playerMP = (ServerPlayerEntity) player;
-        TinkerNetwork.getInstance().sendTo(new EntityMovementChangePacket(player), playerMP);
-      }
-
-      player.playSound(Sounds.SLIME_SLING.getSound(), 1f, 1f);
-      SlimeBounceHandler.addBounceHandler(player);
+    switch (this.slimeType) {
```
KnightMiner: This giant switch statement seems pretty messy. At this point I think I'd rather you just make separate item classes for each that extend a base as there is relatively little duplicated code in this method.

For the record, `EnumObject` has a builder syntax you can use in place of the `registerEnum` loop.

src/main/java/slimeknights/tconstruct/gadgets/item/SlimeSlingItem.java
```diff
@@ -59,22 +68,92 @@ public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity en
...
+          player.playSound(Sounds.SLIME_SLING.getSound(), 1f, 1f);
+        }
+        break;
+      case PURPLE:
+        look = player.getLookVec();
+        double offX = look.x * f;
+        double offY = look.y * f;
+        double offZ = look.z * f;
+
+        player.setPosition(player.getPosX() + offX, player.getPosY() + offY, player.getPosZ() + offZ);
```
KnightMiner: Needs a teleport sound effect. Might be worth ensuring the position is valid as well (not inside a block), not sure the best way to handle that.

src/main/java/slimeknights/tconstruct/gadgets/item/SlimeSlingItem.java
```diff
@@ -59,22 +68,92 @@ public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity en
...
+          if (entity.getBoundingBox().intersects(Math.min(start.x, direction.x), Math.min(start.y, direction.y), Math.min(start.z, direction.z), Math.max(start.x, direction.x), Math.max(start.y, direction.y), Math.max(start.z, direction.z))) {
+            if (look.distanceTo(entity.getLookVec()) < dist) {
+              dist = look.distanceTo(entity.getLookVec());
+              closestEntity = entity;
+            }
+          }
+        }
+        // TODO: Ensure there isn't a block in the way
+        if (closestEntity != null) {
+          closestEntity.addVelocity(look.x * f,
```
KnightMiner: Pretty sure there is an add knockback method, probably better to use that instead. Means knockback resistance (iron golems) properly works

src/main/java/slimeknights/tconstruct/gadgets/item/SlimeSlingItem.java
```diff
@@ -44,8 +53,8 @@ public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity en
     }
 
     PlayerEntity player = (PlayerEntity) entityLiving;
-    // has to be on ground to do something
-    if (!player.isOnGround()) {
+    // has to be on ground to do something (except Blue)
+    if ((!player.isOnGround() && this.slimeType != StickySlimeBlock.SlimeType.BLUE) || player.isElytraFlying()) {
```
KnightMiner: Honestly, only green needs to be on the ground. Not even certain green being on ground is a requirement. Consider this check for every type separately.

Notably, I see absolutely no problems with using orange in the air. Purple is a similar usage to blue, so in the air is good there too. Green would be cool to use when you are sliding against a honey block.

src/main/java/slimeknights/tconstruct/gadgets/TinkerGadgets.java
```diff
@@ -93,8 +93,8 @@
...
-  public static final EnumObject<SlimeType,SlimeBootsItem> slimeBoots = ITEMS.registerEnum(SlimeType.values(), "slime_boots", (type) -> new SlimeBootsItem(type, UNSTACKABLE_PROPS));
+  public static final EnumObject<SlimeType,SlimeSlingItem> slimeSling = ITEMS.registerEnum(SlimeType.TRUE_SLIME, "slime_sling", (type) -> new SlimeSlingItem(type, UNSTACKABLE_PROPS));
+  public static final EnumObject<SlimeType,SlimeBootsItem> slimeBoots = ITEMS.registerEnum(SlimeType.TRUE_SLIME, "slime_boots", (type) -> new SlimeBootsItem(type, UNSTACKABLE_PROPS));
```
KnightMiner: If you are going to remove items, you have to add migrations in `TConstruct`. There are some existing migrations there already.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/BaseSlimeSlingItem.java
```diff
@@ -0,0 +1,64 @@
...
+
+  /**
+   * returns the action that specifies what animation to play when the items is being used
+   */
+  @Override
+  public UseAction getUseAction(ItemStack stack) {
+    return UseAction.BOW;
+  }
+
+  public float getForce(int i) {
```
KnightMiner: What does i mean? Not a great parameter name

Honestly, I would just make this stack sensitive, add a `timeLeft` parameter, run `this.getUseDuration(stack) - timeLeft` inside. Also make sure to add javadocs for new stuff, I just don't care about javadocs for methods marked `@Override`

Stick-2357:
Here's what I've got for now. There are still some major additions needed, which I have todos marked for.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/PurpleSlimeSlingItem.java
```diff
@@ -39,6 +42,13 @@ public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity en
       TinkerNetwork.getInstance().sendTo(new EntityMovementChangePacket(player), playerMP);
     }
 
+    // particle effect from EnderPearlEntity
+    Random rand = new Random();
```
KnightMiner: You don't need a new instance. World has one, pretty sure item does as well.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/BlueSlimeSlingItem.java
```diff
@@ -0,0 +1,52 @@
...
+
+  /** Called when the player stops using an Item (stops holding the right mouse button). */
+  @Override
+  public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity entityLiving, int timeLeft) {
+    if (!(entityLiving instanceof PlayerEntity)) {
+      return;
+    }
+
+    PlayerEntity player = (PlayerEntity) entityLiving;
+    if (player.isElytraFlying()) {
```
KnightMiner: Add a comment here explaining why we disallow elytra flying. Loosely, the problem is slimeslings behave similarly to fireworks, so this is essentially a cheap version of fireworks.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/BlueSlimeSlingItem.java
```diff
@@ -0,0 +1,52 @@
...
+      speed,
+      (MathHelper.cos(player.rotationYaw / 180.0F * (float) Math.PI) * MathHelper.cos(player.rotationPitch / 180.0F * (float) Math.PI) * speed));
+
+    if (player instanceof ServerPlayerEntity) {
+      ServerPlayerEntity playerMP = (ServerPlayerEntity) player;
+      TinkerNetwork.getInstance().sendTo(new EntityMovementChangePacket(player), playerMP);
+    }
+
+    player.playSound(Sounds.SLIME_SLING.getSound(), 1f, 1f);
+    SlimeBounceHandler.addBounceHandler(player);
```
KnightMiner: I'm not certain this part is needed, but its been awhile. As I recall the method is for anything that expects the player to be bouncing. If it works though, I suppose we can leave it in and look into it later

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/MagmaSlimeSlingItem.java
```diff
@@ -0,0 +1,63 @@
...
+          dist = look.distanceTo(entity.getLookVec());
+          closestEntity = entity;
+        }
+      }
+    }
+    if (closestEntity != null) {
+      if (closestEntity instanceof LivingEntity) {
+        ((LivingEntity) closestEntity).applyKnockback(f , -look.x, -look.z);
+
+        // TODO: Test on other players (I wasn't able to figure out setting that up)
```
KnightMiner: Yeah, its a bit of a pain to set up (though at least its better than testing shield blocking). I'll test it when I get a chance to test everything in game

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/MagmaSlimeSlingItem.java
```diff
@@ -0,0 +1,63 @@
...
+    float range = 5F;
+    Vector3d start = player.getEyePosition(1F);
+    Vector3d look = player.getLookVec();
+    Vector3d direction = start.add(look.x * range, look.y * range, look.z * range);
+    AxisAlignedBB bb = player.getBoundingBox().expand(look.x * range, look.y * range, look.z * range).expand(1, 1, 1);
+    List<Entity> entitiesInArea = player.getEntityWorld().getEntitiesWithinAABBExcludingEntity(player, bb);
+    double dist = range;
+    Entity closestEntity = null;
+    for (Entity entity : entitiesInArea) {
+      if (entity.getBoundingBox().intersects(Math.min(start.x, direction.x), Math.min(start.y, direction.y), Math.min(start.z, direction.z), Math.max(start.x, direction.x), Math.max(start.y, direction.y), Math.max(start.z, direction.z))) {
```
KnightMiner: These min and max values are not changing with each iteration. It may be a micro-optimization, but we might as well calculate them just once before the for loop. Can give them names like `minX`, `maxY`, etc.

src/main/resources/assets/tconstruct/lang/en_us.json
```diff
@@ -628,16 +628,14 @@
...
-    "item.tconstruct.blood_slime_sling": "Congealed Slimesling",
-    "item.tconstruct.blood_slime_sling.tooltip": "Charge up, aim low, get flinging!\nUse Slime Boots if you value your life!",
-    "item.tconstruct.magma_slime_sling": "Slimesling",
-    "item.tconstruct.magma_slime_sling.tooltip": "Charge up, aim low, get flinging!\nUse Slime Boots if you value your life!",
+    "item.tconstruct.blue_slime_sling": "Sky Slimesling",
+    "item.tconstruct.blue_slime_sling.tooltip": "Charge up, let go, get back in the air!\nUse Slime Boots if you value your life!",
+    "item.tconstruct.purple_slime_sling": "Ender Slimesling",
+    "item.tconstruct.purple_slime_sling.tooltip": "Charge up, point it at a wall, come out the other side!",
+    "item.tconstruct.magma_slime_sling": "Magma Slimesling",
+    "item.tconstruct.magma_slime_sling.tooltip": "Charge up, point at a creature, get it out of my face!",
```
KnightMiner: This sentence reads weirdly, the switch from second person of the rest of the tooltips to first person here is awkward. Something like the following would be more consistent:
> Charge up, point at a creature, get it out of your face!

Though we could also phrase it a bit less aggressively:
> Charge up, point at a creature, and send it flying!

KnightMiner (changes requested):
Alright, I think that is everything I can comment on until I get a chance to test it in game. With how my week is going, probably won't be until this weekend.

Most of the remaining comments are relatively minor

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/MagmaSlimeSlingItem.java
```diff
@@ -0,0 +1,79 @@
...
+    Vector3d look = player.getLookVec();
+    Vector3d direction = start.add(look.x * range, look.y * range, look.z * range);
+    AxisAlignedBB bb = player.getBoundingBox().expand(look.x * range, look.y * range, look.z * range).expand(1, 1, 1);
+    double minX = Math.min(start.x, direction.x);
+    double maxX = Math.max(start.x, direction.x);
+    double minY = Math.min(start.y, direction.y);
+    double maxY = Math.max(start.y, direction.y);
+    double minZ = Math.min(start.z, direction.z);
+    double maxZ = Math.max(start.z, direction.z);
+    List<Entity> entitiesInArea = player.getEntityWorld().getEntitiesWithinAABBExcludingEntity(player, bb);
```
KnightMiner: This feels super awkward, when looking directly at an entity half the time it does not trigger. Also, as far as the logic, have you looked into `ProjectileHelper.rayTraceEntities`? There is a version of the method that works on servers, used by arrows.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/MagmaSlimeSlingItem.java
```diff
@@ -0,0 +1,79 @@
...
+      }
+    }
+
+    // cancel if there's a block in the way
+    BlockRayTraceResult mop = rayTrace(worldIn, player, RayTraceContext.FluidMode.NONE);
+    if (mop.getType() == RayTraceResult.Type.BLOCK && dist * dist < mop.getPos().distanceSq(player.getPosX(), player.getPosY(), player.getPosZ(), true)) {
+      return;
+    }
+
+    if (closestEntity != null) {
```
KnightMiner: Need a sound for when it misses. Maybe we play a slightly different squish sound? Something to show it was released but missed

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/BlueSlimeSlingItem.java
```diff
@@ -0,0 +1,54 @@
...
+      return;
+    }
+
+    float f = getForce(stack, timeLeft);
+
+    player.addExhaustion(0.2F);
+    player.getCooldownTracker().setCooldown(stack.getItem(), 3);
+    player.setSprinting(true);
+
+    float speed = f / 3F;
```
KnightMiner: This feels too strong. Green slime sling should be the strongest, with blue weaker but working without kicking off a block.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/MagmaSlimeSlingItem.java
```diff
@@ -0,0 +1,79 @@
...
+      if (closestEntity instanceof LivingEntity) {
+        ((LivingEntity) closestEntity).applyKnockback(f , -look.x, -look.z);
```
KnightMiner: I am wondering if this is a bit too strong. Currently at max value this goes up to 6.0. For comparison, a level of knockback is 0.5, so effectively this is knockback 12. Maybe we should half it so its just knockback 6?

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/PurpleSlimeSlingItem.java
```diff
@@ -0,0 +1,51 @@
...
+      return;
+    }
+
+    PlayerEntity player = (PlayerEntity) entityLiving;
+    float f = getForce(stack, timeLeft);
+
+    Vector3d look = player.getLookVec();
+    double offX = look.x * f;
+    double offY = look.y * f + 1;
+    double offZ = look.z * f;
```
KnightMiner: This is a bit hard to use precisely, lets half the power when you are sneaking so if you want to teleport to the other side of a wall you can, otherwise if you want to go faster you just don't crouch. This might be fixing the wrong problem though, address my later comment in this class first and see how it feels.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/PurpleSlimeSlingItem.java
```diff
@@ -0,0 +1,51 @@
...
+
+    if (player.attemptTeleport(player.getPosX() + offX, player.getPosY() + offY, player.getPosZ() + offZ, true)) {
```
KnightMiner: This feels awkward when it cannot teleport, as the player does not move. If I am staring at a block, and teleporting would put me inside it, I'd expect to teleport right next to it instead of staying still.

Not sure of the best way to do this, I suppose we could start from the block position and iterate towards the player backwards along the look vector. Or maybe we could approximate the behavior by just selecting a second position to try via raytrace if the first does not move the player.

Also, this behaves pretty weirdly in water, we might want to write our own attempt teleport method that lets us ignore liquids.

KnightMiner (changes requested):
Just tested in game, good start overall but have a few comments for improvements.

I did also notice you did not add a package-info for the new package. That lets IDEA properly do nullability checks, so make sure you add one (you can copy the one from the items package)

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/MagmaSlimeSlingItem.java
```diff
@@ -31,49 +34,42 @@ public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity en
...
       }
+
+      player.playSound(Sounds.SLIME_SLING.getSound(), 1f, 1f);
+    } else {
+      playMissSound(player);
     }
   }
+
+  // Using function for consistency
+  public void playMissSound(PlayerEntity player) {
```
KnightMiner: Since its just for consistency, can probably be private. I doubt anyone will extend this, but I'd rather make too much private than too much public.

KnightMiner:
I had to push the commit that renames slime types to their proper names a bit earlier than I planned, so make sure to rebase this to the new branch. Shouldn't impact your code outside that EnumObject you made, though Git dislike the lang file being changed by both

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/EnderSlimeSlingItem.java
```diff
@@ -29,13 +34,26 @@ public void onPlayerStoppedUsing(ItemStack stack, World worldIn, LivingEntity en
...
+    while (Math.abs(offX) > .5 || Math.abs(offY) > .5 || Math.abs(offZ) > .5) { // while not too close to player
+      posToCheck.add(posAttempt);
+      offX -= (Math.abs(offX) > .25 ? (offX >= 0 ? 1 : -1) * .25 : 0);
+      offY -= (Math.abs(offY) > .25 ? (offY >= 0 ? 1 : -1) * .25 : 0);
+      offZ -= (Math.abs(offZ) > .25 ? (offZ >= 0 ? 1 : -1) * .25 : 0);
+      posAttempt = new BlockPos(player.getPosX() + offX, player.getPosY() + offY, player.getPosZ() + offZ);
+    }
+
+    // filter to only teleportable blocks
+    posToCheck = posToCheck.stream().distinct().filter(block -> !worldIn.getBlockState(block).isSuffocating(worldIn, block)).collect(Collectors.toList());
```
KnightMiner: Switch to findFirst instead of collect. No need to build a whole list when we only want the first element.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/EarthSlimeSlingItem.java
```diff
@@ -0,0 +1,44 @@
...
+    if (!(entityLiving instanceof PlayerEntity)) {
+      return;
+    }
+
+    PlayerEntity player = (PlayerEntity) entityLiving;
+    float f = getForce(stack, timeLeft);
+
+    // check if player was targeting a block
+    BlockRayTraceResult mop = rayTrace(worldIn, player, RayTraceContext.FluidMode.NONE);
+    if (mop.getType() == RayTraceResult.Type.BLOCK) {
```
KnightMiner, line 31: Needs the miss sound when its not targeting a block.

Might be worth adding a method to `BaseSlimeSlingItem`, something like `playSlingSound(PlayerEntity entity, boolean isSuccess)`

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/EnderSlimeSlingItem.java
```diff
@@ -0,0 +1,65 @@
...
+      BlockPos furthestPos = furthestPosOp.get();
+      player.setPosition(furthestPos.getX(), furthestPos.getY(), furthestPos.getZ());
+
+      // particle effect from EnderPearlEntity
+      for (int i = 0; i < 32; ++i) {
+        worldIn.addParticle(ParticleTypes.PORTAL, player.getPosX(), player.getPosY() + worldIn.rand.nextDouble() * 2.0D, player.getPosZ(), worldIn.rand.nextGaussian(), 0.0D, worldIn.rand.nextGaussian());
+      }
+      playerServerMovement(player);
+      player.playSound(Sounds.SLIME_SLING.getSound(), 1f, 1f);
+      player.playSound(SoundEvents.ENTITY_ENDERMAN_TELEPORT, 1f, 1f);
```
KnightMiner, line 60: Probably want a cooldown similarly to the blue slime sling, I assume a click macro could use this to fly. Not sure orange or green needs a cooldown. Probably a good idea for safety for both, and consistency

KnightMiner (changes requested):
Ender slime sling feels a lot better now. I think these are the last two changes, both small. After these should be good to merge

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/EarthSlimeSlingItem.java
```diff
@@ -0,0 +1,45 @@
...
+      Vector3d vec = player.getLookVec().normalize();
+      player.addVelocity(vec.x * -f,
+        vec.y * -f / 3f,
+        vec.z * -f);
+
+      playerServerMovement(player);
+      SlimeBounceHandler.addBounceHandler(player);
+    }
+    playSlingSound(player, mop.getType() == RayTraceResult.Type.BLOCK);
+
```
alexbegt: Any particular reason for this blank empty space?

alexbegt (approved):
Looks good, just wondering why there's an empty space on line 43 in EarthSlimeStringItem.

src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/BaseSlimeSlingItem.java
```diff
@@ -68,9 +68,19 @@ public void playerServerMovement(LivingEntity player) {
     }
   }
 
-  // Using function for consistency
+  protected void playSlingSound(PlayerEntity player, boolean isSuccess) {
+    if (isSuccess) {
+      playSuccessSound(player);
```
KnightMiner: Considering the only change is the pitch, this might be a bit too much abstraction. If you are just going to have two separate methods, just have two separate methods and ditch the one with the boolean parameter.

My intention was not something to override (so leave the teleport sound out), and more something to simplify your logic.

KnightMiner (approved):

PR #4278, opened 2021-03-30
update tool and toolpart base textures to new style

RCXcrafter:
- base textures
- stone base textures
- bone base textures
- wood base textures
- broken textures
Currently unimplemented tools also have textures added.
Images:
image
image
image
image
image
image
image

Note that this pr breaks the colors of the mattock button in the tinker station:
image


PrincessOfEvil:
mDiyo wanted more shovel-like mattocks. Was that resolved differently?

RCXcrafter:
I did edit the mattock to try to make it more shovel like but I forgot to ask mDiyo for feedback.

@mDiyo What do you think of the mattock texture? Is it shovely enough?

PR #4348, opened 2021-04-12
Kama tripwire

Stick-2357:


KnightMiner:
Alright, I tested this out, and it works perfect when you are targeting the tripwire block. However, when you are targeting a block next to the tripwire and AOE breaks the tripwire, it seems we trigger some sort of vanilla race condition, I believe [MC-129055](https://bugs.mojang.com/browse/MC-129055). I can produce that race condition inconsistently with vanilla shears, especially if things are paused in the debugger, but it seems to happen 100% of the time with the AOE block break.

Not sure what the best approach is here, I see a potential fix in the vanilla logic where I believe this bug is caused, so I may look into a Forge PR for it (MinecraftForge/MinecraftForge#7718). For now going to leave this PR open.

Stick-2357:
I am still having the issue, but I was having the issue when using shears as well, so there's a chance that my Forge version just wasn't updating right.

KnightMiner:
Need to be on at least `1.16.5-36.1.15`, I did not update the root project, but I'll make sure to do so when I test this again.

PR #4352, opened 2021-04-13
Add CraftTweaker support

jaredlll08:
As discussed with @KnightMiner on discord.


src/main/java/slimeknights/tconstruct/plugin/crt/CRTHelper.java
```diff
@@ -0,0 +1,150 @@
...
+import slimeknights.tconstruct.library.modifiers.ModifierEntry;
+import slimeknights.tconstruct.library.modifiers.ModifierId;
+import slimeknights.tconstruct.library.tinkering.IMaterialItem;
+import slimeknights.tconstruct.library.tools.item.ToolCore;
+
+import java.util.stream.Collectors;
+
+public class CRTHelper {
+  
+  public static MaterialId getMaterialId(String materialId) {
```
KnightMiner, line 48: In general, I like javadocs on public methods without `@Override` annotations. You can skip it on event listeners, but on utilities just a simple javadoc explaining the method for consistency. On this method specifically, worth mentioning it validates the material ID.

src/main/java/slimeknights/tconstruct/plugin/crt/CRTHelper.java
```diff
@@ -0,0 +1,150 @@
...
+    event.addResource(location("material"));
+    event.addResource(location("melting"));
+    event.addResource(location("molding_basin"));
+    event.addResource(location("molding_table"));
+    event.addResource(location("part_builder"));
+    event.addResource(location("tinker_station"));
+  }
+  
+  private ResourceLocation location(String location) {
+    return new ResourceLocation(TConstruct.modID, TConstruct.modID + "/" + location);
```
KnightMiner: I assume this is an API detail, but any reason the mod ID is here twice?

src/main/java/slimeknights/tconstruct/plugin/crt/CRTHelper.java
```diff
@@ -0,0 +1,150 @@
...
+      final StringTextComponent message = new StringTextComponent(TextFormatting.GREEN + "Material Items written to the log" + TextFormatting.RESET);
+      commandContext.getSource().sendFeedback(message, true);
+      return 0;
+    });
+    
+    
+    event.registerDump("ticModifiers", "Lists the different Tinkers Construct Modifiers", commandContext -> {
+      
+      CraftTweakerAPI.logDump("List of all Tinkers Construct Modifiers: ");
+      TinkerRegistries.MODIFIERS.getValues().forEach(iMaterial -> {
```
KnightMiner: Minor, but `iMaterial` should probably be `modifier` to make this more readible

src/main/java/slimeknights/tconstruct/plugin/crt/managers/AlloyingManager.java
```diff
@@ -0,0 +1,59 @@
...
+import java.util.Arrays;
+import java.util.List;
+import java.util.stream.Collectors;
+
+@ZenRegister
+@ZenCodeType.Name("mods.tconstruct.Allying")
+public class AlloyingManager implements IRecipeManager {
+  
+  @ZenCodeType.Method
+  public void addRecipe(String name, IFluidStack[] ingredients, IFluidStack output, int temperature) {
```
KnightMiner: Do you have a fluid ingredient, or just fluid stacks? It should be possible to map from that to the fluid stack, or to extend our `FluidIngredient` for that.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/BeheadingManager.java
```diff
@@ -0,0 +1,49 @@
...
+  public void addRecipe(String name, MCEntityType ingredient, IItemStack output) {
+    name = fixRecipeName(name);
+    ResourceLocation id = new ResourceLocation("crafttweaker", name);
+    BeheadingRecipe recipe = new BeheadingRecipe(id, EntityIngredient.of(ingredient.getInternal()), ItemOutput.fromStack(output.getInternal()));
+    CraftTweakerAPI.apply(new ActionAddRecipe(this, recipe));
+  }
+  
+  @Override
+  public void removeRecipe(IItemStack output) {
+    throw new IllegalArgumentException("Cannot remove Beheading Recipes by an IItemStack output! Use `removeRecipe(MCEntityType input)` instead!");
```
KnightMiner: Why can't you remove recipes by an item stack output? They have an item stack output.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/CastingBasinManager.java
```diff
@@ -0,0 +1,51 @@
...
+import slimeknights.tconstruct.library.recipe.casting.material.MaterialCastingRecipe;
+import slimeknights.tconstruct.library.tinkering.IMaterialItem;
+import slimeknights.tconstruct.plugin.crt.managers.base.ICastingManager;
+
+@ZenRegister
+@ZenCodeType.Name("mods.tconstruct.CastingBasin")
+public class CastingBasinManager implements ICastingManager {
+  
+  @Override
+  public ItemCastingRecipe makeItemCastingRecipe(ResourceLocation id, String group, Ingredient cast, FluidIngredient fluid, ItemOutput result, int coolingTime, boolean consumed, boolean switchSlots) {
```
KnightMiner: Probably want to name the consume parameter something like `consumeCast`, probably a non-great name in our API

src/main/java/slimeknights/tconstruct/plugin/crt/managers/CastingBasinManager.java
```diff
@@ -0,0 +1,51 @@
...
+    return new ItemCastingRecipe.Basin(id, "", cast, fluid, result, coolingTime, consumed, switchSlots);
```
KnightMiner: You have a group parameter, but then pass in an empty string for group here. Few other places in this file

src/main/java/slimeknights/tconstruct/plugin/crt/managers/FuelManager.java
```diff
@@ -0,0 +1,52 @@
...
+    name = fixRecipeName(name);
+    ResourceLocation id = new ResourceLocation("crafttweaker", name);
+    FluidIngredient fluidIngredient = FluidIngredient.of(input.getInternal());
+    MeltingFuel recipe = new MeltingFuel(id, "", fluidIngredient, duration, temperature);
+    CraftTweakerAPI.apply(new ActionAddRecipe(this, recipe));
+  }
+  
+  @Override
+  public void removeRecipe(IItemStack output) {
+    throw new IllegalArgumentException("Cannot remove Alloy Recipes by an IItemStack output as it doesn't output anything! Use `removeRecipe(Fluid input)` instead!");
```
KnightMiner: Fuel recipes

src/main/java/slimeknights/tconstruct/plugin/crt/managers/MaterialManager.java
```diff
@@ -0,0 +1,40 @@
...
+import net.minecraft.util.ResourceLocation;
+import org.openzen.zencode.java.ZenCodeType;
+import slimeknights.tconstruct.library.materials.MaterialId;
+import slimeknights.tconstruct.library.recipe.RecipeTypes;
+import slimeknights.tconstruct.library.recipe.material.MaterialRecipe;
+import slimeknights.tconstruct.plugin.crt.CRTHelper;
+
+@ZenRegister
+@ZenCodeType.Name("mods.tconstruct.Material")
+public class MaterialManager implements IRecipeManager {
```
KnightMiner: I would name this `MaterialRecipeManager`/`mods.tconstruct.MaterialRecipe` to make it more clear. `Material` implies the tool material JSONs as CT is not just recipes, this recipe adds the mapping from items to materials.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/MaterialManager.java
```diff
@@ -0,0 +1,40 @@
...
+  
+  @ZenCodeType.Method
+  public void addMaterial(String name, IIngredient ingredient, int value, int needed, String materialId) {
```
KnightMiner: With the above rename, you can call this `addItem`. I am looking into making a version of `MaterialRecipe` for fluids.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/PartBuilderManager.java
```diff
@@ -0,0 +1,43 @@
...
+import slimeknights.tconstruct.library.recipe.RecipeTypes;
+import slimeknights.tconstruct.library.recipe.partbuilder.PartRecipe;
+import slimeknights.tconstruct.library.tinkering.IMaterialItem;
+
+@ZenRegister
+@ZenCodeType.Name("mods.tconstruct.Partbuilder")
+public class PartBuilderManager implements IRecipeManager {
+  
+  @ZenCodeType.Method
+  public void addRecipe(String name, String pattern, int cost, Item output, int outputCount) {
```
KnightMiner: Call this something like `addMaterialRecipe`, I want to add support to the part builder for non-material recipes, sorta like the stonecutter.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/TinkerStationManager.java
```diff
@@ -0,0 +1,99 @@
...
+  @ZenCodeType.Method
+  public void addOverslimeModifierRecipe(String name, IIngredient ingredient, int restoreAmount) {
+    name = fixRecipeName(name);
+    ResourceLocation id = new ResourceLocation("crafttweaker", name);
+    OverslimeModifierRecipe recipe = new OverslimeModifierRecipe(id, ingredient.asVanillaIngredient(), restoreAmount);
+    CraftTweakerAPI.apply(new ActionAddRecipe(this, recipe));
+  }
+  
+  @ZenCodeType.Method
+  public void addModifierRecipe(String name, IIngredientWithAmount[] inputs, IIngredient toolRequired, String modifierResult, int modifierResultLevel, int maxLevel, int upgradeSlots, int abilitySlots) {
```
KnightMiner: `upgradeSlots` and `abilitySlots` are mutually exclusive, setting both would cause problems with JEI and a few future features.

I would suggest possibly making 2 methods: one for upgrades and one for abilities. I suppose a third for slotless modifiers would make sense, but you can just call either of the other 2 with 0 as the parameter.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/PartBuilderManager.java
KnightMiner: Cost should be validated as `>= 1`, we only validate it as part of the JSON parser, not the constructor.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/TinkerStationManager.java
KnightMiner: Some of these numbers could use validation. `maxLevel`/`modifierResultLevel` are both 0 or more. `upgradeSlots` and `abilitySlots` are both 1 or more.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/TinkerStationManager.java
```diff
@@ -0,0 +1,99 @@
...
+    ModifierEntry result = new ModifierEntry(resultModifier, modifierResultLevel);
+    ModifierRecipe recipe = new ModifierRecipe(id, collect, toolRequired.asVanillaIngredient(), entry, "", result, maxLevel, upgradeSlots, abilitySlots);
+    CraftTweakerAPI.apply(new ActionAddRecipe(this, recipe));
+  }
+  
+  @ZenCodeType.Method
+  public void addModifierRecipe(String name, IIngredientWithAmount[] inputs, IIngredient toolRequired, String modifierResult, int modifierResultLevel, int maxLevel, int upgradeSlots, int abilitySlots, IData modifierRequirements, int minMatch, String requirementsError) {
+    name = fixRecipeName(name);
+    Modifier resultModifier = CRTHelper.getModifier(modifierResult);
+    ModifierMatch[] modifierMatches = modifierRequirements.asMap().entrySet().stream().map(stringIDataEntry -> ModifierMatch.entry(CRTHelper.getModifier(stringIDataEntry.getKey()), stringIDataEntry.getValue().asNumber().getInt())).toArray(ModifierMatch[]::new);
```
KnightMiner: ModifierMatch is a recursive data structure, so its a bit weird to see a non-recursive method parsing the data. If you don't want to support it thats fine, just making sure I understand how you implemented this.

src/main/resources/data/tconstruct/scripts/tconstruct/material.zs
```diff
@@ -0,0 +1,20 @@
...
+ * 
+ * The `needed` value is how many of the input (in this case Diamonds) are needed to make `1` Material.
+ * An example of this being Ingots vs Nuggets. Ingots have a `needed` value of `1` while Nuggets have a `needed` value of `9` (since it takes 9 Nuggets to make 1 Ingot).
+ * 
+ * You can find valid materialIds by running `/ct dump ticMaterials` in-game.
+ */
+
+// <recipetype:tconstruct:material>.addMaterial(name as string, ingredient as IIngredient, value as int, needed as int, materialId as string)
+
+<recipetype:tconstruct:material>.addMaterial("test_material", <item:minecraft:diamond>, 3, 5, "tconstruct:bone");
```
KnightMiner: Diamond might not be the best example as when used in repair it conflicts with the diamond modifier. I'd suggest something weird like cocoa beans.

src/main/resources/data/tconstruct/scripts/tconstruct/material.zs
```diff
@@ -0,0 +1,20 @@
...
+
+/*
+ * Removes the Flint Material.
```
KnightMiner: Echoing my comment on the Java code, this does not remove the flint material. This removes the recipe that allows repairing/making it in the part builder.

src/main/resources/data/tconstruct/scripts/tconstruct/part_builder.zs
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * The pattern is set as the Axe Head pattern, so in the Part Builder, the recipe will show under the Axe Head, but it will still make Tool Rods.
+ * Valid patterns are defined by the texture in the `tconstruct:gui/tinker_pattern/` folder in the resources.
+ *
+ * The output Item needs to be an Item that works with Materials, you can find a list of valid items by running `/ct dump ticMaterialItems`.
+ */
+
+// <recipetype:tconstruct:part_builder>.addRecipe(name as string, pattern as String, cost as int, output as Item, outputCount as int)
+
+<recipetype:tconstruct:part_builder>.addRecipe("part_builder_test", "tconstruct:axe_head", 2, <item:tconstruct:tool_rod>, 3);
```
KnightMiner: Recent commit renamed `tconstruct:tool_rod` to `tconstruct:tool_handle`. If you don't want to rebase, `tconstruct:pickaxe_head` is an option

src/main/resources/data/tconstruct/scripts/tconstruct/tinker_station.zs
```diff
@@ -0,0 +1,37 @@
...
+// <recipetype:tconstruct:tinker_station>.addIncrementalModifierRecipe(String name, IIngredient input, int amountPerInput, int neededPerLevel, IIngredient toolRequirement, String modifierResult, int modifierResultLevel, int maxLevel, int upgradeSlots, int abilitySlots, IItemStack leftover)
+// <recipetype:tconstruct:tinker_station>.addIncrementalModifierRecipe(String name, IIngredient input, int amountPerInput, int neededPerLevel, IIngredient toolRequirement, String modifierResult, int modifierResultLevel, int maxLevel, int upgradeSlots, int abilitySlots, IItemStack leftover, IData modifierRequirements, int minMatch, String requirementsError)
+
+<recipetype:tconstruct:tinker_station>.addOverslimeModifierRecipe("overslime_test", <item:minecraft:dirt>, 800);
+<recipetype:tconstruct:tinker_station>.addModifierRecipe("modifier_test", [<item:minecraft:stick> * 3], <item:tconstruct:pickaxe>, "tconstruct:diamond", 2, 5, 1, 1);
+<recipetype:tconstruct:tinker_station>.addModifierRecipe("modifier_test_requirements", [<item:minecraft:apple> * 3], <tag:items:tconstruct:modifiable>, "tconstruct:emerald", 2, 5, 1, 1, {"tconstruct:haste": 1}, 1, "Haste needs to be applied first!");
```
KnightMiner: The requirements error is really a translation key, not sure we should encourage passing in a raw string even if it technically works.

KnightMiner (changes requested):
Overall seems you got the idea of the recipes. Left a few comments on the trickier ones, and a couple of style comments.

src/main/java/slimeknights/tconstruct/plugin/crt/CRTHelper.java
jaredlll08: So the first mod ID is so that it can find the files in the first place, the second is the folder that the files go in when someone runs `/ct examples`. So with how it is setup now, running `/ct examples` produces the following:
image

If we removed the second modid, it would be with the rest of the CraftTweaker examples.

I can change it to be something else if you would like? maybe "tinkers_construct"?

src/main/java/slimeknights/tconstruct/plugin/crt/managers/AlloyingManager.java
jaredlll08: We only have FluidStacks and MCtag<Fluid>.

At the end of the day, all mods are implementing their own FluidIngredient differently, so we decided that just getting a stack and passing it off to mods to turn into their own FluidIngredient would probably be the best way to do it.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/BeheadingManager.java
jaredlll08: The only output I see that I can actually use (without having to pass an entity) in is here
https://github.com/jaredlll08/TinkersConstruct/blob/66512ba258bf6b8986ffb05ec8cbcec677ab14e7/src/main/java/slimeknights/tconstruct/library/recipe/modifiers/BeheadingRecipe.java#L54

Based on that javadoc saying it is only used for JEI, I figured I should probably use a more reliable approach to removals.

I'm happy to use it if you think it is fine though

src/main/java/slimeknights/tconstruct/plugin/crt/CRTHelper.java
KnightMiner: Nah, was just curious as it looked unintentional. Maybe add a comment explaining the two mod IDs?

src/main/java/slimeknights/tconstruct/plugin/crt/managers/AlloyingManager.java
KnightMiner: Thats probably fine, if that changes let me know and I can make sure mantle's ingredient supports it.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/BeheadingManager.java
KnightMiner: Javadoc is because we have two recipes that extend this one, `PlayerBeheadingRecipe` and `SnowGolemBeheadingRecipe`. Both depend on the entity instance to set the proper NBT and alike on the output. However, if one was to remove by `Items.PLAYER_HEAD`, that would properly remove the player beheading recipe, so I think its fine.

Only downside is if you try removing the snow golem beheading recipe by `Blocks.SNOW_BLOCK`, however since that only drops when it lacks a pumpkin and none of the recipe viewers show it, I doubt anyone would do that.

src/main/java/slimeknights/tconstruct/plugin/crt/CRTHelper.java
jaredlll08: The event is now deprecated, so no need for the method anymore.

src/main/resources/data/tconstruct/scripts/tconstruct/tinker_station.zs
```diff
@@ -1,32 +1,78 @@
...
+<recipetype:tconstruct:tinker_station>.addAbilityModifierRecipe("ability_modifier_example", [<item:minecraft:gravel>], <tag:items:tconstruct:modifiable>, "tconstruct:smite", 1, 3, 1);
+<recipetype:tconstruct:tinker_station>.addAbilityModifierRecipe("ability_modifier_requirements_example", [<item:minecraft:orange_wool>], <tag:items:tconstruct:modifiable>, "tconstruct:sharpness", 1, 3, 1, {"tconstruct:emerald": 1}, 1, "recipe.tconstruct.modifier.requirements_error");
+<recipetype:tconstruct:tinker_station>.addIncrementalModifierRecipe("incremental_modifier_example", <item:minecraft:black_wool>, 2, 8, <tag:items:tconstruct:modifiable>, "tconstruct:haste", 1, 8, <item:minecraft:black_dye>);
+<recipetype:tconstruct:tinker_station>.addIncrementalModifierRecipe("incremental_modifier_requirements_example", <item:minecraft:yellow_wool>, 3, 8, <tag:items:tconstruct:modifiable>, "tconstruct:searing", 1, 8, <item:minecraft:yellow_dye>, {"tconstruct:emerald": 1}, 1, "recipe.tconstruct.modifier.requirements_error");
```
KnightMiner: Incremental recipes only really work on certain modifiers. Requires some additional code added in the modifier logic, so most don't support it at this time. Probably better for examples to use existing incremental modifiers.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/TinkerStationManager.java
jaredlll08: I have added some checks, but I can't do the check for `upgradeSlots` and `abilitySlots` as those *can* both be 0.

src/main/resources/data/tconstruct/scripts/tconstruct/casting_basin.zs
```diff
@@ -0,0 +1,35 @@
...
+ * 2) Adds a Material Casting Recipe that will allow a Diamond to be used as a Cast to make any Axe Head (Part), regardless of material (The Diamond acts as an Axe Head Cast). The Diamond is consumed by the recipe and does not switch slots with the Axe Head.
+ * 2.1) The result Item needs to be an Item that works with Materials, you can find a list of valid items by running `/ct dump ticMaterialItems`.
+ * 3) Adds a Composite Casting Recipe that uses Iron Tool Parts (Iron Axe Head for example) and Water (Fluid) to create a Bone version of that part. The recipe has a cooling temperature of 0°C. 
+ * 3.1) Composite Casting Recipes are Recipes where the Fluid is Cast onto a pre-existing tool part to create a new part.
+ * 3.2) You can find valid materialIds by running `/ct dump ticMaterials` in-game.
+ * 4) Adds a Container Filling Recipe that will fill a Copper Can with 100mb of any Fluid when cast.
+ * 4.1) This script is commented out as it conflicts with default Tinkers Recipes.
```
KnightMiner, line 15: Thats not going to work anyways. The number is just a hint to how much fluid we need to consume, it does not overcome the fact that containers need a specific number of mb input

jaredlll08: So you could add a recipe that would use 100mb but fill a bucket fully?

KnightMiner: No, you cannot. The recipe would fail to fill the container, but I think it would consume the fluid. The main reason for the number is so you can set a sane amount for variable size containers, like if you wanted to try to fill a thermal expansion portal tank.

For something in the mod to try, maybe a pickaxe? You can add a tank modifier to them, which accepts variable amounts of fluid.

src/main/resources/data/tconstruct/scripts/tconstruct/casting_table.zs
```diff
@@ -0,0 +1,36 @@
...
+ * 4.2) The amount of fluid is the maximum amount of fluid that can be inserted into the container at a time.
+ */
+
+// <recipetype:tconstruct:casting_table>.addItemCastingRecipe(name as string, cast as IIngredient, fluid as IFluidStack, result as IItemStack, coolingTime as int, consumeCast as boolean, switchSlots as boolean)
+// <recipetype:tconstruct:casting_table>.addMaterialCastingRecipe(name as string, cast as IIngredient, fluidAmount as int, result as Item, consumeCast as boolean, switchSlots as boolean)
+// <recipetype:tconstruct:casting_table>.addCompositeCastingRecipe(name as string, materialId as string, fluidstack as IFluidStack, outputMaterialId as string, coolingTemperature as int)
+// <recipetype:tconstruct:casting_table>.makeContainerFillingRecipe(name as string, fluidAmount as int, containerIn as Item)
+
+<recipetype:tconstruct:casting_table>.addItemCastingRecipe("item_casting_table_test", <item:minecraft:honey_bottle>, <fluid:tconstruct:blood> * 1000, <item:minecraft:diamond>, 200, false, true);
+<recipetype:tconstruct:casting_table>.addMaterialCastingRecipe("material_casting_table_test", <item:minecraft:diamond>, 250, <item:tconstruct:axe_head>, true, false);
```
KnightMiner: `tconstruct:axe_head` got renamed to `tconstruct:small_axe_head`

src/main/resources/data/tconstruct/scripts/tconstruct/casting_basin.zs
```diff
@@ -0,0 +1,36 @@
...
+ * 4.2) The amount of fluid is the maximum amount of fluid that can be inserted into the container at a time.
+ */
+
+// <recipetype:tconstruct:casting_basin>.addItemCastingRecipe(name as string, cast as IIngredient, fluid as IFluidStack, result as IItemStack, coolingTime as int, consumeCast as boolean, switchSlots as boolean)
+// <recipetype:tconstruct:casting_basin>.addMaterialCastingRecipe(name as string, cast as IIngredient, fluidAmount as int, result as Item, consumeCast as boolean, switchSlots as boolean)
+// <recipetype:tconstruct:casting_basin>.addCompositeCastingRecipe(name as string, materialId as string, fluidstack as IFluidStack, outputMaterialId as string, coolingTemperature as int)
+// <recipetype:tconstruct:casting_basin>.makeContainerFillingRecipe(name as string, fluidAmount as int, containerIn as Item)
+
+<recipetype:tconstruct:casting_basin>.addItemCastingRecipe("item_casting_basin_test", <item:minecraft:honey_bottle>, <fluid:tconstruct:blood> * 1000, <item:minecraft:diamond>, 200, false, false);
+<recipetype:tconstruct:casting_basin>.addMaterialCastingRecipe("material_casting_basin_test", <item:minecraft:diamond>, 250, <item:tconstruct:axe_head>, true, false);
```
KnightMiner: `tconstruct:axe_head` got renamed to `tconstruct:small_axe_head`

src/main/resources/data/tconstruct/scripts/tconstruct/melting.zs
```diff
@@ -0,0 +1,32 @@
...
+ * 4.3) The item needs to be an Item that works with Materials, you can find a list of valid items by running `/ct dump ticMaterialItems`.
+ */
+
+// <recipetype:tconstruct:melting>.addMeltingRecipe(name as string, input as IIngredient, output as IFluidStack, temperature as int, time as int, @Optional byproduct as IFluidStack[])
+// <recipetype:tconstruct:melting>.addDamageableMeltingRecipe(name as string, input as IIngredient, output as IFluidStack, temperature as int, time as int, @Optional byproduct as IFluidStack[])
+// <recipetype:tconstruct:melting>.addOreMeltingRecipe(name as string, input as IIngredient, output as IFluidStack, temperature as int, time as int, @Optional byproduct as IFluidStack[])
+// <recipetype:tconstruct:melting>.addMaterialMeltingRecipe(name as string, item as Item, cost as int)
+
+<recipetype:tconstruct:melting>.addMeltingRecipe("melting_test", <item:minecraft:diamond>, <fluid:minecraft:water>, 0, 50);
+<recipetype:tconstruct:melting>.addDamageableMeltingRecipe("damageable_melting_test", <item:minecraft:diamond_axe>, <fluid:minecraft:water> * 1000, 0, 50);
```
KnightMiner: So, we added molten diamond as a fluid making this recipe redundant. I would suggest elytra or wooden tools, it won't melt any time soon.

I can still test it showing in JEI so at least I know it works

src/main/resources/data/tconstruct/scripts/tconstruct/melting.zs
```diff
@@ -0,0 +1,32 @@
...
+ * 
+ * 1) Adds a normal Melting Recipe that produces 1mb of Water (Fluid) when a Diamond is melted at 0°C.
+ * 2) Adds a Damageable Melting Recipe that producess 1000mb of Water (Fluid) when a Diamond Axe is melted at 0°C.
+ * 2.1) Damageable Melting Recipes are recipes who's output amount (the amount of Fluid that is produced) changes depending on how damaged the input item is.
+ * 3) Adds an Ore Melting recipe that produces 1mb of Water (Fluid) when a Stick is melted at 0°C.
+ * 3.1) Ore Melting Recipes are recipes who's output amount (the amount of Fluid that is produced) changes depending on which Controller (Smeltery or Melter) the recipe is made in.
+ * 3.2) This recipe produces a byproduct of 50mb of Lava (Fluid).
+ * 4) Adds a Material Melting Recipe that allows an Axe Head to be melted into 5 ingots worth of the material the Axe Head was made of.
+ * 4.1) Material Melting Recipes are recipes who's output Fluid depends on the Material of the input Item, for example, if you smelt a Gold Axe Head, the recipe will produce 5 ingots worth (720mb) Molten Gold (Fluid).
+ * 4.2) This script is commented out as it conflicts with default Tinkers Recipes.
```
KnightMiner: Is it possible to add and remove the recipe so you can make the axe head melt for more? Otherwise this handler will get little use honestly as all addons with new tool parts should be adding a copy for their parts

src/main/resources/data/tconstruct/scripts/tconstruct/part_builder.zs
```diff
@@ -0,0 +1,20 @@
...
+ *
+ * The pattern is set as the Axe Head pattern, so in the Part Builder, the recipe will show under the Axe Head, but it will still make Tool Rods.
+ * Valid patterns are defined by the texture in the `tconstruct:gui/tinker_pattern/` folder in the resources.
+ *
+ * The output Item needs to be an Item that works with Materials, you can find a list of valid items by running `/ct dump ticMaterialItems`.
+ */
+
+// <recipetype:tconstruct:part_builder>.addMaterialRecipe(name as string, pattern as String, cost as int, output as Item, outputCount as int)
+
+<recipetype:tconstruct:part_builder>.addMaterialRecipe("part_builder_test", "tconstruct:axe_head", 2, <item:tconstruct:pickaxe_head>, 3);
```
KnightMiner: The pattern name is arbitrary, I would suggest using something that is not a tool part so it does not conflict with existing recipes

src/main/resources/data/tconstruct/scripts/tconstruct/casting_basin.zs
```diff
@@ -0,0 +1,36 @@
...
+ *  
+ * 1) Removes the all Casting Basin recipes for Blazewood.
+ * 2) Removes the Copper Can filling recipe.
+ */
+
+// <recipetype:tconstruct:casting_basin>.removeRecipe(output as IItemStack)
+// <recipetype:tconstruct:casting_basin>.removeByName(name as string)
+
+<recipetype:tconstruct:casting_basin>.removeRecipe(<item:tconstruct:blazewood>);
+<recipetype:tconstruct:casting_basin>.removeByName("tconstruct:smeltery/casting/filling/copper_can");
```
KnightMiner: This recipe does not exist for basins, thats a casting table recipe

src/main/resources/data/tconstruct/scripts/tconstruct/tinker_station.zs
```diff
@@ -0,0 +1,83 @@
...
+<recipetype:tconstruct:tinker_station>.addIncrementalAbilityModifierRecipe("incremental_ability_modifier_example", <item:minecraft:brown_wool>, 1, 2, <tag:items:tconstruct:modifiable>, "tconstruct:luck", 1, 4, 1, <item:minecraft:brown_dye>);
+<recipetype:tconstruct:tinker_station>.addIncrementalAbilityModifierRecipe("incremental_ability_modifier_requirements_example", <item:minecraft:magenta_wool>, 1, 2, <tag:items:tconstruct:modifiable>, "tconstruct:antiaquatic", 1, 4, 1, <item:minecraft:magenta_wool>, {"tconstruct:diamond": 1, "tconstruct:haste": 2}, 2, "recipe.tconstruct.modifier.requirements_error");
+
+/*
+ * Removes the Expanded upgrade from the Tinker Station.
+ */
+
+// <recipetype:tconstruct:tinker_station>.removeByName(name as string)
+
+<recipetype:tconstruct:tinker_station>.removeByName("tconstruct:tools/modifiers/upgrade/expanded");
```
KnightMiner: That recipe was wrongly put in the upgrade folder, should have been in ability, so that path no longer exists.

You want `"tconstruct:tools/modifiers/ability/expanded"` probably

src/main/resources/data/tconstruct/scripts/tconstruct/casting_basin.zs
```diff
@@ -0,0 +1,36 @@
+/*
+ * Adds four Casting Basin recipes that do the following:
+ * 
+ * 1) Adds an Item Casting Recipe that, when a Honey Bottle is used as a Cast for 1000mb of Blood (Fluid), it gets cast into a Diamond of 200 ticks of cooling time. The Honey Bottle is not consumed and does not switch slots with the Diamond.
+ * 2) Adds a Material Casting Recipe that will allow a Diamond to be used as a Cast to make any Axe Head (Part), regardless of material (The Diamond acts as an Axe Head Cast). The Diamond is consumed by the recipe and does not switch slots with the Axe Head.
+ * 2.1) The result Item needs to be an Item that works with Materials, you can find a list of valid items by running `/ct dump ticMaterialItems`.
+ * 3) Adds a Composite Casting Recipe that uses Iron Tool Parts (Iron Axe Head for example) and Water (Fluid) to create a Bone version of that part. The recipe has a cooling temperature of 0°C.
```
KnightMiner, line 11: Worth noting that composite casting recipes will only work with parts that have an existing material casting recipe. The basin has none by default, so it may make people wonder why the recipe is not working.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/EntityMeltingManager.java
```diff
@@ -0,0 +1,56 @@
...
+    CraftTweakerAPI.apply(new ActionAddRecipe(this, recipe));
+  }
+  
+  @Override
+  public void removeRecipe(IItemStack output) {
+    throw new IllegalArgumentException("Cannot remove Entity Melting Recipes by an IItemStack output as it outputs Fluids! Use `removeRecipe(Fluid output)` instead!");
+  }
+  
+  @ZenCodeType.Method
+  public void removeRecipe(Fluid output) {
```
KnightMiner, line 41: Would it be worth having a remove by entity variant like you have for beheading? Also worth noting that removing a recipe means that entity now melts into the default (blood)

src/main/resources/data/tconstruct/scripts/tconstruct/melting.zs
```diff
@@ -0,0 +1,32 @@
...
+ * 4.3) The item needs to be an Item that works with Materials, you can find a list of valid items by running `/ct dump ticMaterialItems`.
+ */
+
+// <recipetype:tconstruct:melting>.addMeltingRecipe(name as string, input as IIngredient, output as IFluidStack, temperature as int, time as int, @Optional byproduct as IFluidStack[])
+// <recipetype:tconstruct:melting>.addDamageableMeltingRecipe(name as string, input as IIngredient, output as IFluidStack, temperature as int, time as int, @Optional byproduct as IFluidStack[])
+// <recipetype:tconstruct:melting>.addOreMeltingRecipe(name as string, input as IIngredient, output as IFluidStack, temperature as int, time as int, @Optional byproduct as IFluidStack[])
+// <recipetype:tconstruct:melting>.addMaterialMeltingRecipe(name as string, item as Item, cost as int)
+
+<recipetype:tconstruct:melting>.addMeltingRecipe("melting_test", <item:minecraft:diamond>, <fluid:minecraft:water>, 0, 50);
```
KnightMiner: Diamond is now a molten type, so maybe wood as an example?

src/main/resources/data/tconstruct/scripts/tconstruct/melting.zs
```diff
@@ -0,0 +1,32 @@
...
+<recipetype:tconstruct:melting>.addDamageableMeltingRecipe("damageable_melting_test", <item:minecraft:diamond_axe>, <fluid:minecraft:water> * 1000, 0, 50);
+<recipetype:tconstruct:melting>.addOreMeltingRecipe("ore_melting_test", <item:minecraft:stick>, <fluid:minecraft:water>, 0, 50, [<fluid:minecraft:lava> * 50]);
```
KnightMiner: 1mb does not scale super well to the ore rates, those being 1.33x, 1.67x, and 2.0x. Probably do something 100 or more.

src/main/java/slimeknights/tconstruct/plugin/crt/managers/TinkerStationManager.java
```diff
@@ -0,0 +1,144 @@
...
+  }
+
+  @ZenCodeType.Method
+  public void addIncrementalAbilityModifierRecipe(String name, IIngredient input, int amountPerInput, int neededPerLevel, IIngredient toolRequirement, String modifierResult, int modifierResultLevel, int maxLevel, int abilitySlots, IItemStack leftover, @ZenCodeType.Optional IData modifierRequirements, @ZenCodeType.OptionalInt(-1) int minMatch, @ZenCodeType.OptionalString(defaultError) String requirementsError) {
+    addIncrementalModifierRecipeInternal(name, input, amountPerInput, neededPerLevel, toolRequirement, modifierResult, modifierResultLevel, maxLevel, 0, abilitySlots, leftover, modifierRequirements, minMatch, requirementsError);
+  }
+
+  @Override
+  public void removeRecipe(IItemStack output) {
+    throw new IllegalArgumentException("Cannot remove Tinker Station Recipes by an IItemStack output! Use `removeByName(String name)` instead!");
```
KnightMiner, line 81: Removing by modifier ID might be useful In case I want to remove all recipes for haste for instance. Can do so using recipes that implement `IDisplayModifierRecipe`, assuming you can use an interface like that for removal.

src/main/resources/data/tconstruct/scripts/tconstruct/tinker_station.zs
```diff
@@ -0,0 +1,83 @@
...
+ * 8.1) The Modifier can be applied 3 times, for a total of Haste 3.
+ * 8.2) The Recipe has 8 increments and each Black Wool adds 2 increments, meaning a total of 4 Black Wool will be required per level.
+ * 8.3) If there are any increments left over (if there was somehow 7 increments and another Black Wool is used, for a total of 9 increments), the recipe will return a Black Dye.
+ * 9) Adds an Incremental Modifier Recipe that uses Yellow Wool to add the Sharpness Modifier to any item in the tconstruct:modifiable Item Tag.
+ * 9.1) The Modifier can be applied 3 times, for a total of Sharpness 3.
+ * 9.2) The Recipe has 8 increments and each Yellow Wool adds 3 increments, meaning a total of 3 Yellow Wool will be required per level.
+ * 9.3) If there are any increments left over (if there was 6 increments and another Yellow Wool is used, for a total of 9 increments), the recipe will return a Yellow Dye.
+ * 9.4) The Recipe requires the Tool that it is being applied on to have the Emerald Modifier at level 1 or more.
+ * 10) Adds an Incremental Upgrade Modifier Recipe that uses Blue Wool to add the Bane of Arthropods Modifier to any item in the tconstruct:modifiable Item Tag.
```
KnightMiner, line 34: For incremental recipes, possibly worth mentioning that you should not have two incremental recipes for the same modifier with different sizes. Will lead to unexpected behavior, and the log yells at you. For the example recipes its fine, but for the end user's pack its important to say somewhere

KnightMiner (commented):
Overall looks good, some of your examples need updating for items/recipes that were renamed, or because we added recipes using the same input since this PR opened. Other than those, I have two main comments I left:

* Would be good to remove modifier recipes by modifier ID. You can make use of `IDisplayModifierRecipe` for this.
* Would be good to remove entity melting recipes by entity ID. Can do the same thing you did for beheading

src/generated/resources/data/tconstruct/recipes/tools/parts/melting/pickaxe_head.json
```diff
@@ -1,5 +1,5 @@
 {
   "type": "tconstruct:material_melting",
-  "item": "tconstruct:pickaxe_head",
+  "item": "tconstruct:pick",
```
KnightMiner: Why was this file changed? Were you testing something?

KnightMiner (approved):

PR #4399, opened 2021-04-23
1.16.5 

xVodkaX:
hello is it possible to make version 1.7.10 work for minecraft 1.16.5? (or it doesn't work because the code is no longer the same)

I downloaded the code for tconstruct 1.16.5 and I added ardite in the code the problem and I do not know how to recompile the mod it does not work I would like to know if I could have some help or a link for a tutorial please


KnightMiner:
That is not at all how modding works. If you want 1.7.10 features on 1.16, code it yourself. There is not a button you can hit to make it magically work, otherwise porting would not have taken so long.

As far as recompiling, this is not the place for basic modding tutorials. Go ask on the Minecraft Forums or the Forge Forums.

PR #4415, opened 2021-04-28
Right click modifier hooks

PrincessOfEvil:
First draft of a simple way to run right click hooks. Moves all existing right clicks on tools onto modifiers that run as late as physically possible.
Hooks added are:
* onItemUseFirst: Called when the tool is used on a block.
* itemInteractionForEntity: Called when the tool is used on an entity.
* onItemRightClick: Called when the tool is used on air, or otherwise when no other use hook causes an action first.

* onPlayerStoppedUsing: Called when the tool stopped being used.
* getUseDuration: For how long the tool can be held rightclicked.
* getUseAction: The action shown while the tool is being held.


src/main/java/slimeknights/tconstruct/tools/modifiers/ability/BlockTransformModifier.java
```diff
@@ -0,0 +1,46 @@
...
+
+  @Override
+  public boolean shouldDisplay(boolean advanced) {
+    return priority > Short.MIN_VALUE;
+  }
+
+  @Override
+  public ActionResultType onItemUse(ToolStack tool, int level, ItemStack stack, ItemUseContext context) {
+    Item item = stack.getItem();
+    if (item instanceof ToolCore) {
```
KnightMiner: There is an interface you can use instead of `ToolCore`, I try to avoid direct dependencies on that class when possible (still not updated the old

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/BlockTransformModifier.java
```diff
@@ -0,0 +1,46 @@
...
+      ToolCore toolCore = (ToolCore) item;
+      return toolCore.getToolHarvestLogic().transformBlocks(context, toolType, sound, requireGround);
```
KnightMiner, line 41: I half wonder if this logic should move into the modifier instead of tool harvest logic. None of the tools should be overriding it, it was just in ToolHarvestLogic as a clean place to reuse it.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/KamaAbilityModifier.java
```diff
@@ -0,0 +1,280 @@
...
+import net.minecraftforge.eventbus.api.Event.Result;
+import slimeknights.tconstruct.common.TinkerTags;
+import slimeknights.tconstruct.library.modifiers.SingleUseModifier;
+import slimeknights.tconstruct.library.tools.events.TinkerToolEvent.ToolHarvestEvent;
+import slimeknights.tconstruct.library.tools.helper.ToolDamageUtil;
+import slimeknights.tconstruct.library.tools.helper.ToolHarvestLogic.AOEMatchType;
+import slimeknights.tconstruct.library.tools.item.ToolCore;
+import slimeknights.tconstruct.library.tools.nbt.ToolStack;
+
+public class KamaAbilityModifier extends SingleUseModifier {
```
KnightMiner: Rename this (and the relevant registration) to `HarvestAbilityModifier`. Kama contains it, but I've been calling it harvest elsewhere.

src/main/java/slimeknights/tconstruct/tools/TinkerModifiers.java
```diff
@@ -142,7 +148,15 @@
   public static final RegistryObject<AutosmeltModifier> autosmelt = MODIFIERS.register("autosmelt", AutosmeltModifier::new);
   public static final RegistryObject<Modifier> expanded = MODIFIERS.register("expanded", () -> new Modifier(0xff9f50));
   public static final RegistryObject<ReachModifier> reach = MODIFIERS.register("reach", ReachModifier::new);
-
+  
+  public static final RegistryObject<BlockTransformModifier> shovelTransform = MODIFIERS.register("shovel_transform", () -> new BlockTransformModifier(0x8a361e, Integer.MIN_VALUE, ToolType.SHOVEL, SoundEvents.ITEM_SHOVEL_FLATTEN, true));
```
KnightMiner: Minor, but toss in a `// internal abilities` header or something similar here, for consistency. These are "abilities", but not applied via modifiers.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/ShearsAbilityModifier.java
```diff
@@ -0,0 +1,106 @@
...
+   *
+   * @param player the current player
+   * @param hand the given hand the tool is in
+   */
+  protected void swingTool(PlayerEntity player, Hand hand) {
+    player.swingArm(hand);
+    player.spawnSweepParticles();
+  }
+
+  protected boolean isShears(ToolStack tool) {
```
KnightMiner: Javadoc

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
```diff
@@ -324,6 +330,41 @@ public void onInventoryTick(IModifierToolStack tool, int level, World world, Liv
...
+   * @param level 
+   * @param stack 
+   * @param context        Full item use context
+   * @return  Action result
+   */
+  public ActionResultType onItemUse(ToolStack tool, int level, ItemStack stack, ItemUseContext context) {
+    return ActionResultType.PASS;
+  }
+  
+  public ActionResultType itemInteractionForEntity(ToolStack tool, int level, ItemStack stack, PlayerEntity playerIn, LivingEntity target, Hand hand) {
```
KnightMiner: Any reason this is `ToolStack`? Most other modifier hooks use `IModifierToolStack`. If one of the modifiers downstream requires `ToolStack`, we should update that.

Reason being, eventually I plan to split off `ToolStack` to allow non-multipart modifiable items.

Aside, do we need the item stack here? I assume thats for the shearing, arguably they could refetch the item stack using the player (its cheap), I want to discourage using the item stack directly when possible

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/ShearsAbilityModifier.java
```diff
@@ -0,0 +1,106 @@
...
+    return true;
+  }
+  
+  @Override
+  public ActionResultType itemInteractionForEntity(ToolStack tool, int level, ItemStack stack, PlayerEntity playerIn, LivingEntity target, Hand hand) {
+    // only run AOE on shearable entities
+    if (isShears(tool) && target instanceof IForgeShearable) {
+      int fortune = EnchantmentHelper.getEnchantmentLevel(Enchantments.FORTUNE, stack);
```
KnightMiner: This is probably just outdated code as it won't work for our tools anyways. You could hardcode it to the level of luck, or you could throw the forge event that the luck modifier subscribes to.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
KnightMiner: All public methods should get either javadocs or `@Override`, especially in library. For these specifically, do the alternatives list like I did above, and note somewhere which order they run in. Also note the return value for pass and to prevent interaction

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
```diff
@@ -324,6 +330,41 @@ public void onInventoryTick(IModifierToolStack tool, int level, World world, Liv
...
+  }
+  
+  public ActionResult<ItemStack> onItemRightClick(IModifierToolStack tool, int level, World worldIn, PlayerEntity playerIn, Hand handIn) {
+    return ActionResult.resultPass(playerIn.getHeldItem(handIn));
+  }
+
+  /**
+  * @return  Whether the modifier should block any incoming ones from firing
+  */
+  public boolean onPlayerStoppedUsing(IModifierToolStack tool, int level, World worldIn, LivingEntity entityLiving, int timeLeft) {
```
KnightMiner: Might be worth mentioning how we get this hook to call in the javadoc (setting active hand, duration, etc.). I know its vanilla, but its useful information.

src/main/java/slimeknights/tconstruct/library/tools/item/ToolCore.java
```diff
@@ -374,7 +379,80 @@ public void inventoryTick(ItemStack stack, World worldIn, Entity entityIn, int i
...
   }
+  
+  /* Right click hooks */
+  
+  @Override
+  public ActionResultType onItemUseFirst(ItemStack stack, ItemUseContext context) {
+    ToolStack tool = ToolStack.from(stack);
+    for (ModifierEntry entry : tool.getModifierList()) {
+      ActionResultType result = entry.getModifier().onItemUse(tool, entry.getLevel(), stack, context);
+      if (result != ActionResultType.PASS) {
```
KnightMiner: Vanilla has two variants of `PASS`, the other is `FAIL`, should probably be using `isSuccessOrConsume` instead of an equals check, or were you intending `FAIL` to prevent other modifier hooks from running?

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
```diff
@@ -324,6 +330,41 @@ public void onInventoryTick(IModifierToolStack tool, int level, World world, Liv
...
 
+  /**
+    * Called when this item is used when targeting a Block
+   * @param tool 
+   * @param level 
+   * @param stack 
+   * @param context        Full item use context
+   * @return  Action result
+   */
+  public ActionResultType onItemUse(ToolStack tool, int level, ItemStack stack, ItemUseContext context) {
```
KnightMiner: Back in 1.12, I had a interaction between rapiers and shields where the rapier would leap back, but return pass so the shield could block during the leap. Would that be worth supporting directly?

Its a bit weird behavior, I just ask now as supporting it in the API later would require using a different enum likely, which would be a breaking change.

KnightMiner: Per below comment, any reason you are using `ToolStack` directly? Can we switch it to the interface and switch callers to support the interface as well? Likewise, do we need `ItemStack` here, or can they just fetch it from the context?

KnightMiner (changes requested):
Couple things related to documentation, couple functional changes, a couple questions

src/main/java/slimeknights/tconstruct/tools/TinkerModifiers.java
PrincessOfEvil: They could be, tbh. But sure.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
PrincessOfEvil: IForgeShearable. I guess i could switch it to player access, just feels kinda weird to be asking when the stack is right here.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
PrincessOfEvil: Why support directly? Nothing stops a modifier from doing their logic and returning a pass.

PrincessOfEvil: ToolHarvestEvent and ToolHarvestLogic.getAOEBlocks.

PrincessOfEvil: And Shield and Leap can just check for each other.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
PrincessOfEvil: That would mean i would need to learn all that. :'D

src/main/java/slimeknights/tconstruct/library/tools/item/ToolCore.java
PrincessOfEvil: Nah, modifiers that want to stop without running code can just `CONSUME`.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/BlockTransformModifier.java
PrincessOfEvil: ...Maybe someone might want to AoE transform blocks in a different place, i dunno.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/ShearsAbilityModifier.java
PrincessOfEvil: I'm... very much not smart enough to figure out what code i should be using here. :'D

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
KnightMiner: Mostly for a more general behavior. Saying rapier supports leaping back then any other tool action running. I suppose we could always do it via a pass return then returning success in another hook, so probably not a huge deal to support

KnightMiner: Change those methods to take an `IModifierToolStack`, let me know if anything there is not possible with the interface.

For the stack, if we need it we can always fetch from the context, just goes back to discouraging using it directly unless needed

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/BlockTransformModifier.java
KnightMiner: Could always make it static here. I just don't see a ton of potential for reuse of that outside this modifier. We could always do it after this PR as part of library cleanup, its just not something strictly related to AOE.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/ShearsAbilityModifier.java
KnightMiner: I suppose this is a case that was broken before your PR, not strictly related. Add a `// FIXME: wrong function` to here and we can fix it later

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
KnightMiner: IIRC, you use what the bow does, `setActiveHand` IIRC.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
PrincessOfEvil: Yeah, just pass it.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
```diff
@@ -372,8 +372,8 @@ public ActionResultType itemInteractionForEntity(IModifierToolStack tool, int le
    * <br>
    * Alternatives:
    * <ul>
-   *   <li>{@link #onItemUse(ToolStack, int, ItemStack, ItemUseContext)}: Processes use actions on blocks.</li>
-   *   <li>{@link #itemInteractionForEntity(IModifierToolStack, int, ItemStack, ItemUseContext)}: Processes use actions on entities.</li>
+   *   <li>{@link #onItemUse(ToolStack, int, ItemUseContext)}: Processes use actions on blocks.</li>
```
KnightMiner: This link needs update, wrong first parameter

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/BlockTransformModifier.java
```diff
@@ -35,8 +34,8 @@ public boolean shouldDisplay(boolean advanced) {
   }
 
   @Override
-  public ActionResultType onItemUse(ToolStack tool, int level, ItemStack stack, ItemUseContext context) {
-    Item item = stack.getItem();
+  public ActionResultType onItemUse(IModifierToolStack tool, int level, ItemUseContext context) {
+    Item item = context.getItem().getItem();
```
KnightMiner: You can do `tool.getItem()` here

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/BlockTransformModifier.java
PrincessOfEvil: Either way, this is kind of out of scope of this PR anyway and it can always be discussed later. ToolHarvestLogic just feels like a nicer place for it.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
KnightMiner: We just got it from player access in the same way earlier. Its mostly about keeping the API smaller, fewer things to mismatch (plus, most things won't be vanilla/forge hooks I assume)

src/main/java/slimeknights/tconstruct/tools/data/ModifierRecipeProvider.java
```diff
@@ -352,7 +384,34 @@ private void addModifierRecipes(Consumer<IFinishedRecipe> consumer) {
                          .setMaxLevel(2)
                          .setAbilitySlots(1)
                          .build(consumer, prefixR(TinkerModifiers.reach, upgradeFolder));
-
+    // block transformers
+    ModifierRecipeBuilder.modifier(TinkerModifiers.shovelTransform.get())
+                         .setTools(new IngredientWithout(Ingredient.fromTag(TinkerTags.Items.HARVEST), Ingredient.fromItems(TinkerTools.mattock, TinkerTools.excavator)))
+                         .addInput(TinkerModifiers.ancientShovelHead.get())
+                         .setMaxLevel(1)
+                         .setUpgradeSlots(1)
```
KnightMiner: These transform modifiers let you essentially replace the need for another tool (notably a kama, as thats the mosst useful one), so I am inclined to call them abilities.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/PlaceBlockModifier.java
```diff
@@ -0,0 +1,249 @@
...
+import net.minecraft.util.math.shapes.ISelectionContext;
+import net.minecraft.world.World;
+import slimeknights.tconstruct.library.Util;
+import slimeknights.tconstruct.library.modifiers.SingleUseModifier;
+import slimeknights.tconstruct.library.tools.nbt.IModifierToolStack;
+import slimeknights.tconstruct.library.tools.nbt.ToolStack;
+
+/** Modifier version of a {@link net.minecraft.item.BlockItem}
+ * */
+public abstract class PlaceBlockModifier extends SingleUseModifier /*, BlockItem*/ {
```
KnightMiner: I feel like this is way overcomplicated for something that would be as simple as an `onItemUse` hook. For the blocks that tools will actually place, we don't need 90% of this code. Notably, we have no need for a tile entity placing behavior right now, and I'd rather not have that much dead code sitting untested. Plus a lot of this is hardcoded in a way you cannot have multiple block placement modifiers.

Just add a hook into `onItemUse`, and place the block using `setBlockState`.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/GlowingModifier.java
```diff
@@ -0,0 +1,31 @@
...
+  @Override
+  public ActionResultType onItemUse(IModifierToolStack tool, int level, ItemUseContext context) {
+    if (tool.getCurrentDurability() >= 5 && context.getPlayer().isSneaking()) {
+      if (!context.getWorld().isRemote) {
+        if (TinkerCommons.glow.get().addGlow(context.getWorld(), context.getPos(), context.getFace().getOpposite())) {
+          tool.setDamage(tool.getDamage() + 5);
+          }
+        }
+      return ActionResultType.func_233537_a_(context.getWorld().isRemote);
+    }
```
KnightMiner: Looks like this is targeting the wrong position, should offset backwards by the face. Also, `getPlayer()` is nulllable on the context. The snippet below addresses those issues and adds a sound effect.

```java
    PlayerEntity player = context.getPlayer();
    if (tool.getCurrentDurability() >= 5 && (player == null || player.isSneaking())) {
      if (!context.getWorld().isRemote) {
        World world = context.getWorld();
        Direction face = context.getFace();
        BlockPos pos = context.getPos().offset(face);
        if (TinkerCommons.glow.get().addGlow(world, pos, face.getOpposite())) {
          tool.setDamage(tool.getDamage() + 5);
          world.playSound(null, pos, world.getBlockState(pos).getSoundType(world, pos, player).getPlaceSound(), SoundCategory.BLOCKS, 1.0f, 1.0f);
        }
      }
      return ActionResultType.func_233537_a_(context.getWorld().isRemote);
    }
 ```

File: src/main/java/slimeknights/tconstruct/tools/data/ModifierRecipeProvider.java [hunk 16]
```diff
@@ -88,6 +93,33 @@ private void addItemRecipes(Consumer<IFinishedRecipe> consumer) {
                           .addIngredient(TinkerTags.Items.WITHER_BONES)
                           .addCriterion("has_bone", hasItem(TinkerTags.Items.WITHER_BONES))
                           .build(withCondition(consumer, ConfigEnabledCondition.WITHER_BONE_CONVERSION), location(folder + "wither_bone_conversion"));
+    
+    // TODO: use addCastingWithCastRecipe
+    FluidStack debris = new FluidStack(TinkerFluids.moltenDebris.get(), MaterialValues.INGOT);
```
KnightMiner, line 99: These should probably melt back into debris as well. Can make one recipe that accepts all 3 as input probably

KnightMiner (approved):

KnightMiner (approved):

PR #4416, opened 2021-04-28
Add minecraft:impermeable tag to clear class, fixes #4412

### LINKED ISSUE #4412: Clear Glass Water Dripping
Could the clear glasses also be able to stop water doplets like vanilla glass?

Current Version: Tinkers' Construct 3.0.3.126 for 1.16.5

sciguyryan:
This commit adds the minecraft:impermeable tag to the variants of clear class, which puts them in line with vanilla glass. This fixes bug #4412.

The following is a screenshot demonstrating the fix. No particle effects from the water can be seen dripping from the bottom of the glass after this change.

### DIFF
--- a/src/generated/resources/data/minecraft/tags/blocks/impermeable.json
+++ b/src/generated/resources/data/minecraft/tags/blocks/impermeable.json
@@ -1,0 +1,22 @@
+{
+  "replace": false,
+  "values": [
+    "tconstruct:clear_glass",
+    "tconstruct:white_clear_stained_glass",
+    "tconstruct:orange_clear_stained_glass",
+    "tconstruct:magenta_clear_stained_glass",
+    "tconstruct:light_blue_clear_stained_glass",
+    "tconstruct:yellow_clear_stained_glass",
+    "tconstruct:lime_clear_stained_glass",
+    "tconstruct:pink_clear_stained_glass",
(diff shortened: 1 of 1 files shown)


KnightMiner:
Thats not how you add generated tags. Them being generated means we use code that automatically creates them. Because of where you put the tag it will be automatically deleted next time datagenerators run (there is a different directory that will not be deleted by datagenerators). Plus, should probably tag seared glass as well.

If you want to do this fix properly using datagenerators, feel free to update your PR. Otherwise I will just do the fix myself when I work on the mod again.

sciguyryan:
> 
> 
> Thats not how you add generated tags. Them being generated means we use code that automatically creates them. Because of where you put the tag it will be automatically deleted next time datagenerators run (there is a different directory that will not be deleted by datagenerators). Plus, should probably tag seared glass as well.
> 
> If you want to do this fix properly using datagenerators, feel free to update your PR. Otherwise I will just do the fix myself when I work on the mod again.

I will do so if I have time tomorrow between uni work. I've made a note of that for the future in any event, thanks for the heads up.

sciguyryan:
I am going to close this for now, I likely won't have time to come back and look at this for a few days. By then you will probably have looked at it yourself.

Again, thanks for the review. I will take a look at how you resolve this for future reference!

PR #4433, opened 2021-05-03
Add minecraft:impermeable tag to glass blocks, fixes #4412

### LINKED ISSUE #4412: Clear Glass Water Dripping

sciguyryan:
This commit adds the minecraft:impermeable tag to the various glass blocks added by Tinkers. This puts the glass variants in line with vanilla glass, preventing fluids from "dripping" through the block. This fixes bug #4412.

This is attempt #2, this time using generators! I have included the output JSON file, which was generated using the Gradle runData command.

Note: there were a bunch of other JSON files created or modified, but I have only included the one that relates to this commit.

### DIFF
--- a/src/generated/resources/data/minecraft/tags/blocks/impermeable.json
+++ b/src/generated/resources/data/minecraft/tags/blocks/impermeable.json
@@ -1,0 +1,24 @@
+{
+  "replace": false,
+  "values": [
+    "tconstruct:clear_glass",
+    "tconstruct:soul_glass",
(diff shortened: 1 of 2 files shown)


KnightMiner:
The block registry lookup was a overcomplicated hack that I just had not gotten around to removing. Our enum objects let us fetch all clear stained glass in a less fragile way.

PR #4496, opened 2021-05-28
Balance changes for Netherite and Necrotic modifiers

mDiyo:
Netherite changes: 
- 1.2x boost to durability and mining speed, down from 1.25x
- 1.1x boost to attack damage, down from 1.25x

Necrotic:
- Entities only heal on kill
- Healing amount has been buffed from 0-5% per level to 2-10% per level


src/main/java/slimeknights/tconstruct/tools/modifiers/shared/NecroticModifier.java
```diff
@@ -13,9 +13,9 @@ public NecroticModifier() {
 
   @Override
   public int afterLivingHit(IModifierToolStack tool, int level, LivingEntity attacker, LivingEntity target, float damageDealt, boolean isCritical, float cooldown) {
-    if (cooldown > 0.9 && damageDealt > 0) {
+    if (cooldown > 0.9 && damageDealt > 0 && target.getHealth() <= 0) {
       // heals between 0 and (level) * 5% of damage dealt
-      float heal = attacker.getRNG().nextFloat() * damageDealt * level * 0.05f;
+      float heal = attacker.getRNG().nextFloat() * damageDealt * level * 0.08f + damageDealt * level * 0.02f;
```
KnightMiner, line 18: Add some parenthesis here, as the order of operations gets a little unclear with such long chains. Math makes sense but its a little unreadable.

src/main/java/slimeknights/tconstruct/tools/modifiers/upgrades/NetheriteModifier.java
```diff
@@ -21,9 +21,9 @@ public void addVolatileData(ToolDefinition toolDefinition, StatsNBT baseStats, I
 
   @Override
   public void addToolStats(ToolDefinition toolDefinition, StatsNBT baseStats, IModDataReadOnly persistentData, IModDataReadOnly volatileData, int level, ModifierStatsBuilder builder) {
-    builder.multiplyDurability(1.25f);
-    builder.multiplyAttackDamage(1.25f);
-    builder.multiplyMiningSpeed(1.25f);
+    builder.multiplyDurability(1.20f);
+    builder.multiplyAttackDamage(1.10f);
```
KnightMiner, line 25: 10% might be a little low. Vanilla axes are 11%, and vanilla swords are 14%. Not sure if that is significant, so worth discussing on discord.

PR #4588, opened 2021-08-01
Create ru_ru.json

AlgorithmLX:


KnightMiner:
For the record, I did not intentionally ignore this, the mod is just changing fast enough I'm not sure I see much value in translating yet, so I've been holding off on reviewing these. It might be easier to do once I mark the mod beta.

AlgorithmLX:
that's why i closed it initially Pull request,  for I saw that people before me translated the mod.

PR #4606, opened 2021-08-18
Create vi_vn.json

ALR2310:
I hope this localization is added to the game.


KnightMiner:
There is a lot missing in this translation. The first 30 or so item/block names are entirely untranslated, just seems like the tooltips were translated.

Generally its good to translate both item names and tooltips.

KnightMiner:
Any update on this? Still looks like you have not translated much of the mod, too many item names are left untranslated.

PR #4622, opened 2021-09-16
fix typo ("sandwhich" -> "sandwich")

joker876:


KnightMiner:
Its more effort to review and test a PR like this then it is to make the change myself. I doubt this is the only typo, so if you want to scan all translations and fix errors I will look into that PR. Otherwise I will just fix this typo directly

joker876:
I am in the process of translating the mod, and am 75% done. So far it's the only typo I found

joker876:
It actually IS the only typo I found. Do what you will with this PR

joker876:
Actually, I will include this fix within another PR, and close this one

PR #4644, opened 2021-10-21
Fixed error in class doc comment

LMongoose:


KnightMiner:
I don't really care about PRs fixing a single javadoc comment. That is probably not the only typo in the whole codebase, if you want to fix them all I'll consider the PR but something like this is so trivial I could do it myself more easily than the effort it would take to merge this PR into my current workspace.

PR #4652, opened 2021-11-04
Assets optimization

K0-RR:
Lossless compression using the Leanify tool


KnightMiner:
For a PR like this, I would want some sort of guarantee there is no loss of quality. Unfortunately, the best way I can think of to do that is manually reviewing each texture, which would take forever with 1539 files. In addition, the mod is still in alpha so images are subject to change. Anyways, as I mentioned biggest problem is the scope of this PR, I have never seen your name before, and while I do not doubt good intentions, its possible among those images is a malicious change so I would not feel comfortable merging without reviewing each one. It may end up being easier to just download the tool and run the optimization myself.

----

I think what I will suggest doing is splitting this into several PRs. Start with one targeting the files that give the biggest benefit, such as book images and other screenshots. Leave out individual sprites like items or blocks for now as those make up the majority off the textures.

Providing some stats on the total benefit would also help your case. I see individual files are -X bytes, what is the overall change? Can we view those numbers broken down by blocks/items/books?

BeansBeefBroccoli:
Oh I've seen this dude before, but ig u can't trust me either lol

K0-RR:
>It may end up being easier to just download the tool and run the optimization myself.

This is the tool I used https://github.com/JayXon/Leanify, sadly there is no option to exclude files - it optimizes everything it can and for some reason, it breaks .md files like readme. I use Github Desktop and discard broken or unnecessary compressions (like compressed JSON as it makes it harder to read them, or nbt files) before submitting PR. 

**Oxipng** is a good alternative (with -Z or at least -D mode, for example - `oxipng -arvZ -o max --strip all --fix`) but its compression is slightly worse and slower. It only works on PNG files. (For JPG use **jpegoptim**)
Instead of using it manually, you can use this application https://github.com/EnnuiL/Detonater/ as it can help with automating this process.

There is also this **GitHub bot** https://imgbot.net/ - it's good because you can enable it for all your current and future repositories and forget about it. Then, if there is anything that can be optimized it will open ready to merge PR. Just keep in mind imgbot's lossless compression is not as good as the tools I mentioned before.

>what I will suggest doing is splitting this into several PRs. Start with one targeting the files that give the biggest benefit

Sure, I could do it this way if you don't want to compress by yourself.

>I would want some sort of guarantee there is no loss of quality

While the tools I use are verified to be lossless I understand that trusting some big PR from a random person may not be reasonable.

>Providing some stats on the total benefit would also help

Here are the results from the bot I mentioned earlier - 20% smaller but it did way less files  - https://github.com/RDKRACZ/TinkersConstruct/pull/2

Looking at my compression with the leanify - all png files weighed 5,6 MB (book, textures, logo.png together) and now weigh 3,8 MB. If you only count the book subfolder - it was 4.6 MB, went down to 3,2 MB.

KnightMiner:
2a4dfd5 implemented this.

PR #4687, opened 2021-12-23
Italian translation (1.16)

maicol07:


Insane96:
`Sticks all sticky blocks besides ichor` should definitely be `Si attacca a tutti i blocchi appiccicosi eccetto l'ichor` (Google Translate gives me about what you've translated it into 😏)

I would have translated `Congealed` to `Rappreso` as it's more a solidification than like blood.

`Clear glass` as `Vetro Pulito` sounds like normal glass is dirty. Should be `Vetro Limpido`. Also that's some mismatch with the later translations as `Clear` is translated as `Trasparente` and no longer `Pulito`.

Also order of the translations is messed up. The `gadgets advancement` comment is now on top.

Will definitely check out more when I have time.

maicol07:
@Insane96 thank you for the feedback! I've fixed the first three issues. I don't think order is so important, since it's a JSON key-value file

Insane96:
Oh, forgot to mention that I'm pretty sure the file name must be all lowercase since around MC 1.12

KnightMiner:
Yeah, file needs to be lowercase, making me suspect this needs more testing in game. Also has not been updated recently, so likely missing a lot of content. In the case the author in inactive I can review this again closer to the final translation update, but that means a much lower quality translation will be added.

maicol07:
> Yeah, file needs to be lowercase, making me suspect this needs more testing in game. Also has not been updated recently, so likely missing a lot of content. In the case the author in inactive I can review this again closer to the final translation update, but that means a much lower quality translation will be added.

I'll check them and update the translation

maicol07:
The translation is now updated and ready to merge!

KnightMiner:
I am merging this to ensure if we do another 1.16 release its included,  but at this time another 1.16 release is not guaranteed.

PR #4708, opened 2022-01-05
Update russian translation

Externallex:


Externallex:
@KnightMiner, this can be merged with branches 1.16 and 1.18. However, for 1.18 will need to update some keys and translate new strings, but this is already in the new PR for 1.18.

KnightMiner:
Sorry for the late response.

1.18 I would rather have a second translation made even if you start by copying this one. I can always copy it over directly to give that version a heads up.

If you have time, please look over this and double check you have the new strings I added in the final 1.16.5 feature update, there will be no new features after that update, so should be relatively easy to make one more build with translations.

KnightMiner:
I am merging this to ensure if we do another 1.16 release its included, but at this time another 1.16 release is not guaranteed.

PR #4718, opened 2022-01-15
Update materials.xml

AL523-1:
incorrect display of materials in the book
https://drive.google.com/file/d/1OSNmHaJw1Kj0IFGW1g0Qxa-KQUrfKu2x/view?usp=sharing


KnightMiner:
We have not released a 1.7.10 build for 6 years. Even if I merged this, it would never be released. Make a resource pack to fix this if you want.

PR #4761, opened 2022-02-16
Add usage information to slurping modifier

roosemberth:
The manual is very clear on what this effect does, but it is not clear to a new player how to fill the helmet.

I may have been a bit distracted, but had to come and look at the code to understand this. I think it should be part of the in-game documentation.


KnightMiner:
That sentence you added is very ambiguous. Am I supposed to have it in my hand? On my head? In my inventory? Left click a tank? Right click?

I think I explained the basics of tank modifiers under the tank modifier, so if you want to reference that to make this a bit clearer it would be better. And honestly, just updating slurping ignores the other 4 or 5 tank modifiers, all should get such info.

roosemberth:
I tried to keep the sentence as small as possible since I have no way of testing (or rather the necessary time) how text overflows in the in-game manual. But I think your remarks are adequate.

I do not want to invest the time of looking at the other modifiers. But I'm fine with closing this MR and opening an issue so that somebody with more time can implement and test a cleaner solution.

PR #4822, opened 2022-04-04
Portuguese Brazillian Translation

Abiel12:
src/main/resources/assets/tconstruct/lang/pt_br.json

https://github.com/Abiel12/TinkersConstruct/blob/1.18/src/main/resources/assets/tconstruct/lang/pt_br.json


Abiel12:
Translation for Tinkers 1.18

KnightMiner:
Why are you adding a .zip file as part of this translation? If you are trying to add books, add the individual files to the correct places. If you are not done yet, mark this PR as a draft.

PR #4841, opened 2022-04-24
TinkerTabsScreen rework

kirderf1:
The primary goal of this and some future PRs is to rework certain gui objects that extend `ModuleScreen`, but are ill-fit to do so. Extending `Screen` or `AbstractContainerScreen` are ill-suited for many module screens, as the interface functions (such as `screen.render()` and `screen.mouseClicked()`) and other functionality (container menu slots) appear to be deliberately avoided.
I therefore want to convert several of these gui objects off `ModuleScreen`, to instead be based on interfaces such as `Widget` and `GuiEventListener`. In particular, I am targeting gui objects that extend `ModuleScren`, and by extension `AbstractContainerScreen`, while also not using any slots.

For this PR, I have done so for `TinkerTabsScreen`, now renamed to `TinkerTabsWidget`.
Effects of this are:
- The class no longer uses any inheritance to function. Because it no longer extends `AbstractContainerScreen` or `Screen`, it will no longer be treated as a screen by other mods. For example, the tabs will no longer have an `InitScreenEvent` posted for it.
- It implements `Widget` and `GuiEventListener` and is thus added to the parent screen as a screen widget instead of as a module screen. `NarratableEntry` is also implemented in order to make the tabs usable with `screen.addRenderableWidget()`, however it is set to not have any narration for now.
- The parent class no longer has its size (as defined by `leftPos` and `imageWidth` etc) expanded to include the tabs that normally happens in `MultiModuleScreen.updateSubmodule()`. With how `MultiModuleScreen.render()` works, this will not affect any render functions. After an usage search, I concluded that it was fine outside of that after manually adding the tabs area in `screen.getModuleAreas()` to be used by jei, and checking for the area in `screen.hasClickedOutside()` (both functions in the parent screen). With this, the area considered outside the screen is now more closely fitted to the screen and its tabs.
- Background and and tooltip rendering is now done in the same `render()` call. This is done to fit everything in the `Widget` interface, and because the rendering order wouldn't differ much by doing so (and more importantly, doesn't cause a problem with the tabs). It should not be difficult at all to update the class if we later decide that we need a "render tooltip" interface for some widgets.
- The return values for mouse function calls have changed, and should now be more in line with vanilla, where "true" is returned for handled mouse input, and "false" for unhandled mouse input.
- The priority order for handling mouse input has now shiften. This should be fine as the tabs doesn't overlap with anything else, and should thus not change which gui component that gets to handle the mouse input.

There are also some additional changes that were made on top of this:
- The tabs widget now get recreated during the `screen.init()` call, instead of created in the constructor and getting its position updated in `screen.init()`. This is more inline with how vanilla widgets are handled, and allows us to make the tabs position immutable (which they now also are).
- `TinkerTabsWidget` has now taken over more responsibilities from `BaseTabbedScreen`. Tabs from the container menu is now prepared in the widget, and tab selection is now fully handled by the widget. Following this, a lot more widget fields are now private and immutable.
- In `BaseTabbedScreen`, remove the container field because the same container menu already exists by the same type in `AbstractContainerScreen`. I also added nullable annotations to the tile field and getter to reflect it also being nullable in the container menu.
- Improved accuracy of the size of the tabs widget to be more narrow along the X-axis. For comparison:
size_before
is now narrowed down to
size_after
(Illustrated with `Screen.fill(poseStack, widget.leftPos, widget.topPos, widget.guiRight(), widget.guiBottom(), 0x77FF0000);`)

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tables/client/inventory/BaseTabbedScreen.java
+++ b/src/main/java/slimeknights/tconstruct/tables/client/inventory/BaseTabbedScreen.java
@@ -3,31 +3,21 @@ package slimeknights.tconstruct.tables.client.inventory;
 import com.mojang.blaze3d.systems.RenderSystem;
 import com.mojang.blaze3d.vertex.PoseStack;
 import net.minecraft.client.renderer.Rect2i;
-import net.minecraft.client.resources.sounds.SimpleSoundInstance;
-import net.minecraft.core.BlockPos;
 import net.minecraft.network.chat.Component;
 import net.minecraft.network.chat.TextComponent;
 import net.minecraft.resources.ResourceLocation;
-import net.minecraft.sounds.SoundEvents;
 import net.minecraft.world.MenuProvider;
 import net.minecraft.world.entity.player.Inventory;
 import net.minecraft.world.inventory.Slot;
-import net.minecraft.world.item.ItemStack;
-import net.minecraft.world.level.Level;
 import net.minecraft.world.level.block.entity.BlockEntity;
-import net.minecraft.world.level.block.state.BlockState;
-import org.apache.commons.lang3.tuple.Pair;
 import slimeknights.mantle.client.screen.ElementScreen;
 import slimeknights.mantle.client.screen.MultiModuleScreen;
 import slimeknights.tconstruct.TConstruct;
-import slimeknights.tconstruct.common.network.TinkerNetwork;
 import slimeknights.tconstruct.library.client.Icons;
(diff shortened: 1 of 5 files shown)


kirderf1:
Rebased onto the 1.18.2 branch without any issues. Still waiting for a review.

KnightMiner:
Did some brief testing and this still seems to work and looks cleaner, we really should migrate towards widgets for most of our stuff. For now though, lets just fix this bug.

PR #4861, opened 2022-05-16
Create uk_ua.json

MrSvyatik:
L like ukraine


ghost:
Зачем перевод русского языка под украинский переделал? Ты своровал перевод у @Bytegm

KnightMiner:
@DrHesperus can you clarify what you mean? I agree there are a lot of similarities (my understanding is they are similar lanaguages), but with my limited knowledge of the lanaguage best i can say is it has a different set of lines translated from both the committed language file and either open Russian PR. In any case, the license of TiC does allow someone to modify the original translation, its all MIT.

@RobloMinerYT can you clarify whether you based this off the Russian translation, and if so which one? Even though the license allows it, it is still good to give credit if you use someone else's work.

MrSvyatik:
так

Externallex:
> так
Ещё не все мои переводы на украинский переделал, фанат?

KnightMiner:
I am going to close this PR due to lack of response to my question above, since I am cleaning up 1.16 PRs.

PR #4882, opened 2022-05-27
Tools place item from inventory

SrYthan:
Tools can place blocks in the world if they are in the last slot in the inventory, just right-click holding any tool to place the block in the last slot in the hotbar
Screenshot_1


KnightMiner:
Okay, couple of things:
* You should not have to change 4000+ files for the change described
* I have absolutely no desire to reimplement the offhand on tools, you have the offhand

As stated on the contribution gudelines, talk to us before making pull requests to prevent wasted effort.

PR #4891, opened 2022-06-03
Fixed minor typos in 2 advancements

Gilberry:


KnightMiner:
I don't accept pull requests for minor typo fixes as it takes me more effort to pull the branch to test than it does to fix it myself. An exception would be if you put in the effort to fix a lot of errors (I can guarantee articles are not the biggest error that exist). I fixed this in dev, so closing the pull request.

PR #4911, opened 2022-06-15
Added id_id.json(re-upload)

ArizTheGamer:
the previous file deletes the en_us file, so i re-do the id_id.json upload.


KnightMiner:
Went ahead and merged this as I would like to get a build out, but wanted to mention for those following translations that only about half of this is done. I figure it is easier to start from a half finished translation than nothing though

PR #4953, opened 2022-07-28
add TinkerToolCraftingEvent

TT432:
For the record, I am not a native English speaker.

I'm making a addon, I want add a `Modifier` to crafted tinker tools, but `PlayerEvent.ItemCraftedEvent#crafting` can't modify, so I write `TinkerToolCraftingEvent`.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/events/TinkerEventFactory.java
+++ b/src/main/java/slimeknights/tconstruct/library/events/TinkerEventFactory.java
@@ -1,0 +1,24 @@
+package slimeknights.tconstruct.library.events;
+
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraftforge.common.MinecraftForge;
+import net.minecraftforge.eventbus.api.IEventBus;
+import org.jetbrains.annotations.Nullable;
+
+/**
+ * @author TT432
+ */
+public class TinkerEventFactory {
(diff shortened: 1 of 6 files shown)


Insane96:
Can't you change the default modifiers of a tool from https://github.com/SlimeKnights/TinkersConstruct/tree/1.18.2/src/generated/resources/data/tconstruct/tinkering/tool_definitions

KnightMiner:
you cannot change default modifiers in a practical way, not without overwriting all the base tools individually

Can you please discuss your goal on discord? Need to decide if this really is the best approach, or if you would be better suited to a different hook (e.g. one called on tool rebuild to add new modifiers)

TT432:
> you cannot change default modifiers in a practical way, not without overwriting all the base tools individually
> 
> Can you please discuss your goal on discord? Need to decide if this really is the best approach, or if you would be better suited to a different hook (e.g. one called on tool rebuild to add new modifiers)

I add new [Origins](https://www.curseforge.com/minecraft/mc-mods/origins-forge)'s [power](https://origins.readthedocs.io/en/latest/json/power/), if player have this power, they crafted tool will be add new `Modifier`, I can't do it by recipe, unless directly replace the original recipe class

KnightMiner:
I have not forgotten about this PR, I just am not convinced a crafting event is the right way to do this, as there are many ways to get modifiers

My plan was to add a way to add "global tool traits" which will be applied like any of the modifiers from JSON on stat rebuild. Will leave this open until I implement such a thing, just cannot promise it will be 1.18

PR #4964, opened 2022-08-06
Create uk_ua.json (original translation)

BetaCarotina:


BetaCarotina:
Should've actually given some description

I had made this translation for a 1.16 server from the original en_us.json file and then updated it to 1.18. It doesn't copy from ru_ru.json and is quite different in terms of translation approach. It also has custom tool name formats for every material, thus evading such nonsense as "Золото Меч"

However, it doesn't include the books (yet?)

PR #4988, opened 2022-09-07
Info panel screen rework and tinker station screen tweaks

kirderf1:
A continuation of the module-screen-to-widget rework that started in  #4841.
For the reason described in the last PR, `InfoPanelScreen` and `PartInfoPanelScreen` had been converted from instances of `ModuleScreen`, to widgets using vanillas system.

There are also some changes to how the info panel is used, to the point that the part builder no longer needed its own extension of the info panel.

Effects of the conversion from a module screen to a widget are akin to last time:
- The class no longer uses any inheritance to function. Because it no longer extends `AbstractContainerScreen` or `Screen`, it will no longer be treated as a screen by other mods. For example, the info panel will no longer have an `InitScreenEvent` posted for it.
- The info panel implements `Widget` and `GuiEventListener` and is thus added to the parent screen as a screen widget instead of as a module screen. `NarratableEntry` is also implemented in order to make it usable with `screen.addRenderableWidget()`, however it is set to not have any narration for now.
- The parent class no longer has its size (as defined by `leftPos` and `imageWidth` etc) expanded to include the the info panels that normally happens in `MultiModuleScreen.updateSubmodule()`. With how `MultiModuleScreen.render()` works, this will not affect any render functions. The size of the info panels are still accounted for jei and when dropping items from the screen.
- The priority order for handling mouse input may now be changed. This should be fine as the info panel doesn't overlap with anything else, and should thus not change which gui component that gets to handle the mouse input.

There are also a number of other cleanup and changes included:
- The info panels now get recreated during the `screen.init()` call, instead of created in the constructor and getting its position updated in `screen.init()`. This is more inline with how vanilla widgets are handled, and allows us to make the position immutable (which it now also is).
- As the info panels will now be recreated when the game window changes its size, mutable info panel data such as the text and the position of the slider, will be kept by getting and setting this data during the `screen.resize()` call.
- Info panels now use a style enum to define texture uv offsets for different styles. The style is now determined by passing an enum to the constructor, instead of by calling `infoPanel.wood()` or `infoPanel.stone()`.
- The info panel can now use a list of captions, which replaces the need for the part builder-specific class. However to keep the setter-style that was used with `PartInfoPanelScreen`, a helper class `CaptionsBuilder` was introduced.
- The info panel has seen some cleanup with numbers and constants.
- A new widget interface has been introduced for widgets that cover an area outside the parent screen. There is also a new base screen class that manages a list for these interfaces. This base class could be expanded in the future to handle additional widget interfaces if we add new ones (for example, if we were to add one for rendering tooltips). Both the base class and the interface might be suited to be moved to mantle in the future.
- The height of the tinker station info panels has been reduced so that they do not stick out below the tinker station screen.
- The tinker station beams have been tweaked to be better suited for the texture change from #4157.
- Since the size of the tinker station screen no longer expands to cover the info panels and inadvertently also cover the info panel beam, the area of both beams are now explicitly accounted for. same goes for the panel decorations.
- A new widget class `SimpleElementWidget` has been added to achieve the previous step in a simple manner.
- The side buttons widget is now also added to the parent screen as a traditional widget. NarratableEntry is implemented in order to make it usable with screen.addRenderableWidget(), however it also is set to not have any narration for now.

For comparison on the info panel height and beam changes, here's the screen before:

and here's the screen after the changes:

After this PR has been merged, I will as a final step also rework `SideButtonsScreen` and `TinkerStationButtonsScreen`.


BeansBeefBroccoli:
Why are guis so complicated?

KnightMiner:
Any update on this? Seems the implementation of the modifier workstation broke this PR.

kirderf1:
I'll have a look at it to resolve the merge conflict and make sure it still works as intended.

KnightMiner:
Might also be worth moving some of these changes back to Mantle in the future, been meaning to deprecate the nested screen method of modules for awhile, would probably be good to fully do it when we port to 1.19.

I'll double check with the other people who use Mantle to see if any of them use Mantle's GUI stuff at all, if not we could probably just port the new classes there and ditch the multimodule stuff.

kirderf1:
Yeah, `ExtraAreaWidget` and `ExtendedContainerScreen` would certainly be candidates of that.

Anyways, this PR should be ready again now. I did some basic testing and noticed no issues with `TinkerStationScreen`, and spotted no problems or changes to `ModifierWorktableScreen` (the latter which had to be updated as it was using info panels).

KnightMiner:
Great, should have more time this weekend to look into this again. I will likely do a release without this change just as I don't have time to really test it, but this will be some good cleanup for the release after.

PR #4990, opened 2022-09-11
Side buttons rework

### LINKED ISSUE #9: Inventory Locks show up twice in Tinkers' Construct GUI
**Describe the bug**
When using certain Tinkers' Construct GUIs the locks will appear more than once (they are doubled):

image

After speaking with the Tinkers' dev, he suspects the locked slots are hardcoded location in all UIs, whether it has slots or not. If you iterate the player inventory slots instead of drawing at a constant location it should work. The GUIs in question are actually using 3 sets of inventory slots: left, middle, and right

**To Reproduce**
1. Open up Tinker's Station or Tinker's Anvil
2. Observe inventory slot locks are now doubled

**Expected behavior**
To show only 1 set of locked slots

**Versions (please complete the following information):**

I am currently using the mod in 1.18.2, but this has been an issue since 1.18.1 so you can use that version to test also.

 - Forge version: 40.0.45
 - Mod Version: 3.1.0

kirderf1:
Related to #4841 and #4988.

Reworks `SideButtonsScreen` to not use `ModuleScreen`, and does some minor additional cleanup.

- The class no longer uses any inheritance to function. Because it no longer extends `AbstractContainerScreen` or `Screen`, it will no longer be treated as a screen by other mods. For example, the side buttons will no longer have an `InitScreenEvent` posted for it. THIS FIXES kirderf1/InventoryFree#9.
- The side buttons widget implements `Widget` and `GuiEventListener` but is not added to the parent screen as a screen widget. Instead, render and mouse input functions are called by the parent class directly.
- The parent class no longer has its size (as defined by `leftPos` and `imageWidth` etc) expanded to include the the side buttons that normally happens in `MultiModuleScreen.updateSubmodule()`. With how `MultiModuleScreen.render()` works, this will not affect any render functions. The size of the widget is still accounted for jei and when dropping items from the screen.
- The side buttons widget now get recreated during the `screen.init()` call, instead of created in the constructor. This is more inline with how vanilla widgets are handled, and allows us to make the position immutable (which has not been done at this time).
- To remove somenot strictly neccessary instanceof checks and casts in `TinkerStationButtonsWidget`, `SideButtonsWidget`was made generic for the button type.
- `TinkerStationButtonsWidget.selected` was replaced by instead checking `TinkerStationScreen.currentLayout`. Doing so removed any need to copy over data when the widget was recreated during a window resize.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
+++ b/src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
@@ -10,7 +10,7 @@ import slimeknights.mantle.client.screen.MultiModuleScreen;
 
 import java.util.List;
 
-public class SideButtonsWidget implements Widget, GuiEventListener {
+public class SideButtonsWidget<T extends Button> implements Widget, GuiEventListener {
 
   protected final MultiModuleScreen<?> parent;
 
@@ -25,7 +25,7 @@ public class SideButtonsWidget implements Widget, GuiEventListener {
   public int yOffset;
 
   private final int columns;
-  protected final List<Widget> renderables = Lists.newArrayList();
+  protected final List<T> buttons = Lists.newArrayList();
   private Button clickedButton;
   protected int buttonCount = 0;
 
@@ -67,7 +67,7 @@ public class SideButtonsWidget implements Widget, GuiEventListener {
     this.topPos += this.yOffset;
   }
 
-  public void addSideButton(Button button) {
+  public void addSideButton(T button) {
     int rows = (this.buttonCount - 1) / this.columns + 1;
 
     this.imageWidth = button.getWidth() * this.columns + this.spacing * (this.columns - 1);
@@ -84,7 +84,7 @@ public class SideButtonsWidget implements Widget, GuiEventListener {
       button.x += parent.imageWidth;
     }
 
-    this.renderables.add(button);
+    this.buttons.add(button);
     this.buttonCount++;
   }
 
@@ -95,13 +95,10 @@ public class SideButtonsWidget implements Widget, GuiEventListener {
 
   public boolean handleMouseClicked(double mouseX, double mouseY, int mouseButton) {
     if (mouseButton == 0) {
-      for (Widget widget : this.renderables) {
-        if (widget instanceof Button button) {
-
-          if (button.mouseClicked(mouseX, mouseY, mouseButton)) {
-            this.clickedButton = button;
-            return true;
-          }
+      for (T button : this.buttons) {
+        if (button.mouseClicked(mouseX, mouseY, mouseButton)) {
+          this.clickedButton = button;
+          return true;
         }
       }
     }
@@ -121,8 +118,8 @@ public class SideButtonsWidget implements Widget, GuiEventListener {
 
   @Override
   public void render(PoseStack matrices, int mouseX, int mouseY, float partialTicks) {
-    for (Widget widget : this.renderables) {
-      widget.render(matrices, mouseX, mouseY, partialTicks);
+    for (T button : this.buttons) {
+      button.render(matrices, mouseX, mouseY, partialTicks);
     }
   }
 }
--- a/src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
+++ b/src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
@@ -12,6 +12,8 @@ import java.util.List;
 
 public class SideButtonsWidget<T extends Button> implements Widget, GuiEventListener {
 
+  public static final int SPACING = 4;
+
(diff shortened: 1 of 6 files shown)


src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
```diff
@@ -0,0 +1,120 @@
...
+  public int imageWidth;
+  public int imageHeight;
+  public int xOffset;
+  public int yOffset;
+
+  private final int columns;
+  protected final List<T> buttons = Lists.newArrayList();
+  private Button clickedButton;
+
+  public int spacing = 4;
```
KnightMiner: While we are refactoring this, I do not like random public non-static variables. This should either be private or static final.

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/TinkerStationButtonsWidget.java
```diff
@@ -0,0 +1,77 @@
...
+        this.addInfoButton(slotButtonItem);
+        if (layout == parent.getCurrentLayout()) {
+          slotButtonItem.pressed = true;
+        }
+        index++;
+      }
+    }
+  }
+
+  public void addInfoButton(SlotButtonItem slotButtonItem) {
```
KnightMiner: While we are refactoring, do these functions need to be public? I find its best to use a minimal access level when possible. In general check if any of the functions in renamed classes can be given a lower access level (ignoring package, I tend to stick with public, private, and protected)

Would be nice to get javadocs on these if they need to be public, but I will not require that for the PR since they existed before.

KnightMiner (commented):
Just a couple of minor things until I have a chance to test this in game.

src/main/java/slimeknights/tconstruct/tables/client/inventory/widget/TinkerStationButtonsWidget.java
```diff
@@ -0,0 +1,75 @@
...
+      this.addInfoButton(slotButtonItem, style);
+      if (layout == parent.getCurrentLayout()) {
+        slotButtonItem.pressed = true;
+      }
+    }
+
+    this.setButtonPositions();
+  }
+
+  public static List<StationSlotLayout> createLayoutsList(TinkerStationScreen parent) {
```
kirderf1: Creating this list might be better suited to be done inside `TinkerStationScreen`. Thoughts?

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
```diff
@@ -12,33 +12,28 @@
...
-  public int leftPos;
-  public int topPos;
-  public int imageWidth;
-  public int imageHeight;
-  public int xOffset;
-  public int yOffset;
+  public final int leftPos;
+  public final int topPos;
+  public final int imageWidth;
+  public final int imageHeight;
```
KnightMiner: the note about making things private applied to these too. I do not like public non-static fields on things unless needed (which they rarely are)

Looks like your commit already made these so they are set through the constructor, so should be trivial to make them private.

src/main/java/slimeknights/tconstruct/tables/client/inventory/widget/TinkerStationButtonsWidget.java
KnightMiner: given it takes the screen as a parameter, yes

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
```diff
@@ -12,33 +12,28 @@
 
 public class SideButtonsWidget<T extends Button> implements Widget, GuiEventListener {
 
-  protected final MultiModuleScreen<?> parent;
+  public static final int SPACING = 4;
```
KnightMiner: Is this used outside the file at all? I have a suspicion the intention was to allow changing the spacing on a per widget basis, but we are not currently using that feature hence making it static. On the chance we need it in the future, may be better to make it private/protected to make a refactor later lower risk.

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
kirderf1: Sure. I left them public as they were being used for rendering the buttons beam in `TinkerStationScreen`, which was reworked in #4988 by f62c1f03e0a4d56041f36f8923f34a81e43e8cc0. I'll make these fields private with getters for those fields that were being used.

src/main/java/slimeknights/tconstruct/tables/client/inventory/module/SideButtonsWidget.java
kirderf1: Sounds reasonable.

PR #5029, opened 2022-11-26
DispenserItemBehavior for flint and brick.

Waterpicker:


KnightMiner:
General comments, will leave specific code comments later

* I cannot get the flint and brick to light fires at all, with no modifiers it should work identically to flint and steel. Use whatever logic it has to determine side hit for other modifier logic
* Glowing, pathing, stripping, and tilling seem to work as expected. Turns out pathing is the one that had unique behavior, but that is handled automatically with how you wrote the code
* Regarding expanders, it seems for firestarter expanded and fireprimer act identically, so together they can give a radius 2 circle. On anything else (tilling, stripping, pathing, glowing) it currently does nothing as its technically impossible to get both. Not sure we need to support both, if you wish to you could always hardcode it to circle.

src/main/java/slimeknights/tconstruct/library/modifiers/Modifier.java
```diff
@@ -969,6 +970,18 @@ public void onEquip(IToolStackView tool, int level, EquipmentChangeContext conte
...
+  /**
+   * Called when a stack is being used within a dispenser.
+   * <br>
+   * @param tool      Tool instance
+   * @param level     Modifier level
+   * @param source    Block source
+   * @param stack     Stack being dispensed
+   * @return True if dispenser action was taken
+   */
+  public boolean onDispenserUse(IToolStackView tool, int level, BlockSource source, ItemStack stack) {
```
KnightMiner, line 982: Do we need the item stack parameter? I generally leave it out of modifier hooks as people should be using the tool stack.

Looks like you use it to create a use on context below if I am understanding the code, double check if that is avoidable, if not modify the javadoc to be clear about the relationship between the tool and the stack and that the stack should not be used to create a `ToolStack` instance.

src/main/java/slimeknights/tconstruct/library/tools/definition/aoe/CircleAOEIterator.java
```diff
@@ -67,11 +83,27 @@ public static Iterable<BlockPos> calculate(IToolStackView tool, ItemStack stack,
...
     return () -> new CircleIterator(origin, directions.width(), directions.height(), directions.traverseDown(), directions.depth(), radiusSq, diameter / 2, is3D, posPredicate);
   }
 
+  private static ExpansionDirections getDirections(Direction harvestDirection, Direction sideHit) {
+    // depth is always direction into the block
+    Direction depth = sideHit.getOpposite();
+    Direction width, height;
+    // for Y, direction is based on facing
+    if (sideHit.getAxis() == Direction.Axis.Y) {
+      height = harvestDirection;
```
KnightMiner, line 97: To be honest, the harvest direction parameter does not matter for circles. It matters for a rectangle as it determines the rectangle orientation and was conveient to reuse the rectangle logic for a circle, but if you are going to extract it I would just ditch the side/player argument entirely (could be accomplished by making player nullable in the above method signature then ignoring it here)

src/main/java/slimeknights/tconstruct/library/tools/item/ModifiableItem.java
```diff
@@ -585,4 +588,8 @@ public boolean shouldCauseReequipAnimation(ItemStack oldStack, ItemStack newStac
   public static BlockHitResult blockRayTrace(Level worldIn, Player player, ClipContext.Fluid fluidMode) {
     return Item.getPlayerPOVHitResult(worldIn, player, fluidMode);
   }
+
+  public static UseOnContext contextFromBlockSource(BlockSource source, ItemStack stack) {
```
KnightMiner, line 592: Needs javadoc, and probably belongs in `ModifierUtils` or one of those similar helper classes. Above method is only here as `getPlayerPOVHitResult` is protected.

src/main/java/slimeknights/tconstruct/tools/TinkerTools.java
```diff
@@ -146,6 +153,21 @@ void commonSetup(FMLCommonSetupEvent event) {
...
+        ToolStack tool = ToolStack.from(stack);
+
+        for (ModifierEntry entry : tool.getModifierList()) {
+          if (entry.getModifier().onDispenserUse(tool, entry.getLevel(), source, stack)) break;
+        }
+
+        return stack;
+      };
+
+      DispenserBlock.registerBehavior(TinkerTools.flintAndBrick, behavior);
```
KnightMiner, line 168: Flint and steel fires particles if its unable to do anything. Looks like they use a class called `OptionalDispenseItemBehavior` for that fallback particle, I suggest using that

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/interaction/BlockTransformModifier.java
```diff
@@ -127,6 +129,80 @@ public InteractionResult afterBlockUse(IToolStackView tool, int level, UseOnCont
...
+
+      // if the tool breaks or it was a campfire, we are done
+      if (ToolDamageUtil.damage(tool, 1, null, null)) {
+        return true;
+      }
+    }
+
+    // AOE transforming, run even if we did not transform the center
+    // note we consider anything effective, as hoes are not effective on all tillable blocks
+//    if (player != null && !tool.isBroken()) {
```
KnightMiner, line 167: Lets just hardcode to circle AOE here, using the same logic you were planning to use in fireprimer.

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/interaction/FirestarterModifier.java
```diff
@@ -161,4 +165,64 @@ public InteractionResult afterBlockUse(IToolStackView tool, int level, UseOnCont
     }
     return didIgnite ? InteractionResult.sidedSuccess(world.isClientSide) : InteractionResult.PASS;
   }
+
+  /** Ignites the given block */
+  private static boolean igniteDispenser(IToolStackView tool, Level world, BlockPos pos, BlockState state, Direction sideHit, Direction horizontalFacing) {
```
KnightMiner, line 170: Why did you duplicate this whole method just to remove a nullable player argument? Just call the original method by passing null.

If there are any other differences, add the needed parameter to distinguish the two

KnightMiner (changes requested):

PR #5043, opened 2022-12-23
Implement block transform modifier hook

Pyre540:
This PR adds new hook that allows modifiers to hook into the block transform modifiers.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tools/modifiers/ability/interaction/BlockTransformModifier.java
+++ b/src/main/java/slimeknights/tconstruct/tools/modifiers/ability/interaction/BlockTransformModifier.java
@@ -18,6 +18,7 @@ import net.minecraft.world.level.material.Material;
 import net.minecraftforge.common.ToolAction;
 import slimeknights.tconstruct.library.modifiers.ModifierEntry;
 import slimeknights.tconstruct.library.modifiers.TinkerHooks;
+import slimeknights.tconstruct.library.modifiers.hook.BlockTransformModifierHook;
 import slimeknights.tconstruct.library.modifiers.hook.interaction.BlockInteractionModifierHook;
 import slimeknights.tconstruct.library.modifiers.hook.interaction.InteractionSource;
 import slimeknights.tconstruct.library.modifiers.impl.InteractionModifier;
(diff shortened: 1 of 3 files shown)


src/main/java/slimeknights/tconstruct/tools/modifiers/ability/interaction/BlockTransformModifier.java
```diff
@@ -139,6 +143,13 @@ public InteractionResult afterBlockUse(IToolStackView tool, ModifierEntry modifi
     return didTransform ? InteractionResult.sidedSuccess(world.isClientSide) : InteractionResult.PASS;
   }
 
+  /** Runs the hook after transforming a block */
+  private void runTransformHook(IToolStackView tool, UseOnContext context, BlockState state, BlockPos pos) {
```
KnightMiner: Move this to a static method in the interface. Some of the newer hooks do similar things

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/interaction/BlockTransformModifier.java
```diff
@@ -121,6 +124,8 @@ public InteractionResult afterBlockUse(IToolStackView tool, ModifierEntry modifi
             totalTransformed++;
             didTransform = true;
 
+            BlockTransformModifierHook.afterTransformBlock(tool, context, newTarget, newPos, action);
```
KnightMiner, line 131: This placement is inconsistent. The above call to the hook runs serverside only, but this call potentially could run on the client if the targeted block did not transform. Could lead to really weird bugs with the code consuming this hook that are really hard to reproduce.

Just move the call to after the client side hook check. I am undecided on if you should damage the tool before or after this hook runs, check what I did for other hooks like harvest and make it consistent with them (and of course, internally consistent in this modifier)

Pyre540:
It seems that there is a small inconsistency in other modifiers as well. In `HarvestAbilityModifier` you run hooks before tool damage and in `ShearsAbilityModifier` the other way around. Other hooks that I looked run before damage.

Intuition tells me that hooks should run for the same state as the tool when it was used, so before the tool damage. But that probably depends on what hooks are doing. 

For now I moved the hook call before damage, but I am happy to tweak it further if neccesary.

PR #5050, opened 2023-01-03
Update Japanese translation(1.18)

koh-gh:
- Updated ja_jp.json
- Added Japanese book translation


karakufire:
I think it's better to use `x`(latin alphabet ex) or `×`(multiplication sign) than asterisk, `*`, for multiplying factor, in book's text or other description. 
Because using the asterisk outside of programing context, especially in texts, can lead us to read them other meanings, for example annotation.

本の文章や説明などでは、乗算を表すのには`*`より`x`(エックス)、あるいは`×`(乗算記号)を使うほうが良いと思います。
アスタリスクだと、プログラミングの文脈以外だと注釈の存在など別に意味に誤解されるおそれが高いからです。

koh-gh:
Sorry for Japanese.
この翻訳でアスタリスクを使っているのはEncyclopediaだけで、文脈的に掛け算以外と誤解される可能性は低いと思いますが、ひとまず「×」に変えて見やすくなるかを確認してみます。「×」は2バイト幅になるので、敢えて使っていませんでした。

KnightMiner:
Sorry for lack of reply, translations tend to be a lot of effort to review as I do not speak the language.

Planning to release an update in the next few days. There will be an additional update in the near future to finish some slimestaff features later, so let me know if you think this PR is in a state that it can be merged, or if you'd rather wait for that next release.

koh-gh:
Thank you for your reply. Current translation does not include the translation about shield and slimestaff, but I have translated the previous ones. So even if the translation is not updated, please merge it. Of course, if I have time, I will also update translation.

The above discussion is trivial, and there is no big problem in translation.

koh-gh:
Updated to latest. Regarding the above discussion, I have also heard the opinions of other Japanese players.
Additional updates will be made if necessary.

koh-gh:
I have changed the multiplication sign as a result of referring to the opinions of Japanese players.
The discussion above is complete. Is it OK? @karakufire

PR #5073, opened 2023-01-28
Backport unbreakable change

Ethryan:
Backported the Reinforcement Modifier item from 1.12.2, so you don't have to use an Obsidian Plate Anymore.
image

image


KnightMiner:
I have no plans to change how reinforcements work. They are meant to encourage you to look into automating smeltery recipes, and iron is way nicer than gold.

That said, nothing stops you from changing this via datapacks.

----

Now, its hard to tell as this PR is mostly nonsense, but if you were trying to change 1.7.10, that version is long, long, long, long, long dead. We will never release another build for 1.7.10.

Ethryan:
This was a misclick, it was supposed to affect the GTNH fork, but i missed to change the master.

I closed it 1 min after i clicked to open the pull request.

PR #5077, opened 2023-02-02
Skyslime Sling fixes

Tictim:
While debugging an odd interaction between TCon and my own mod, I discovered a possible desync state caused by Skyslime Sling in relation to the player's sprinting flag. This PR addresses the issue, while also fixing some of the oddities in the same file.

Skyslime Sling makes player sprint by calling `player.setSprinting(true)` on both client and server. This approach has possible point of failure, in which client side flag is ignored and only server side flag is set. This creates a state where server thinks player is sprinting but the client is not. This state has little to no effect on most of gameplay, but it caused some issues when it's coupled together with my mod.

This desync state can be prevented by setting the flag on client side only. It also keeps the auto-sprint feature unchanged.

Additionally there was a second `onSuccess()` call which made the Sling lose durability two times per use and play sound twice. I just removed the first call, since that one looked like it's put in by a mistake. Pardon me if this change was inappropriate.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/SkySlimeSlingItem.java
+++ b/src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/SkySlimeSlingItem.java
@@ -45,7 +45,9 @@ public class SkySlimeSlingItem extends BaseSlimeSlingItem {
     }
 
     player.causeFoodExhaustion(0.2F);
-    player.setSprinting(true);
+    if (worldIn.isClientSide) {
+      player.setSprinting(true);
+    }
 
     float f = getForce(stack, timeLeft);
     float speed = f / 3F;
--- a/src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/SkySlimeSlingItem.java
+++ b/src/main/java/slimeknights/tconstruct/gadgets/item/slimesling/SkySlimeSlingItem.java
@@ -52,7 +52,6 @@ public class SkySlimeSlingItem extends BaseSlimeSlingItem {
       (1 + look.y) * speed / 2f,
       (look.z * speed));
 
-    onSuccess(player, stack);
     SlimeBounceHandler.addBounceHandler(player);
     if (!worldIn.isClientSide) {
       player.getCooldowns().addCooldown(stack.getItem(), 3);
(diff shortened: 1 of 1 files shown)


KnightMiner:
I need to take a look as to whether we actually need to call that or not, it seems a little odd to manually set the sprinting state.

What is the case where the server would get it called and the client would not? The tool always executes on both sides.

Tictim:
The issue comes from the difference of syncing method for `isSprinting` flag, on client-to-server and server-to-client respectively.

On client-to-server sync, packets are used. Each tick, LocalPlayer re-evaluates `isSprinting` flag, then compares the value with `wasSprinting` field. If it's different value, a packet is sent and `wasSprinting` is set to new `isSprinting` value. Server-to-client sync is handled by `SynchedEntityData`.

On client side, due to processing order inside tick method, `SkySlimeSlingItem#releaseUsing` runs before `isSprinting` flag re-evaluation. This creates a corner case that goes like this:
* If `wasSprinting` is `false`,
* and `isSprinting` flag is re-evaluated to `false` (ex. while standing still),
* then the previous value `true` set by `SkySlimeSlingItem#releaseUsing` is ignored - without impacting any state or even sending a packet.

In this case, calling `player.setSprinting(true)` on server-side creates the desync state. Since server relies on the client for checking sprinting state, server-side player entities can't un-set `isSprinting` flag by themselves, until a new value is written from outside. Additionally, the flag is supposed to be synced by SynchedEntityData when `player.setSprinting(true)` is called on server-side (atleast I think so), but the synced value is probably ignored too, in similar fashion as described above.

Calling `player.setSprinting(true)` only on client-side solves this issue because:
* If they can sprint, they will start sprinting, and the flag change will subsequently sent to the server.
* If they _cannot_ sprint, they won't, and the server wouldn't incorrectly assume the player successfully started sprinting.

I will link a [drop-in test code](https://gist.github.com/Tictim/935c5e3ed71e0d3caad8e3616adf0247) I used to find the issue. running `/tcontest` logs `isSprinting` flag for both client-side player and server-side player.

KnightMiner:
The new modifiers added to 1.18 include this fix, and the old slings are deleted in 1.19, so going to go ahead and close this.

PR #5094, opened 2023-02-23
Added REPLACEABLE AOEMatchType

Insane96:
I'm making a [modifier to place blocks in the air](https://discord.com/channels/381483562576379907/395801350731989003/1078284200983134238), and [Diyo wanted the placing area to match the tools AOE](https://discord.com/channels/381483562576379907/395801350731989003/1078287966599921664).

Currently this is not possible due to `AOEMatchType` fixed to harvestable or non-air blocks (`BREAKING` or `TRANSFORM`), and for the mid air placing I need it to match replaceable blocks.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/tools/definition/aoe/IAreaOfEffectIterator.java
+++ b/src/main/java/slimeknights/tconstruct/library/tools/definition/aoe/IAreaOfEffectIterator.java
@@ -76,7 +76,11 @@ public interface IAreaOfEffectIterator extends IHaveLoader<IAreaOfEffectIterator
         }
         return false;
       };
-    } else {
+    }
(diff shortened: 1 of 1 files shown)


KnightMiner:
I feel like hardcoding another match type is the wrong solution. Adding a way for addons to add their own match types is a better approach. I dont believe that file you patched is the only place making use of the match type though.

Look at the modifier hooks for inspiration on how to handle match types, seems like they need a function that maps `IToolStackView tool, ItemStack stack, Level world, BlockPos origin` to a block position predicate. Might need another function for the other uses of that enum.

KnightMiner:
Going to go ahead and close this as last time I talked to Insane they could not recall the reason for this PR.

PR #5114, opened 2023-03-20
Add Falling Anvils

ferriarnus:
Adds the ability to fall to tinkers anvils. This is done by using a slightly modified version of the vanilla falling entity, to make sure that the anvil is rendered correctly and can store its data before falling. Please let me know if you want any changes or clarify something.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/tables/TableClientEvents.java
+++ b/src/main/java/slimeknights/tconstruct/tables/TableClientEvents.java
@@ -18,6 +18,7 @@ import slimeknights.tconstruct.common.ClientEventBase;
 import slimeknights.tconstruct.library.client.model.block.TableModel;
 import slimeknights.tconstruct.shared.block.entity.TableBlockEntity;
 import slimeknights.tconstruct.tables.block.entity.chest.TinkersChestBlockEntity;
+import slimeknights.tconstruct.tables.client.FallingAnvilEntityRenderer;
 import slimeknights.tconstruct.tables.client.TableTileEntityRenderer;
 import slimeknights.tconstruct.tables.client.inventory.CraftingStationScreen;
 import slimeknights.tconstruct.tables.client.inventory.ModifierWorktableScreen;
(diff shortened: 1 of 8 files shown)


KnightMiner:
For the record, the reason I said to talk to me first was so I could tell you if putting in effort to code something is a good idea. The goal was not to alert me you were doing it, the goal was to discuss how it should be done and whether the idea still fits.

For this feature, there are two noteworthy things that need to be dealt with before it would be accepted.

### World breaking

It is possible someone has an anvil placed in a now illegal position, potentially leading to it falling to an area where they cannot retrive it. Skyblock packs would be the most likely cause of this. Best way to avoid them losing their anvil would be to hold off on this feature until 1.19 likely.

### Gameplay considerations

The reason I wanted anvils to fall is because vanilla anvils fall. However, vanilla anvils also take damage when they fall, suggesting ours should. However, taking damage only makes sense when falling if its not the only source of taking damage. However, having it take damage on usage would just be annoying, so not sure I want that at all.

Most of the reason I never got around to this feature is I never could decide how I wanted to handle that, it did not feel right simply making it a weaker vanilla anvil, but fully re-creating the vanilla feature also does not quite fit. Its worth discussing on Discord at a time we are both online.

ferriarnus:
Understandable. The reason I started before waiting on confirmation was because I wanted to firstly see if I *could* fix it. This is something I do in general as not to waste time/promise things I can't do to to project owner. This time I was able to finish it without too many issues, causing me to make this pr before you were able to respond. For the next time, I will take it a bit slower and wait.

I hadn't considered what kind of effect this could have on existing worlds like skyblock. Besides waiting on 1.19 like you suggested, the only other option I can see would be to lock the behavior behind a config for 1.18, and remove/deprecate in the future. Though personally either options here are valid, falling anvils isn't really an important change that needs to be in 1.18.

I see that damaging the anvil would be annoying and not properly fit with tinkers. I'll also think about some possible solutions discus these in the discord.

PR #5117, opened 2023-03-20
multi Level Cache

ferriarnus:
Adds a second level cache to the melting inventory. 
The constructor of a melting module now takes the parent inventory instead of the block entity. If the Block entity is needed, it can be gotten from the inventory.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MeltingModule.java
+++ b/src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MeltingModule.java
@@ -10,12 +10,10 @@ import net.minecraftforge.items.ItemHandlerHelper;
 import slimeknights.mantle.block.entity.MantleBlockEntity;
 import slimeknights.tconstruct.common.network.InventorySlotSyncPacket;
 import slimeknights.tconstruct.common.network.TinkerNetwork;
-import slimeknights.tconstruct.library.recipe.TinkerRecipeTypes;
 import slimeknights.tconstruct.library.recipe.melting.IMeltingContainer;
 import slimeknights.tconstruct.library.recipe.melting.IMeltingRecipe;
 
 import javax.annotation.Nullable;
-import java.util.Optional;
 import java.util.function.Predicate;
 
 /**
@@ -33,7 +31,9 @@ public class MeltingModule implements IMeltingContainer, ContainerData {
   private static final int REQUIRED_TEMP = 2;
 
   /** Tile entity containing this melting module */
-  private final MeltingModuleInventory parent;
+  private final MantleBlockEntity be;
+  /** Melting Inventory containing this melting module */
+  private final MeltingModuleInventory inventory;
   /** Function that accepts fluid output from this module */
   private final Predicate<IMeltingRecipe> outputFunction;
   /** Function that boosts the ores based on the rate type */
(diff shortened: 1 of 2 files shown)


src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MeltingModule.java
```diff
@@ -165,20 +165,30 @@ public void coolItem() {
...
 
     // first, try last recipe for the slot
     IMeltingRecipe last = lastRecipe;
     if (last != null && last.matches(this, world)) {
+      parent.setLastInventoryRecipe(lastRecipe);
       return last;
     }
+
+    // second, try the last recipe from the inventory
+    last = parent.getLastInventoryRecipe();
```
KnightMiner: Firstly, this is wrong. The cached last recipe should ***only*** be changed if we find a new recipe, replacing it with the parent's cache makes this cache far less useful.

Secondly, this chunk of lines is really the responsibility of the parent, logic should look like
```
if last != null and last.matches(this)
  return last
IMeltingRecipe newRecipe = parent.findRecipe(this);
if newRecipe != null
  last = newRecipe
  return newRecipe
return null
```

We could pass in `last` to `parent.findRecipe(this)` to skip running matches on the same recipe twice if we wish, but not strictly needed.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MeltingModule.java
```diff
@@ -78,9 +78,9 @@ private void resetRecipe() {
    */
   public void setStack(ItemStack newStack) {
     // send a slot update to the client when items change, so we can update the TESR
-    Level world = parent.getLevel();
+    Level world = parent.getParent().getLevel();
```
KnightMiner: This looks messy enough that it might be worth keeping "both parents". I would probably call one of them `be` and the other `parent`, or alternatively `be` and `inventory`.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MeltingModuleInventory.java
```diff
@@ -127,6 +126,25 @@ public int getRequiredTemp(int slot) {
...
+  @Nullable
+  public IMeltingRecipe findRecipe(MeltingModule module, @Nullable IMeltingRecipe slotRecipe) {
+    Level world = parent.getLevel();
+    if (world == null) {
+      return null;
+    }
+    if (lastInventoryRecipe != null && slotRecipe != lastInventoryRecipe && lastInventoryRecipe.matches(module, world)) {
+        return lastInventoryRecipe;
+    }
+    Optional<IMeltingRecipe> newRecipe = world.getRecipeManager().getRecipeFor(TinkerRecipeTypes.MELTING.get(), module, world);
```
KnightMiner, line 144: the whole point of having a cache is to save the recipe in the cache when found, you should do so. I said on discord, this logic will look nearly identical to the old logic, old logic modified `lastRecipe`

src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MeltingModuleInventory.java
```diff
@@ -29,11 +31,8 @@ public class MeltingModuleInventory implements IItemHandlerModifiable {
   private static final String TAG_SIZE = "size";
 
   /** Last used recipe by any slot in the inventory */
-  @Getter
-  @Setter
   private IMeltingRecipe lastInventoryRecipe;
```
KnightMiner: this field can just be `lastRecipe`, no need for such a long name.

ferriarnus:
So testing these changes with a debugger:

1ste slot iron -> Recipe manager look up
1ste slot iron -> Slot cache
2nd slot iron -> Inventory cache
3rd slot gold -> Recipe manger look up
1ste slot gold -> Inventory cache
2nd slot iron -> Slot cache
3rd slot iron -> Recipe look up / 4rd slot gold -> Inventory cache

Looking at these last 2 interactions, for the 2d slot, the slot cache can be used. However it does not set the inventory cache when used, making it so a look up is used later. Should this be changed so using the slot cache also updates the inventory cache? Or am I missing a case? The advantage of this setup is that if instead of the last iron, gold was used in slot 4, the inventory cache would resolve it.

KnightMiner:
I am not sure the slot cache should override the inventory cache on a cache hit, if they differ, odds are the next item in the smeltery will also be different. Probably best to stick with what we have, as that is simplier.

The important thing is that when finding a new recipe, it is cached at both the slot and inventory level. Additionally, when a slot has a cache miss, it should cache the inventory recipe even if that was a cache hit. Both of these appear to be done already

PR #5127, opened 2023-04-14
Giving Seared Blocks a Unique look & Few Tweaks

LopyLuna:
This should make it look like its not Dark version of the Stone counterpart

- Unique look for the Seared Blocks
- Mossy variants of the Bricks & Cobble
- Tweaked color palette of the Copper on the Controllers & IOs
- Tweaked gold trim on the Cast Chest & Tinker Anvil
- Animation for Heater & Melter
- Bottom assets for Tank

image
image
image
image
image
### -
image


KnightMiner:
I am discussing with the relevant team members about whether we want to use these textures. To help with the discussion, can you do some side-by-side screenshots with blackstone and deepslate? Its important seared blocks remain distinct from those. I see a bit of that in one of the screenshots, but its not obvious which is which.

Also, from the comparison screenshot, I should say I disagree with the change of the 2x2 tiles texture to a noisier clone of the 1x1 tile texture. The tiles are designed as a more uniform brick texture, so it should still look like tiles. I would be fine with it being 3x3 or something similar, 1x1 is just too close to an existing variant.

On a similar note, I don't believe we used the "chiseled stone brick" like texture anymore since I trimmed down the variants. I would recommend verifying that everything you retextured is actually used in game.

RCXcrafter:
KnightMiner asked me to go over this so I checked out the textures ingame, here are my notes:
* The seared stone has a lot of contrast, I'm not a fan of that but that's just my opinion
* Seared road is not currently in use, neither is chiseled
* The seared ingot tank and fuel tank have missing textures for their bottom face
* I like the new colors for the copper on the smeltery parts
* I'd like the triangle bricks to have a more connected texture but I don't think those were changed, alternitavely they could be replaced with a different texture but I'll leave that to KnightMiner
* I'm not sure about the new brick shape, it does help distinguish the bricks from deepslate bricks though
* The cast chest model really needs to be changed and the new texture makes that even more obvious, this is somewhat involved though since the part chest and tinkers chest also use this model
* Overall I do like the new textures

LopyLuna:
Thanks for the opinions ill be tweaking & fixing majority of issues!
Screenshot will be included after changes & including side by side blackstone & deepslate!

- I went back to check this out yea & I do agree that it haves alot of contrast now
- will be removing the unnecessary textures
- will be fixing the tanks, I prob forgot the proper textures for the model :P

KnightMiner:
Sorry for leaving this PR for so long, did not do much development in 2023. If you are still interested in this, please make the changes we talked about above, and target 1.19 (shouldn't be too difficult since smeltery is not notably changed in 1.19).

LopyLuna:
ah yeah i might just redo the unique textures i kinda forgot about this :P
ill post the redo textures in the discord

LopyLuna:
im gonna make a new pull-request after its fully finished & gained feedback from the [discord](https://discord.com/channels/381483562576379907/719056206068449289)

PR #5156, opened 2023-08-22
Update zh_tw.lang

Izaria138:
Original file: "item.tconstruct.cast_custom.plate.name=大板鑄模".
But it should be "板鑄模" (Plate Cast). 
It's the cast of "板" (Plate), not the cast of "大板" (Large Plate).
"這把工具已有刻紋，它只能刻一次" sounds strange, it's unusual and can be replaced by "這把工具已有刻紋，它只能被雕刻一次" or, more directly, "這把工具只能被雕刻一次".


ShinyPorygon:
1.12 is no longer receiving updates.

PR #5173, opened 2023-12-12
Updated build.gradle to 1.19.2

LMB01:
This is all I am able to do to update


KnightMiner:
Yeah, that is the easiest part of updating the mod. I have not done it yet because I don't like making a branch that does not compile and is not being worked on, and doing that small chunk will not compile.

PR #5092, opened 2023-02-21
Chinese Translation for 1.18.2

ChuijkYahus:

PR #5191, opened 2024-03-04
Simplified Chinese Translation Typo Fix (1.18.2)

Cactusstudent:
As the PR title says, this is a PR focusing on the Typo Fixing. Include the file name, text and so on. This PR will update when I or the Chinese community find some typo. It maybe spend a long time. Hope this PR can be merged in the next **Release version**.

TODO List
- [ ] ~Optimize the text in each books~ Plan to do that in next couple of weeks, so it will not be contained in this PR
- [x] Fix Typo in zh_cn
- [x] Update based on the latest en_us


KnightMiner:
I have some more typo fixing to do in English which may impact you, notably I've been switch out the term "projectile damage" for "projectile power" as not all uses of the stst are strictly damage. It will be mostly encyclopedia that uses it. Not sure what terms you use, I'd make sure they match the translation for the tool stat in the main lang file

Cactusstudent:
Thanks for warning. Luckily, I have noticed this in the #5092 , and I changed the original term "`弹射物伤害`" to “`弹射物力量`” (additionally, `伤害` means **damage** in English, `力量` means **power** in English) to match the English term's change. But only in part of the zh_cn.json. The translation of guide books may still have tons of incorrect text. But I dont know if all of these should be changed, looks like some of the term use "伤害" is suitable, another seems more suitable to use "力量" (Such as the slings modifier). Maybe I do need a standard from you.

KnightMiner:
> Thanks for warning. Luckily, I have noticed this in the #5092 , and I changed the original term "`弹射物伤害`" to “`弹射物力量`” (additionally, `伤害` means **damage** in English, `力量` means **power** in English) to match the English term's change. But only in part of the zh_cn.json. The translation of guide books may still have tons of incorrect text. But I dont know if all of these should be changed, looks like some of the term use "伤害" is suitable, another seems more suitable to use "力量" (Such as the slings modifier). Maybe I do need a standard from you.

In updating the english text, everywhere we used "projectile damage" as a tool stat was more suited to "projectile power". The one case I saw of "projectile damage" that did not get changed was the description of projectile protection, as thats not the tool stat there

PR #5226, opened 2024-05-14
Resprited reinforcements

Zen9890:
This PR adds resprited versions of new reinforcements, because people in discord noted that specifically iron and gold variants dont look good


RCXcrafter:
I'm not sure about the edit for the normal pattern. I don't think it's really needed to make sense, the darker outline could just come from material being cast onto the pattern.
The retextured reinforcements look good though, the border helps to make the icon stand out.

KnightMiner:
If the edit to the normal pattern is the base for the new reinforcements, it could always just be saved in the reinforcements folder as the base texture (sorta like how the blank cast is used) then leaving the original pattern unchanged.

KnightMiner:
> If the edit to the normal pattern is the base for the new reinforcements, it could always just be saved in the reinforcements folder as the base texture (sorta like how the blank cast is used) then leaving the original pattern unchanged.

Please revert the pattern to the original texture, per this comment.

KnightMiner:
Given the lack of response, I went ahead and cherry picked the commit with the new reinforcements, but left the commit with the new pattern out.

PR #5236, opened 2024-06-09
Port protection modifiers to json.

Waterpicker:


KnightMiner:
None of your modifier JSON include the secondary effects of the modifiers you ported to JSON. As I said on discord, those secondary effects need to be ported to new modules.

The code you wrote will also fail on world load. The dynamic modifier filter will never match, as the JSON modifier will always be composable. Just filter to modifier

src/generated/resources/data/tconstruct/tinkering/modifiers/melee_protection.json
```diff
@@ -0,0 +1,25 @@
+{
+  "type": "tconstruct:composable",
+  "level_display": "tconstruct:default",
+  "modules": [
+    {
+      "type": "tconstruct:max_armor_stat",
+      "allow_broken": false,
+      "each_level": 0.005,
+      "key": "tconstruct:use_item_speed",
+      "key2": "tconstruct:melee_protection"
```
KnightMiner: These are not good names for the fields. Its very unclear in the JSON what either does

src/main/java/slimeknights/tconstruct/library/data/tinkering/AbstractModifierProvider.java
```diff
@@ -122,7 +122,11 @@ public void run(CachedOutput cache) throws IOException {
   private static JsonObject serializeModifier(@Nullable Modifier modifier, @Nullable ICondition condition, JsonRedirect[] redirects) {
     JsonObject json;
     if (modifier != null) {
-      json = ModifierManager.MODIFIER_LOADERS.serialize(modifier).getAsJsonObject();
+      try {
+        json = ModifierManager.MODIFIER_LOADERS.serialize(modifier).getAsJsonObject();
+      } catch (Exception e) {
+        throw e;
+      }
```
KnightMiner: What is the reason for this change? This block does nothing effectively

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,119 @@
...
+import slimeknights.tconstruct.library.tools.capability.TinkerDataCapability.TinkerDataKey;
+import slimeknights.tconstruct.library.tools.capability.TinkerDataKeys;
+import slimeknights.tconstruct.library.tools.context.EquipmentChangeContext;
+import slimeknights.tconstruct.library.tools.nbt.IToolStackView;
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+/**
+ * Module for keeping track of the total level of a modifier across all pieces of equipment. Does not support incremental, use {@link ArmorStatModule} for that.
```
KnightMiner: This doc should be updated for max armor level

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /* Helpers */
+
+  /**
+   * Adds levels to the given key in entity modifier data for an armor modifier
+   * @param context  Equipment change context
+   * @param key      Key to modify
+   * @param amount   Amount to add
+   */
+  public static void findMaxLevel(EquipmentChangeContext context, TinkerDataKey<Integer> key, int amount) {
```
KnightMiner: `updateMaxLevel` seems more accurate to what this does

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,119 @@
...
+
+  /**
+   * Gets the total level from the key in the entity modifier data
+   * @param cap    Capability instance
+   * @param key    Key to get
+   * @return  Level from the key
+   */
+  public static int getLevel(LazyOptional<TinkerDataCapability.Holder> cap, TinkerDataKey<Integer> key) {
+    return cap.resolve().map(data -> data.get(key)).orElse(0);
+  }
```
KnightMiner: These seem redundant to those in `ArmorLevelModule`, I'd just ditch the extra copies.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,168 @@
...
+import java.util.List;
+
+public record MaxArmorStatModule(TinkerDataCapability.TinkerDataKey<Float> key, LevelingValue amount, TinkerDataCapability.TinkerDataKey<Integer> key2, boolean allowBroken, @Nullable TagKey<Item> heldTag, ArmorStatModule.TooltipStyle tooltipStyle, ModifierCondition<IToolStackView> condition) implements HookProvider, EquipmentChangeModifierHook, ModifierModule, TooltipModifierHook, ModifierCondition.ConditionalModule<IToolStackView> {
+  private static final List<ModuleHook<?>> TOOLTIP_HOOKS = HookProvider.<ArmorStatModule>defaultHooks(ModifierHooks.EQUIPMENT_CHANGE, ModifierHooks.TOOLTIP);
+  private static final List<ModuleHook<?>> NO_TOOLTIP_HOOKS = HookProvider.<ArmorStatModule>defaultHooks(ModifierHooks.EQUIPMENT_CHANGE);
+  public static final RecordLoadable<MaxArmorStatModule> LOADER = RecordLoadable.create(
+    TinkerDataKeys.FLOAT_REGISTRY.requiredField("key", MaxArmorStatModule::key),
+    LevelingValue.LOADABLE.directField(MaxArmorStatModule::amount),
+    TinkerDataKeys.INTEGER_REGISTRY.requiredField("key2", MaxArmorStatModule::key2),
+    BooleanLoadable.INSTANCE.defaultField("allow_broken", false, MaxArmorStatModule::allowBroken),
```
KnightMiner, line 47: We don't need to register these. Just create a key as part of the module. Use the same name as `key`

Can do either by making this not a record, or just adding a constructor overload.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,168 @@
...
+  public static void addStat(EquipmentChangeContext context, TinkerDataCapability.TinkerDataKey<Float> key, float amount) {
+    context.getTinkerData().ifPresent(data -> {
+      float totalLevels = data.get(key, 0f) + amount;
+      if (totalLevels <= 0.005f) {
+        data.remove(key);
+      } else {
+        data.put(key, totalLevels);
+      }
+    });
+  }
```
KnightMiner: This is wrong. This looks like a duplicate of the normal stat module to me, you also don't want to do max on the stat directly. What you want to do is have an int key which keeps track of the level, and then update the float based on the change in the int key. Probably best to do it without the helper in `MaxArmorLevelModule` as we can avoid fetching the capability twice that way.

Also, no need to have these extra helpers. They are just in the `ArmorLevelModule` and `ArmorStatModule` as we use them outside the modules. This one won't have any usage outside, so a single helper (`addStatIfArmor`) is sufficient

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,168 @@
...
+
+    }
+  }
+
+  @Override
+  public void onUnequip(IToolStackView tool, ModifierEntry modifier, EquipmentChangeContext context) {
+    if (condition.matches(tool, modifier)) {
+      context.getTinkerData().ifPresent(data -> {
+        addStatIfArmor(tool, context, key, amount.compute(data.get(key2, 0)), allowBroken, heldTag);
+      });    }
```
KnightMiner: Formatting

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,168 @@
...
+
+  /**
+   * Gets the total level from the key in the entity modifier data
+   * @param living  Living entity
+   * @param key     Key to get
+   * @return  Level from the key
+   */
+  public static float getStat(LivingEntity living, TinkerDataCapability.TinkerDataKey<Float> key) {
+    return living.getCapability(TinkerDataCapability.CAPABILITY).resolve().map(data -> data.get(key)).orElse(0f);
+  }
```
KnightMiner: We already have this helper in the armor stat module

src/main/java/slimeknights/tconstruct/library/tools/capability/TinkerDataKeys.java
```diff
@@ -42,6 +43,9 @@ static void init() {}
   /** Soul belt level for hotbar preservation */
   TinkerDataKey<Integer> SOUL_BELT = intKey("soul_belt");
 
+  TinkerDataKey<Integer> MELEE_PROTECTION = intKey("melee_protection");
```
KnightMiner: As mentioned above, we don't need this. The reason for the key registry is because we are going to make use of these keys in Java code. If the key is only being used inside a single module, it need not be part of the registry; just create it in the module in a non-static field.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,119 @@
...
+/**
+ * Module for keeping track of the total level of a modifier across all pieces of equipment. Does not support incremental, use {@link ArmorStatModule} for that.
+ * @see ArmorStatModule
+ * @see TinkerDataKey
+ * @see slimeknights.tconstruct.library.modifiers.modules.behavior.ShowOffhandModule
+ */
+public record MaxArmorLevelModule(TinkerDataKey<Integer> key, boolean allowBroken, @Nullable TagKey<Item> heldTag) implements HookProvider, EquipmentChangeModifierHook, ModifierModule {
+  private static final List<ModuleHook<?>> DEFAULT_HOOKS = HookProvider.<MaxArmorLevelModule>defaultHooks(ModifierHooks.EQUIPMENT_CHANGE);
+  public static final RecordLoadable<MaxArmorLevelModule> LOADER = RecordLoadable.create(
+    TinkerDataKeys.INTEGER_REGISTRY.requiredField("key", MaxArmorLevelModule::key),
```
KnightMiner: I am concerned by the usage of `INTEGER_REGISTRY` here, as this module will entirely overwrite any other usage of this key. Our other usages of `INTEGER_REGISTRY` assume only adding is used.

I am inclined to make a new registry for it (same type), though we will have to see how it ends up getting used.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,119 @@
...
+import javax.annotation.Nullable;
+import java.util.List;
+
+/**
+ * Module for keeping track of the total level of a modifier across all pieces of equipment. Does not support incremental, use {@link ArmorStatModule} for that.
+ * @see ArmorStatModule
+ * @see TinkerDataKey
+ * @see slimeknights.tconstruct.library.modifiers.modules.behavior.ShowOffhandModule
+ */
+public record MaxArmorLevelModule(TinkerDataKey<Integer> key, boolean allowBroken, @Nullable TagKey<Item> heldTag) implements HookProvider, EquipmentChangeModifierHook, ModifierModule {
```
KnightMiner: I feel like we should use `Float` instead of `Integer` for the level key, and then use `getEffectiveLevel` to fill it, given all protection modifiers are incremental.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,168 @@
...
+    if (condition.matches(tool, modifier)) {
+      context.getTinkerData().ifPresent(data -> {
+        addStatIfArmor(tool, context, key, amount.compute(data.get(key2, 0)), allowBroken, heldTag);
+      });    }
+  }
+
+  @Override
+  public void addTooltip(IToolStackView tool, ModifierEntry modifier, @Nullable Player player, List<Component> tooltip, TooltipKey tooltipKey, TooltipFlag tooltipFlag) {
+    if (condition.matches(tool, modifier) && (tool.hasTag(TinkerTags.Items.WORN_ARMOR) || heldTag != null && tool.hasTag(heldTag)) && (!tool.isBroken() || allowBroken)) {
+      float value = amount.compute(modifier.getEffectiveLevel());
```
KnightMiner: For the tooltip, only use `getEffectiveLevel` if the player is null. If non-null I think I'd rather use the level from the capability. Though to be most accurate I'd like to only show the tooltip if this piece has the largest level, which might be hard; best heuristic I can think of would work as long as there is not a tie for max. Let me think about this more.

src/main/java/slimeknights/tconstruct/library/data/tinkering/AbstractModifierProvider.java
Waterpicker: Leftover debug I forgot to remove.

src/generated/resources/data/tconstruct/tinkering/modifiers/melee_protection.json
```diff
@@ -6,8 +6,7 @@
       "type": "tconstruct:max_armor_stat",
       "allow_broken": false,
       "each_level": 0.005,
-      "key": "tconstruct:use_item_speed",
-      "key2": "tconstruct:melee_protection"
+      "stat_key": "tconstruct:use_item_speed"
```
KnightMiner: I originally made the comment about the keys back when you had 2 keys just differentiated by numbers. Though I do wonder whether `stat_key` is not better than `key` which I used in `ArmorStatModule`. Not sure whether consistency or readability is more important here.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -50,12 +46,12 @@ public List<ModuleHook<?>> getDefaultHooks() {
...
   public void onEquip(IToolStackView tool, ModifierEntry modifier, EquipmentChangeContext context) {
-    findMaxLevelIfArmor(tool, context, key, modifier.intEffectiveLevel(), allowBroken, heldTag);
+    updateMaxLevelIfArmor(tool, context, key, modifier.intEffectiveLevel(), allowBroken, heldTag);
   }
 
   @Override
   public void onUnequip(IToolStackView tool, ModifierEntry modifier, EquipmentChangeContext context) {
-    findMaxLevelIfArmor(tool, context, key, -modifier.intEffectiveLevel(), allowBroken, heldTag);
+    updateMaxLevelIfArmor(tool, context, key, -modifier.intEffectiveLevel(), allowBroken, heldTag);
   }
```
KnightMiner: Thinking about this more while I was attempting to fall asleep, I think this won't work. Your logic on equip is great, but on unequip the max will always be what was previously equip (as unlike with add, max is not inverted by a negative).

We thus have two options:
1. Bring back what I did with armor max levels before. There should be an object you can use that sets a max level per slot.
2. On unequip, iterate all 6 slots to find the new max level.

Its a question of whether we want efficiency of memory or efficiency of code. My old solution is arguably better just as it scales better, modifier lookup is a loop over a potentially large set even if the slots are a small loop. Either case probably won't play nicely with multiple modifiers using the same key (you will get the max per modifier rather than the sum of both modifiers); if you end up doing the object key then just make this unserializable for now until we need it serializable.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -36,27 +36,30 @@
...
+public record MaxArmorStatModule(TinkerDataCapability.TinkerDataKey<Float> statKey, LevelingValue amount, TinkerDataCapability.TinkerDataKey<Float> key, boolean allowBroken, @Nullable TagKey<Item> heldTag, ArmorStatModule.TooltipStyle tooltipStyle, ModifierCondition<IToolStackView> condition) implements HookProvider, EquipmentChangeModifierHook, ModifierModule, TooltipModifierHook, ModifierCondition.ConditionalModule<IToolStackView> {
   private static final List<ModuleHook<?>> TOOLTIP_HOOKS = HookProvider.<ArmorStatModule>defaultHooks(ModifierHooks.EQUIPMENT_CHANGE, ModifierHooks.TOOLTIP);
   private static final List<ModuleHook<?>> NO_TOOLTIP_HOOKS = HookProvider.<ArmorStatModule>defaultHooks(ModifierHooks.EQUIPMENT_CHANGE);
   public static final RecordLoadable<MaxArmorStatModule> LOADER = RecordLoadable.create(
-    TinkerDataKeys.FLOAT_REGISTRY.requiredField("key", MaxArmorStatModule::key),
+    TinkerDataKeys.FLOAT_REGISTRY.requiredField("stat_key", MaxArmorStatModule::statKey),
```
KnightMiner: Worth noting, you don't need the JSON name and the record name to match, I'll often use a simpler name in the record. Though in this case I could see the record having the more complicated name and JSON being simplier (the record name being `statKey` is important)

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -36,27 +36,30 @@
...
-    TinkerDataKeys.INTEGER_REGISTRY.requiredField("key2", MaxArmorStatModule::key2),
     BooleanLoadable.INSTANCE.defaultField("allow_broken", false, MaxArmorStatModule::allowBroken),
     Loadables.ITEM_TAG.nullableField("held_tag", MaxArmorStatModule::heldTag),
     new EnumLoadable<>(ArmorStatModule.TooltipStyle.class).defaultField("tooltip_style", ArmorStatModule.TooltipStyle.NONE, MaxArmorStatModule::tooltipStyle),
     ModifierCondition.TOOL_FIELD,
     MaxArmorStatModule::new);
 
-  @Override
+  public MaxArmorStatModule(TinkerDataCapability.TinkerDataKey<Float> statKey, LevelingValue amount, boolean allowBroken, @Nullable TagKey<Item> heldTag, ArmorStatModule.TooltipStyle tooltipStyle, ModifierCondition<IToolStackView> condition) {
+    this(statKey, amount, TConstruct.createKey(statKey.getId().getNamespace() + "_data"), allowBroken, heldTag, tooltipStyle, condition);
```
KnightMiner: Don't use `TConstruct.createKey`, use the proper constructor that `createKey` calls that takes a resource location.

For the `_data` prefix, you could use the `LocationExtender.INSTANCE` helper.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -36,27 +36,30 @@
...
+  }
+
+    @Override
```
KnightMiner: minor indentation

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -36,27 +36,30 @@
 import javax.annotation.Nullable;
 import java.util.List;
 
-public record MaxArmorStatModule(TinkerDataCapability.TinkerDataKey<Float> key, LevelingValue amount, TinkerDataCapability.TinkerDataKey<Integer> key2, boolean allowBroken, @Nullable TagKey<Item> heldTag, ArmorStatModule.TooltipStyle tooltipStyle, ModifierCondition<IToolStackView> condition) implements HookProvider, EquipmentChangeModifierHook, ModifierModule, TooltipModifierHook, ModifierCondition.ConditionalModule<IToolStackView> {
+public record MaxArmorStatModule(TinkerDataCapability.TinkerDataKey<Float> statKey, LevelingValue amount, TinkerDataCapability.TinkerDataKey<Float> key, boolean allowBroken, @Nullable TagKey<Item> heldTag, ArmorStatModule.TooltipStyle tooltipStyle, ModifierCondition<IToolStackView> condition) implements HookProvider, EquipmentChangeModifierHook, ModifierModule, TooltipModifierHook, ModifierCondition.ConditionalModule<IToolStackView> {
```
KnightMiner: Rename `key` to something like `levelKey` or `maxKey` or `maxLevelKey`

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -78,41 +81,29 @@ public void onEquip(IToolStackView tool, ModifierEntry modifier, EquipmentChange
...
     if (condition.matches(tool, modifier) && (tool.hasTag(TinkerTags.Items.WORN_ARMOR) || heldTag != null && tool.hasTag(heldTag)) && (!tool.isBroken() || allowBroken)) {
-      float value = amount.compute(modifier.getEffectiveLevel());
-      if (value != 0) {
-        Component name = Component.translatable(Util.makeTranslationKey("armor_stat", key.getId()));
-        switch (tooltipStyle) {
-          case BOOST -> TooltipModifierHook.addFlatBoost(modifier.getModifier(), name, value, tooltip);
-          case PERCENT -> TooltipModifierHook.addPercentBoost(modifier.getModifier(), name, value, tooltip);
-        }
+      if (player != null) {
+        player.getCapability(TinkerDataCapability.CAPABILITY).ifPresent(data -> {
```
KnightMiner: You can use the helper here. `ArmorStatModule.getStat(...)`. Also, if the player is null use the modifier level. Should look something like:

```java
float level = player != null ? ArmorStatModule.getStat(player, levelKey) : modifier.getEffectieLevel();
float value = amount.compute(level);
// ...
```

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -123,33 +114,28 @@ public static void addStat(EquipmentChangeContext context, TinkerDataCapability.
    */
   public static void addStatIfArmor(IToolStackView tool, EquipmentChangeContext context, TinkerDataCapability.TinkerDataKey<Float> key, float amount, boolean allowBroken, @Nullable TagKey<Item> heldTag) {
     if (ArmorLevelModule.validSlot(tool, context.getChangedSlot(), heldTag) && (!tool.isBroken() || allowBroken)) {
-      addStat(context, key, amount);
+      context.getTinkerData().ifPresent(data -> {
```
KnightMiner, line 108: This is still a summation; that is not what we want. A max stat should be something closer to:

```java
float oldLevel = data.get(levelKey, 0f);
float newLevel = MaxArmorLevelModule.updateLevel(data, context, modifierEntry);
data.add(statKey, (newLevel - oldLevel) * scale);
```
`add` is a newer helper hence me not using it in `ArmorStatModule` yet.

src/main/java/slimeknights/tconstruct/library/recipe/worktable/ModifierSetWorktableRecipe.java
```diff
@@ -47,7 +47,7 @@ public class ModifierSetWorktableRecipe extends AbstractWorktableRecipe {
   /** Loader instance */
   public static final RecordLoadable<ModifierSetWorktableRecipe> LOADER = RecordLoadable.create(
     ContextKey.ID.requiredField(),
-    Loadables.RESOURCE_LOCATION.requiredField("data_key", r -> r.dataKey),
+    Loadables.RESOURCE_LOCATION.requiredField("data_key", r -> r.key),
```
KnightMiner: Why did this file change? Overzealous find and replace?

src/main/java/slimeknights/tconstruct/library/recipe/worktable/ModifierSetWorktableRecipeBuilder.java
```diff
@@ -20,21 +20,21 @@
 @RequiredArgsConstructor(access = AccessLevel.PRIVATE)
 @AllArgsConstructor(access = AccessLevel.PRIVATE)
 public class ModifierSetWorktableRecipeBuilder extends AbstractSizedIngredientRecipeBuilder<ModifierSetWorktableRecipeBuilder> {
-  private final ResourceLocation dataKey;
+  private final ResourceLocation key;
```
KnightMiner: Same question here, why did this change?

src/main/java/slimeknights/tconstruct/library/recipe/worktable/ModifierSetWorktableRecipe.java
Waterpicker: I believe so.

src/main/java/slimeknights/tconstruct/library/recipe/worktable/ModifierSetWorktableRecipeBuilder.java
Waterpicker: Yea. It was overzealous replace. I'm just remembering while typing it was far worst so you are seeing the two entries I missed while fixing.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,102 @@
...
+  /**
+   * Calculate the current max value of the given maxLevel in entity modifier data for an armor modifier
+   *
+   * @param context Equipment change context
+   * @param key     Key to modify
+   * @param entry   Entry for the given armor modifier
+   * @return The current max level for key
+   */
+  public static float updateMaxLevel(EquipmentChangeContext context, TinkerDataCapability.ComputableDataKey<ModifierMaxLevel> key, ModifierEntry entry) {
+    var holder = context.getTinkerData().resolve();
```
KnightMiner: I don't like var, makes code needlessly less readible.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,102 @@
...
+
+/**
+ * Module for keeping track of the max level of a modifier across all pieces of equipment.
+ * @see TinkerDataKey
+ * @see slimeknights.tconstruct.library.modifiers.modules.behavior.ShowOffhandModule
+ */
+public record MaxArmorLevelModule(TinkerDataCapability.ComputableDataKey<ModifierMaxLevel> key, boolean allowBroken, @Nullable TagKey<Item> heldTag) implements HookProvider, EquipmentChangeModifierHook, ModifierModule {
+  private static final List<ModuleHook<?>> DEFAULT_HOOKS = HookProvider.<MaxArmorLevelModule>defaultHooks(ModifierHooks.EQUIPMENT_CHANGE);
+  public static final RecordLoadable<MaxArmorLevelModule> LOADER = RecordLoadable.create(
+    new IdParser<>(ResourceLocation::new, "Stat").flatXmap(location -> TinkerDataCapability.ComputableDataKey.of(location, ModifierMaxLevel::new), TinkerDataKey::getId).requiredField("key", MaxArmorLevelModule::key),
```
KnightMiner: Why is "Stat" capitalized? All the other fields are lowercase.

That said, given this key is created specifically for this module, it won't be possible to use it in code so there is currently no point in adding this module via JSON. I'd recommend making this just implement `HookProvider` instead of `ModifierModule` for now and removing its loader. We can always add it back later with a `ModifierMaxLevel` registry if needed.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorLevelModule.java
```diff
@@ -0,0 +1,102 @@
...
+   * @param key     Key to modify
+   * @param entry   Entry for the given armor modifier
+   * @return The current max level for key
+   */
+  public static float updateMaxLevel(EquipmentChangeContext context, TinkerDataCapability.ComputableDataKey<ModifierMaxLevel> key, ModifierEntry entry) {
+    var holder = context.getTinkerData().resolve();
+
+    if(holder.isPresent()) {
+      var data = holder.get();
+      var maxLevel = data.computeIfAbsent(key);
```
KnightMiner: These could just be one line, `ModifierMaxLevel maxLevel = holder.get().computeIfAbsent(key)`. Means you will feel less inclination to use `var` as you don't have to list the type of the holder.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,152 @@
...
+import slimeknights.tconstruct.library.tools.capability.TinkerDataCapability;
+import slimeknights.tconstruct.library.tools.context.EquipmentChangeContext;
+import slimeknights.tconstruct.library.tools.nbt.IToolStackView;
+import slimeknights.tconstruct.library.utils.IdParser;
+import slimeknights.tconstruct.library.utils.Util;
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+public record MaxArmorStatModule(TinkerDataCapability.TinkerDataKey<Float> statKey, LevelingValue amount, TinkerDataCapability.ComputableDataKey<ModifierMaxLevel> maxLevel, boolean allowBroken, @Nullable TagKey<Item> heldTag, ArmorStatModule.TooltipStyle tooltipStyle, ModifierCondition<IToolStackView> condition) implements HookProvider, EquipmentChangeModifierHook, ModifierModule, TooltipModifierHook, ModifierCondition.ConditionalModule<IToolStackView> {
```
KnightMiner: I'd recommend using nested class imports for `TinkerDataKey` and `ComputableDataKey`, I use them in the codebase when its readible. Otherwise they are just super bulky.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,152 @@
...
+    return tooltipStyle == ArmorStatModule.TooltipStyle.NONE ? NO_TOOLTIP_HOOKS : TOOLTIP_HOOKS;
+  }
+
+  @Override
+  public void onEquip(IToolStackView tool, ModifierEntry modifier, EquipmentChangeContext context) {
+    if (condition.matches(tool, modifier)) {
+      context.getTinkerData().ifPresent(data -> {
+        addStatIfArmor(tool, context, statKey, maxLevel, amount, modifier, allowBroken, heldTag);
+      });
+
```
KnightMiner: Why are you fetching the capability here if you never use `data`? You also get it inside `addStatIfArmor`. Just call the `addStatIfArmor` helper directly.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,152 @@
...
+      context.getTinkerData().ifPresent(data -> {
+        addStatIfArmor(tool, context, statKey, maxLevel, amount, modifier, allowBroken, heldTag);
+      });
+    }
+  }
+
+  @Override
+  public void addTooltip(IToolStackView tool, ModifierEntry modifier, @Nullable Player player, List<Component> tooltip, TooltipKey tooltipKey, TooltipFlag tooltipFlag) {
+    if (condition.matches(tool, modifier) && (tool.hasTag(TinkerTags.Items.WORN_ARMOR) || heldTag != null && tool.hasTag(heldTag)) && (!tool.isBroken() || allowBroken)) {
+      float value = player != null ? ArmorStatModule.getStat(player, statKey) : modifier.getEffectiveLevel();
```
KnightMiner: Don't use `statKey`, use `maxLevel`. Otherwise the tooltip will list bonuses not caused by this module. This also lets your next check do `value > 0` since the level cannot be negative.

You may want a helper in `ArmorMaxLevelModule` or perhaps better yet `ArmorMaxLevel` to fetch it now that the type is no longer int or float.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,152 @@
...
+   * @param modifierEntry Modifier entry for stat
+   * @param allowBroken   Whether the tool can work while broken
+   * @param heldTag       Tag to check to validate held items, null means held disallowed
+   */
+  public static void addStatIfArmor(IToolStackView tool, EquipmentChangeContext context, TinkerDataCapability.TinkerDataKey<Float> statKey, TinkerDataCapability.ComputableDataKey<ModifierMaxLevel> maxLevelKey, LevelingValue amount, ModifierEntry modifierEntry, boolean allowBroken, @Nullable TagKey<Item> heldTag) {
+    if (ArmorLevelModule.validSlot(tool, context.getChangedSlot(), heldTag) && (!tool.isBroken() || allowBroken)) {
+      context.getTinkerData().ifPresent(data -> {
+        float oldLevel = data.computeIfAbsent(maxLevelKey).getMax();
+        float newLevel = MaxArmorLevelModule.updateMaxLevel(context, maxLevelKey, modifierEntry);
+        data.add(statKey, amount.compute(newLevel - oldLevel));
```
KnightMiner: I'm noticing a subtle bug here, `LevelingAmount` has both a constant part and a leveling part. The constant part is going to get added every time this is called with the current code. I think the best option would be to modify `LevelingValue#compute` to return `0` if the level is `0`, though I don't feel like analyzing the risk of that. Instead lets add a new helper to `LevelingValue` called `computeForLevel`. We need to call this new helper one for the old level and once for the new, we cannot simply call it on the difference:

```java
public float computeForLevel(float level) {
  return level > 0 ? compute(level) : 0;
}
```

Also, bit of a nit pick, but would rather just fetch the data key once, the helper is not really helping providing any benefit as it has to do a lot of extra work (notably with the cap resolve). Might as well just call set here directly. Updated code would look like:

```java
ModifierMaxLevel maxLevel = data.computeIfAbsent(maxLevelKey);
float oldLevel = maxLevel.getMax();
maxLevel.set(context.getChangedSlot(), modifierEntry.getEffectiveLevel());
data.add(statKey, amount.computeForLevel(maxLevel.get()) - amount.computeForLevel(oldLevel));
```

src/main/java/slimeknights/tconstruct/library/recipe/worktable/ModifierSetWorktableRecipeBuilder.java
```diff
@@ -28,13 +28,13 @@ public class ModifierSetWorktableRecipeBuilder extends AbstractSizedIngredientRe
...
-    return new ModifierSetWorktableRecipeBuilder(dataKey, true);
+  public static ModifierSetWorktableRecipeBuilder setAdding(ResourceLocation key) {
+    return new ModifierSetWorktableRecipeBuilder(key, true);
   }
 
   /** Creates a new recipe for removing from a set */
-  public static ModifierSetWorktableRecipeBuilder setRemoving(ResourceLocation dataKey) {
-    return new ModifierSetWorktableRecipeBuilder(dataKey, false);
+  public static ModifierSetWorktableRecipeBuilder setRemoving(ResourceLocation key) {
+    return new ModifierSetWorktableRecipeBuilder(key, false);
```
KnightMiner: One more case of your accidently rename

KnightMiner (commented):
Looking a lot better. I see one notable bug upon inspection, the rest is mostly stylistic stuff.

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -0,0 +1,152 @@
...
+
+  @Override
+  public RecordLoadable<MaxArmorStatModule> getLoader() {
+    return LOADER;
+  }
+
+  @Override
+  public void addModules(ModuleHookMap.Builder builder) {
+    builder.addModule(new MaxArmorLevelModule(maxLevel, allowBroken, heldTag));
+  }
```
KnightMiner: Since you update the max level in the equipment change hooks, this module is not actually needed, so you can ditch this call. Might mean `MaxArmorLevelModule` ended up unused, hold onto it for now though just in case it ends up being helpful for attributes.

src/main/java/slimeknights/tconstruct/library/json/LevelingValue.java
```diff
@@ -63,7 +68,7 @@ public static LevelingValue eachLevel(float eachLevel) {
     return new LevelingValue(0, eachLevel);
   }
 
-  /** Trait to mix into a builder using leveling values */
+    /** Trait to mix into a builder using leveling values */
```
KnightMiner: Unneeded indentation change, make sure you don't have an autoformatter unless its consistent with the project style (and even then, I'd probably recommend only format changed lines)

src/main/java/slimeknights/tconstruct/library/modifiers/modules/technical/MaxArmorStatModule.java
```diff
@@ -72,10 +67,7 @@ public List<ModuleHook<?>> getDefaultHooks() {
   @Override
   public void onEquip(IToolStackView tool, ModifierEntry modifier, EquipmentChangeContext context) {
     if (condition.matches(tool, modifier)) {
-      context.getTinkerData().ifPresent(data -> {
-        addStatIfArmor(tool, context, statKey, maxLevel, amount, modifier, allowBroken, heldTag);
-      });
-
+      addStatIfArmor(tool, context, statKey, maxLevel, amount, modifier, allowBroken, heldTag);
```
KnightMiner, line 70: This fix was also needed on unequip

src/generated/resources/data/tconstruct/tinkering/modifiers/melee_protection.json
Waterpicker: I chose to use key

src/main/java/slimeknights/tconstruct/library/modifiers/data/ModifierMaxLevel.java
```diff
@@ -40,4 +42,8 @@ public void set(EquipmentSlot slot, float level) {
       }
     }
   }
+
+  public static float getStat(LivingEntity living, TinkerDataCapability.ComputableDataKey<ModifierMaxLevel> key) {
+    return living.getCapability(TinkerDataCapability.CAPABILITY).resolve().map(data -> data.get(key)).map(ModifierMaxLevel::getMax).orElse(0f);
```
KnightMiner, line 48: Don't forget the function docs here. Does not need to be long, just briefly explain when this is useful.

Up to you whether you want to include param in the doc string.

KnightMiner:
Code is looking good to me other than the 1 missing function doc.

Are you planning to do projectile protection in this PR as well? Or just melee protection? Now that you have an idea of how melee works you should be in a pretty good state to do a general attribute version like projectile protection, ping me on discord and I can give you an outline for the parameters to that module I'd expect.

PR #5239, opened 2024-06-12
Added a new JEI category for tool recipes

paypur:


src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ToolRecipes.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  public ItemStack getOutputTool() {
+    return switch(definition.getId().toString()) {
+      case "tconstruct:plate_helmet" -> plateArmor.get(ArmorSlotType.HELMET).getRenderTool();
+      case "tconstruct:plate_chestplate" -> plateArmor.get(ArmorSlotType.CHESTPLATE).getRenderTool();
+      case "tconstruct:plate_leggings" -> plateArmor.get(ArmorSlotType.LEGGINGS).getRenderTool();
+      case "tconstruct:plate_boots" -> plateArmor.get(ArmorSlotType.BOOTS).getRenderTool();
+      default -> StationSlotLayoutLoader.getInstance().get(definition.getId()).getIcon().getValue(ItemStack.class);
+    };
```
KnightMiner: Absolutely no. Do not hardcode things in a giant switch. I told you on Github to add a field to the JSON recipe. If not set, default to the output ID.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ToolRecipes.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  public boolean isBroadTool() {
+    // assumption might not always be true
+    return 8 < slotLayout.getSortIndex() && slotLayout.getSortIndex() < 15;
+  }
```
KnightMiner: Broad tools have 4+ inputs.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ToolRecipes.java
```diff
@@ -0,0 +1,99 @@
...
+import slimeknights.tconstruct.library.tools.part.IToolPart;
+import slimeknights.tconstruct.tools.item.ArmorSlotType;
+
+import java.util.ArrayList;
+import java.util.List;
+
+import static slimeknights.tconstruct.library.tools.definition.module.ToolHooks.TOOL_PARTS;
+import static slimeknights.tconstruct.tools.TinkerTools.plateArmor;
+
+public class ToolRecipes {
```
KnightMiner: What is the point of this file? We already have tool recipes defined as a class

src/main/java/slimeknights/tconstruct/plugin/jei/JEIPlugin.java
```diff
@@ -204,6 +209,15 @@ public void registerRecipes(IRecipeRegistration register) {
...
     register.addRecipes(TConstructJEIConstants.SEVERING, severingRecipes);
 
+    // tool recipes
+    List<ToolRecipes> toolRecipes = ToolDefinitionLoader.getInstance().getRegisteredToolDefinitions()
+      .stream()
+      .filter(definition -> definition.hasMaterials() && !definition.getId().equals(new ResourceLocation("tconstruct", "slime_helmet")) && !definition.getId().equals(new ResourceLocation("tconstruct", "plate_shield")))
+      .sorted(Comparator.comparingInt(a -> StationSlotLayoutLoader.getInstance().get(a.getId()).getSortIndex()))
+      .map(ToolRecipes::new)
+      .toList();
+    register.addRecipes(TConstructJEIConstants.TOOL_RECIPES, toolRecipes);
```
KnightMiner: This is wrong. Having a tool definition does not make it have a recipe. Just fetch the recipes.

src/main/java/slimeknights/tconstruct/plugin/jei/TConstructJEIConstants.java
```diff
@@ -54,6 +55,7 @@ public Modifier getBase(ModifierEntry ingredient) {
   // tinker station
   public static final RecipeType<IDisplayModifierRecipe> MODIFIERS = type("modifiers", IDisplayModifierRecipe.class);
   public static final RecipeType<SeveringRecipe> SEVERING = type("severing", SeveringRecipe.class);
+  public static final RecipeType<ToolRecipes> TOOL_RECIPES = type("tool_recipes", ToolRecipes.class);
```
KnightMiner: Use the existing tool recipe class, not a new one.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolRecipesCategory.java
```diff
@@ -0,0 +1,124 @@
...
+import slimeknights.tconstruct.library.recipe.tinkerstation.ToolRecipes;
+import slimeknights.tconstruct.library.tools.layout.LayoutSlot;
+import slimeknights.tconstruct.tools.TinkerTools;
+
+import javax.annotation.Nonnull;
+import java.util.Collections;
+import java.util.List;
+
+
+// IDK where to put this so im just leaving it the main jei directory
```
KnightMiner: These sorts of comments belong on the PR description or in a comment, not in the code.

src/main/resources/assets/tconstruct/lang/en_us.json
```diff
@@ -3031,6 +3031,7 @@
   "jei.tconstruct.modifiers.level.exact": "For level: %s",
 
   "jei.tconstruct.severing.title": "Severing",
+  "jei.tconstruct.tinkering.tool_recipes": "Tool Recipes",
```
KnightMiner: Call it tool building.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolRecipesCategory.java
```diff
@@ -0,0 +1,124 @@
...
+
+    // centers slots vertically
+    int yOffset = (HEIGHT - (ITEM_SIZE + maxY - minY)) / 2 - minY;
+    // centers slots horizontally within square
+    int xOffset = (HEIGHT - (ITEM_SIZE + maxX - minX)) / 2 - minX;
+
+    return new xy(xOffset, yOffset);
+  }
+
+  private record xy(int x, int y) {}
```
KnightMiner: This is not a name for a class, class names start with capitals. Plus this already exsits in Mojangs libraries, Vec2i or something like that. Look at what I used in `ToolModel`

src/main/java/slimeknights/tconstruct/plugin/jei/ToolRecipesCategory.java
```diff
@@ -0,0 +1,124 @@
...
+    for (LayoutSlot slot : recipe.getSlots()) {
+      // need to offset by 1 because the inventory slot icons are 18x18
+      this.SLOT.draw(stack, slot.getX() + offsets.x - 1, slot.getY() + offsets.y - 1);
+    }
+  }
+
+  @Override
+  public List<Component> getTooltipStrings(ToolRecipes recipe, IRecipeSlotsView recipeSlotsView, double mouseX, double mouseY) {
+    return recipe.isBroadTool() && GuiUtil.isHovered((int) mouseX, (int) mouseY, 73, 42, ITEM_SIZE, ITEM_SIZE) ?
+      Collections.singletonList(MutableComponent.create(new LiteralContents("Broad tools require the Tinker's Anvil to be created!"))) :
```
KnightMiner: Never use string literals for text displayed in game. Always make things translatable.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolRecipesCategory.java
```diff
@@ -0,0 +1,124 @@
...
+  private final IDrawable ANVIL, SLOT;
+  private final int WIDTH = 128;
+  private final int HEIGHT = 66;
+  private final int ITEM_SIZE = 16;
+
+  public ToolRecipesCategory(IGuiHelper guiHelper) {
+    this.icon = guiHelper.createDrawableItemStack(TinkerTools.pickaxe.get().getRenderTool());
+    this.background = guiHelper.createDrawable(BACKGROUND_LOC, 0, 0, WIDTH, HEIGHT);
+    this.SLOT = guiHelper.createDrawable(BACKGROUND_LOC, 128, 0, ITEM_SIZE + 2, ITEM_SIZE + 2);
+    this.ANVIL = guiHelper.createDrawable(BACKGROUND_LOC, 146, 0, ITEM_SIZE, ITEM_SIZE);
```
KnightMiner: All caps is only for static finals, for regular finals don't do all caps.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolRecipesCategory.java
```diff
@@ -0,0 +1,124 @@
...
+  private static  final ResourceLocation BACKGROUND_LOC = TConstruct.getResource("textures/gui/jei/tool_recipes.png");
+  private static final Component TITLE = TConstruct.makeTranslation("jei", "tinkering.tool_recipes");
+  @Getter
+  private final IDrawable icon;
+  @Getter
+  private final IDrawable background;
+  private final IDrawable ANVIL, SLOT;
+  private final int WIDTH = 128;
+  private final int HEIGHT = 66;
+  private final int ITEM_SIZE = 16;
```
KnightMiner: These should be static final since they are not changed by the constructor.

src/main/resources/assets/tconstruct/textures/gui/jei/tool_recipes.png
KnightMiner, line 1: I feel like this could be merged into one of the existing JEI images, that large slot I'm sure we used by now. Perhaps the one we use for modifiers?

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ToolRecipes.java
```diff
@@ -0,0 +1,99 @@
...
+        yield definition.getId();
+    };
+    this.slotLayout = StationSlotLayoutLoader.getInstance().get(id);
+  }
+
+  static List<Material> MATERIALS = MaterialRegistry.getMaterials()
+    .stream()
+    .map(material -> (Material) material)
+    .filter(material -> !material.isHidden())
+    .toList();
```
KnightMiner: This is not reload safe at all. An why are you doing an unchecked cast? `IMaterial` has everything you need. Just run this stream in the method that uses it.

Plus, I am certain we have a method to get all visible materials; we use that a ton. So just call the existing method.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolRecipesCategory.java
```diff
@@ -0,0 +1,124 @@
...
+
+    List<LayoutSlot> slots = recipe.getSlots();
+    List<ItemStack> items = recipe.getDisplayParts();
+
+    assert items.size() == slots.size();
+
+    xy offsets = getOffsets(recipe);
+    for (int i = 0; i < items.size(); i++) {
+      builder.addSlot(RecipeIngredientRole.INPUT, slots.get(i).getX() + offsets.x, slots.get(i).getY() + offsets.y).addItemStack(items.get(i));
+    }
```
KnightMiner: This is missing the extra inputs to the tool recipes which are defined in the tool recipe. It really seems like you do not know about the tool recipe; hence why I mentioned it on Discord.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolRecipesCategory.java
paypur: Where should I put it then?

KnightMiner: Doesn't matter.

paypur:
Things left to fix

- I'm not able to add a new loadable to `ToolBuildingRecipe` without getting `cannot resolve symbol` errors for the fields.
- With my limited testing because the code doesn't build anymore, `getIngredients()` in `ToolBuildingRecipe` just returns an empty list every time so this might be wrong.
- I merged the textures into the tinker station, but because I need a blank background with no input slots I just duplicated a lot of the texture which might not be ideal.

src/generated/resources/data/tconstruct/recipes/tools/armor/plate_boots.json
```diff
@@ -1,5 +1,6 @@
 {
   "type": "tconstruct:tool_building",
   "result": "tconstruct:plate_boots",
-  "result_count": 1
-}
\ No newline at end of file
+  "result_count": 1,
+  "slot_id": "tconstruct:plate_armor"
```
KnightMiner: Call it `layout` or `slot_layout`, `slot_id` is unclear and 90% of recipes are IDs

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -38,6 +39,7 @@ public class ToolBuildingRecipe implements ITinkerStationRecipe {
     LoadableRecipeSerializer.RECIPE_GROUP,
     TinkerLoadables.MODIFIABLE_ITEM.requiredField("result", r -> r.output),
     IntLoadable.FROM_ONE.defaultField("result_count", 1, true, r -> r.outputCount),
+    ResourceLocationLoadable.DEFAULT.defaultField("slot_id", r -> r.getId, r -> r.slotId),
```
KnightMiner: This is not syntactically valid, a defaulting field has to default to a value, not a function. Just use a nullable field. 

Just map null to the value in the constructor (remove `@AllArgsConstructor` and use the default constructor). I probably need a variant of nullable fields for "only nullable in JSON" which also lets me specify to read from a different field name if missing and to skip serializing if the two match. But thats not something for you to worry about (for now you will just have some redundant info in the JSON)

Since you are adding a new constructor, keep the old signature (without the layout ID) around but marked as deprecated calling the new one and redirect to the new

src/generated/resources/data/tconstruct/recipes/tools/armor/plate_leggings.json
```diff
@@ -1,5 +1,6 @@
 {
   "type": "tconstruct:tool_building",
   "result": "tconstruct:plate_leggings",
-  "result_count": 1
-}
```
KnightMiner, line 5: Somehow I feel like you manually modified these, don't do that. Use datagen.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipeBuilder.java
```diff
@@ -39,6 +40,7 @@ public void save(Consumer<FinishedRecipe> consumerIn) {
   @Override
   public void save(Consumer<FinishedRecipe> consumerIn, ResourceLocation id) {
     ResourceLocation advancementId = this.buildOptionalAdvancement(id, "parts");
-    consumerIn.accept(new LoadableFinishedRecipe<>(new ToolBuildingRecipe(id, group, output, outputSize, extraRequirements), ToolBuildingRecipe.LOADER, advancementId));
+    slotId = id;
```
KnightMiner: This is the wrong way to do this, the slot ID will not be the recipe ID. Just add a setter for the slot ID.

src/main/java/slimeknights/tconstruct/plugin/jei/JEIPlugin.java
```diff
@@ -235,6 +241,9 @@ public void registerRecipeCatalysts(IRecipeCatalystRegistration registry) {
     registry.addRecipeCatalyst(new ItemStack(TinkerTables.tinkerStation), TConstructJEIConstants.MODIFIERS);
     registry.addRecipeCatalyst(new ItemStack(TinkerTables.tinkersAnvil), TConstructJEIConstants.MODIFIERS);
     registry.addRecipeCatalyst(new ItemStack(TinkerTables.scorchedAnvil), TConstructJEIConstants.MODIFIERS);
+    registry.addRecipeCatalyst(new ItemStack(TinkerTables.tinkerStation), TConstructJEIConstants.TOOL_BUILDING);
+    registry.addRecipeCatalyst(new ItemStack(TinkerTables.tinkersAnvil), TConstructJEIConstants.TOOL_BUILDING);
+    registry.addRecipeCatalyst(new ItemStack(TinkerTables.scorchedAnvil), TConstructJEIConstants.TOOL_BUILDING);
```
KnightMiner: `addRecipeCatalyst` was varardic last I checked, so you can just append these to the end of the previous three lines.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipeBuilder.java
```diff
@@ -23,6 +23,7 @@ public class ToolBuildingRecipeBuilder extends AbstractRecipeBuilder<ToolBuildin
   private final IModifiable output;
   @Setter
   private int outputSize = 1;
+  private ResourceLocation slotId;
```
KnightMiner: `@Nullable @Setter`, maybe an `= null` for good measure (mostly for readibility)

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,125 @@
...
+    this.slot = guiHelper.createDrawable(BACKGROUND_LOC, 24, 14, SLOT_SIZE, SLOT_SIZE);
+    this.anvil = guiHelper.createDrawable(BACKGROUND_LOC, 128, 61, ITEM_SIZE, ITEM_SIZE);
+  }
+
+  @Override
+  public void setRecipe(IRecipeLayoutBuilder builder, ToolBuildingRecipe recipe, IFocusGroup focuses) {
+    recipe.getIngredients().forEach(ingredient -> builder.addInvisibleIngredients(RecipeIngredientRole.INPUT).addIngredients(ingredient));
+    recipe.getExtraRequirements().forEach(ingredient -> builder.addSlot(RecipeIngredientRole.INPUT, 0, 0).addIngredients(ingredient));
+
+    List<LayoutSlot> slots = StationSlotLayoutLoader.getInstance().get(recipe.getSlotId()).getInputSlots();
```
KnightMiner: Pretty sure `get()` on this line is nullable, if not it has some value that means "no layout was found" which should cause you to stop additional logic.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,125 @@
...
+    List<ItemStack> items = recipe.getIngredients().stream().flatMap(ingredient -> Arrays.stream(ingredient.getItems())).toList();
+
+    if (items.size() != slots.size()) {
+      TConstruct.LOG.error("Part count and slot count for %s do not match!", recipe.getId().toString());
+    }
```
KnightMiner: This will always fail. The condition should be `parts.size() + items.size() != slot.size()`. If you wished, you could add those helpers you had earlier that cached the display stack lists to `ToolBuildingRecipe` to make this easier; `getIngredients` only works with the additional ingredients, not the parts. Take a look at `ToolBuildingRecipe#matches`, should give you insight into how to do this.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,125 @@
...
+  @Override
+  public List<Component> getTooltipStrings(ToolBuildingRecipe recipe, IRecipeSlotsView recipeSlotsView, double mouseX, double mouseY) {
+    return isBroadTool(recipe) && GuiUtil.isHovered((int) mouseX, (int) mouseY, 73, 42, ITEM_SIZE, ITEM_SIZE) ?
+      List.of(TConstruct.makeTranslation("jei", "tinkering.tool_recipes.anvil")) :
+      List.of();
+  }
+
+  private boolean isBroadTool(ToolBuildingRecipe recipe) {
+    return recipe.getIngredients().size() >= 4;
+  }
```
KnightMiner: Put this helper on `ToolBuildingRecipe`. Alternatively, ask the slot layout list.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,125 @@
...
+
+  private Vec2 getOffsets(List<LayoutSlot> slots) {
+    int minX, maxX, minY, maxY;
+    minX = slots.get(0).getX();
+    maxX = slots.get(0).getX();
+    minY = slots.get(0).getY();
+    maxY = slots.get(0).getY();
```
KnightMiner: I asked earlier, why are the offsets needed? The slot positions work fine in the anvil.

src/main/resources/assets/tconstruct/lang/en_us.json
```diff
@@ -3032,6 +3032,9 @@
 
   "jei.tconstruct.severing.title": "Severing",
 
+  "jei.tconstruct.tinkering.tool_building": "Tool Building",
+  "jei.tconstruct.tinkering.tool_building.anvil": "Broad tools require the Tinker's Anvil to be created, and to swap parts!",
```
KnightMiner: Perhaps:

> Broad tools require the Tinker's Anvil to be created or swap parts

src/main/resources/assets/tconstruct/textures/gui/jei/tinker_station.png
KnightMiner, line 1: Hmm, this file keeps getting more and more non-tinker station stuff, I should rename it at some point...

Not something you need to worry about.

src/main/resources/assets/tconstruct/textures/gui/jei/tool_recipes.png
KnightMiner, line 1: This file can be deleted now, right?

paypur: Yeah I forgot to remove it

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
paypur: My texture is smaller than the station/anvil texture so I can't just use the layout slot positions. I could resize my texture to be the same size as station/anvil texture if you would like.

KnightMiner: I think that would be preferable. Offsetting the slots dynamically like this suggests they are wrongly placed in the original UI; which perhaps they are, if its just a single outlier then just fix the outlier. I am not against a static offset (same numbers used for all layouts).

In the future, we may want to set up some validation bounds on them in the future now that we have multiple distinct UIs using the numbers. But that's out of scope of this PR for now.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -48,13 +52,44 @@ public class ToolBuildingRecipe implements ITinkerStationRecipe {
...
+  protected ResourceLocation layoutSlot;
   protected final List<Ingredient> ingredients;
 
+  /** @deprecated Use {@link #ToolBuildingRecipe(ResourceLocation, String, IModifiable, int, ResourceLocation, List)} */
+  @Deprecated
+  public ToolBuildingRecipe(ResourceLocation id, String group, IModifiable output, int outputCount, List<Ingredient> ingredients) {
+    this(id, group, output, outputCount, null, ingredients);
+  }
+
+  public ToolBuildingRecipe(ResourceLocation id, String group, IModifiable output, int outputCount, ResourceLocation layoutSlot, List<Ingredient> ingredients) {
```
KnightMiner: Missing `@Nullable` on `ResourceLocation layoutSlot`

Though, I think this is just a normal `@AllArgsConstructor` so you could delete this and bring back the annotation. You only need to manually define it if you wished to handle the nullable behavior on construct (which on second thought the way you did it in the getter is probably better for the sake of serialization)

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -48,13 +52,44 @@ public class ToolBuildingRecipe implements ITinkerStationRecipe {
...
+    this.outputCount = outputCount;
+    this.layoutSlot = layoutSlot;
+    this.ingredients = ingredients;
+  }
+
+  public ResourceLocation getLayoutSlotId() {
+    return Objects.requireNonNullElse(layoutSlot, getOutput().getToolDefinition().getId());
+  }
+
+  public List<IToolPart> getToolParts() { return ToolPartsHook.parts(getOutput().getToolDefinition()); }
```
KnightMiner: Toss some quick function javadoc comments on these explaining what they do for all these new functions here, see `getExtraRequirements` for an example. My rule of thumb is every method must be either `@Override` or have a javadoc comment. 

Also, please don't make a function in just 1 line, write it out over 3 at minimum.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -48,13 +52,44 @@ public class ToolBuildingRecipe implements ITinkerStationRecipe {
...
   public List<Ingredient> getExtraRequirements() {
     return ingredients;
   }
 
+  public List<LayoutSlot> getLayoutSlots() {
+    return StationSlotLayoutLoader.getInstance().get(getLayoutSlotId()).getInputSlots();
+  }
+
+  public boolean isBroadTool() {
+    return getToolParts().size() >= 4;
```
KnightMiner: You also need to check `getExtraRequirements()`'s size, its the sum of the two being 4 or more that is "broad tools" (though I think a name like `requiresAnvil` might be better)

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipeBuilder.java
```diff
@@ -23,8 +24,15 @@ public class ToolBuildingRecipeBuilder extends AbstractRecipeBuilder<ToolBuildin
   private final IModifiable output;
   @Setter
   private int outputSize = 1;
+  @Nullable
+  private ResourceLocation layoutSlot = null;
   private final List<Ingredient> extraRequirements = new ArrayList<>();
 
+  public ToolBuildingRecipeBuilder setLayoutSlot(ResourceLocation layoutSlot) {
```
KnightMiner: You should be able to just use a `@Setter` on `layoutSlot` for this, note it will name the setter `layoutSlot`, not `setLayoutSlot` but thats fine.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,147 @@
...
+    List<IToolPart> parts = recipe.getToolParts();
+    List<ItemStack> extraItems = recipe.getIngredients().stream().flatMap(ingredient -> Arrays.stream(ingredient.getItems())).toList();
+    List<LayoutSlot> layoutSlots = recipe.getLayoutSlots();
+
+    // every part and its material variants
+    List<ItemStack> allInputParts = parts.stream()
+        .flatMap(part -> MaterialRegistry.getInstance().getVisibleMaterials().stream()
+          .filter(part::canUseMaterial)
+          .flatMap(mat -> Stream.of(part.withMaterial(mat.getIdentifier()))))
+        .toList();
```
KnightMiner: This might be worth caching in `ToolBuildingRecipe` like you did back when you wrote this PR with your custom class.

also, if this is only used for a single invisible ingredient, just add it directly to that line; no need for an extra variable.

----

Aside, the comment should be expanded as this threw me off at first. State that we display "display tool parts" in the recipe and this is just for tool part recipe lookup, hence the single list instead of a list per slot. Does not need to be so worry, just needs to explain why this single list looks wrong but is not.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,147 @@
...
+    this.icon = guiHelper.createDrawableItemStack(TinkerTools.pickaxe.get().getRenderTool());
+    this.background = guiHelper.createDrawable(BACKGROUND_LOC, 128, 77, WIDTH, HEIGHT);
+    this.slot = guiHelper.createDrawable(BACKGROUND_LOC, 24, 14, SLOT_SIZE, SLOT_SIZE);
+    this.anvil = guiHelper.createDrawable(BACKGROUND_LOC, 128, 61, ITEM_SIZE, ITEM_SIZE);
+  }
+
+  @Override
+  public void setRecipe(IRecipeLayoutBuilder builder, ToolBuildingRecipe recipe, IFocusGroup focuses) {
+    List<IToolPart> parts = recipe.getToolParts();
+    List<ItemStack> extraItems = recipe.getIngredients().stream().flatMap(ingredient -> Arrays.stream(ingredient.getItems())).toList();
```
KnightMiner: Don't use `getIngredients()`, use the custom getter, `getExtraRequirements()`. I think `getIngredients()` might actually be a dumb vanilla thing that we don't use.

Also, this looks very wrong. You should not be combining all extra items into a single list. Every ingredient in the list is 1 slot, and that slot contains a list. This should be a `List<List<ItemStack>>`

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,147 @@
...
+    // every part and its material variants
+    List<ItemStack> allInputParts = parts.stream()
+        .flatMap(part -> MaterialRegistry.getInstance().getVisibleMaterials().stream()
+          .filter(part::canUseMaterial)
+          .flatMap(mat -> Stream.of(part.withMaterial(mat.getIdentifier()))))
+        .toList();
+
+    if (parts.size() + extraItems.size() != layoutSlots.size()) {
+      TConstruct.LOG.error(String.format("Tool part count and layout slot count for %s do not match!", recipe.getId()));
+      return;
```
KnightMiner: Is it possible for us to recover instead of giving up? That is:

* If the layout has too many slots, just display the extra slots as empty.
* If the layout has too few slots, add the extra slots in some hardcoded location.

don't put too much effort into this, just want to give the user as much usable info as possible; it will mostly end up helping modders I think.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    builder.addInvisibleIngredients(RecipeIngredientRole.INPUT).addItemStacks(allInputParts);
+    builder.addInvisibleIngredients(RecipeIngredientRole.INPUT).addItemStacks(extraItems);
```
KnightMiner: Extra items should not be invisible. They are part of the layout.

Do you not know how extra items work? Ping me on Discord and I can explain them to you, or look at this picture below for laminar armor:

image

Only 3 of those 5 are tool parts, the other 2 (the leather) are extra items.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,147 @@
...
+      List.of(TConstruct.makeTranslation("jei", "tinkering.tool_building.anvil")) :
+      List.of();
+  }
+
+  // use display parts to be more consistent
+  private ItemStack getDisplayPart(IToolPart toolPart, int i) {
+      ItemStack item = toolPart.withMaterialForDisplay(ToolBuildHandler.getRenderMaterial(i));
+      item.getOrCreateTag().putBoolean(TooltipUtil.KEY_DISPLAY, true);
+      return item;
+  }
```
KnightMiner: Is this not a helper method already? Pretty sure we do this in the books. If its not a helper method already, make it one in `ToolBuildHandler` so this code is only in one place (I believe its `ContentTool` or `ToolContent` (whatever that class is called) that would be doing this already)

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,147 @@
...
+    Vec2 offsets = getOffsets(layoutSlots);
+    for (int i = 0; i < parts.size(); i++) {
+      builder.addSlot(RecipeIngredientRole.INPUT, (int) (layoutSlots.get(i).getX() + offsets.x), (int) (layoutSlots.get(i).getY() + offsets.y))
+             .addItemStack(getDisplayPart(parts.get(i), i));
+    }
+
+    if (recipe.getOutput() instanceof IModifiableDisplay modifiable) {
+      builder.addSlot(RecipeIngredientRole.OUTPUT, WIDTH - 23, 23)
+        .addItemStack(modifiable.getRenderTool());
+    }
```
KnightMiner: This needs a fallback if that instanceof check fails. Probably just a `new ItemStack(recipe.getOutput())`

Probably best to write as
```java
IModifiable output = recipe.getOutput();
ItemStack outputStack = output instanceof IModifiableDisplay display ? display.getRenderTool() : new ItemStack(output);
```
so you need only call `addSlot()` with those location parameters in 1 spot.

src/main/resources/assets/tconstruct/lang/en_us.json
```diff
@@ -3032,6 +3032,9 @@
 
   "jei.tconstruct.severing.title": "Severing",
 
+  "jei.tconstruct.tinkering.tool_building": "Tool Building",
+  "jei.tconstruct.tinkering.tool_building.anvil": "Broad tools require the Tinker's Anvil to be created or swap parts",
```
KnightMiner, line 3036: Wonder if instead of "broad tools" we should just say "Tools with 4 or more parts" since the term "broad" is used kinda sparingly and that gives a more precise definition for players.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
paypur: It is used in `ContentTool` (where I borrowed it from), but its not a separate method.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipeBuilder.java
paypur: I don't understand. Is this not how the builder pattern is done?

KnightMiner: Lombok lets you generate the builder pattern automatically, look at the other `@Setter` in this class.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -48,13 +56,58 @@ public class ToolBuildingRecipe implements ITinkerStationRecipe {
...
+   * Gets all tool parts as and all its variants for JEI input lookups.
+   */
+  public List<List<ItemStack>> getAllToolParts() {
+    return getToolParts().stream()
+      .map(part -> MaterialRegistry.getInstance().getVisibleMaterials().stream()
+        .filter(part::canUseMaterial)
+        .map(mat -> part.withMaterial(mat.getIdentifier()))
+        .toList())
+      .toList();
+  }
```
KnightMiner: Lets cache this in the recipe object instead of recreating it each time.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -48,13 +56,58 @@ public class ToolBuildingRecipe implements ITinkerStationRecipe {
...
+  @Deprecated
+  public ToolBuildingRecipe(ResourceLocation id, String group, IModifiable output, int outputCount, List<Ingredient> ingredients) {
+    this(id, group, output, outputCount, null, ingredients);
+  }
+
+  /**
+   * Station slot layout ids and tool definition ids are not always the same
+   * layoutSlot specified in the tool recipe JSON is potentially null,
+   * so we should fallback to the tool definition id
+   * @return StationSlotLayout id
```
KnightMiner: This `@return` is kinda weird. I'd probably just drop this last line.

You could also simplify this description a lot,
> Gets the ID of the station slot layout for displaying this recipe.
> Typically matches the output definition ID, but some tool recipes share a single layout.

The internal nullability does not matter to the caller.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,119 @@
...
+    List<IToolPart> parts = recipe.getToolParts();
+    List<List<ItemStack>> allParts = recipe.getAllToolParts();
+    List<List<ItemStack>> extras = recipe.getExtraRequirements().stream().map(ingredient -> Arrays.asList(ingredient.getItems())).toList();
+    List<LayoutSlot> layoutSlots = recipe.getLayoutSlots();
+
+    if (parts.size() + extras.size() > layoutSlots.size()) {
+      TConstruct.LOG.error(String.format("Tool part count is greater than layout slot count for %s!", recipe.getId()));
+      int additionalSlots = 0;
+      layoutSlots = new ArrayList<>(layoutSlots);
+      while (parts.size() + extras.size() > layoutSlots.size()) {
```
KnightMiner: This could be a for loop which would be more readible. Same below.

```java
int missingSlots = parts.size() + extras.size() - layoutSlots.size()
for (int additionalSlot = 0; additionalSlot < missingSlots; additionalSlot++)
```

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,119 @@
...
+      TConstruct.LOG.error(String.format("Tool part count is less than layout slot count for %s!", recipe.getId()));
+      extras = new ArrayList<>(extras);
+      while (parts.size() + extras.size() < layoutSlots.size()) {
+        extras.add(List.of(Items.BARRIER.getDefaultInstance()));
+      }
+    }
+
+    for (int i = 0; i < layoutSlots.size(); i++) {
+      builder.addSlot(RecipeIngredientRole.INPUT, layoutSlots.get(i).getX() + X_OFFSET, layoutSlots.get(i).getY() + Y_OFFSET)
+             .addItemStacks(allParts.get(i < parts.size() ? i : i - parts.size()));
```
KnightMiner: this should be using `extras` when the indexes get higher, not whatever you are doing with that indexing

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,119 @@
...
+        layoutSlots.add(new LayoutSlot(null, null, additionalSlots * SLOT_SIZE, 0, null));
+        additionalSlots++;
+      }
+    }
+
+    if (parts.size() + extras.size() < layoutSlots.size()) {
+      TConstruct.LOG.error(String.format("Tool part count is less than layout slot count for %s!", recipe.getId()));
+      extras = new ArrayList<>(extras);
+      while (parts.size() + extras.size() < layoutSlots.size()) {
+        extras.add(List.of(Items.BARRIER.getDefaultInstance()));
```
KnightMiner: eh, just leave it as `List.of()` (or `ItemStack.EMPTY` if JEI refuses to allow an empty list), let JEI draw empty slots. That is most consistent with its functional behavior in game (the extra slots get ignored)

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -38,7 +37,6 @@
  * This recipe is used for crafting a set of parts into a tool
  */
 
-@AllArgsConstructor
```
KnightMiner, line 34: If you make `layoutSlot` final, you can do a `@RequiredArgsConstructor` here. I don't think `layoutSlot` has a need to be non-final.

KnightMiner:
I believe most of my comments are resolved at this point apart from a minor stylistic thing. When I get a chance I'll test it in game.

In the meantime, have you tested a tool that uses the "extra" ingredients? I cannot remember if any exist in the base mod, so easiest way to test might be making a build (`./gradlew build`) then loading up an instance with Tinkers' Construct and Tinkers' Json Things together. Alternatively, temporarily add extra ingredients to one of the base mod tools for testing (datagen the temporary recipe but don't include it in the PR)

paypur:
Yeah tools with extras work

I'm also fixing this in next commit to not be really weird. Presumably you also need to update Tinker's Things to specify a layout slot id

KnightMiner:
Yeah. laminar armor's layout would need specific compat, that is up to the addon to handle. Perhaps have it fallback to the anvil 5 slot layout if the layout is missing?

```
if layoutId not null
  return layoutId 
if tool definition ID is a layout
  return tool definition ID
return anvil 5 slot layout
```
With the 5 slot being the fallback, probably best to remove the warning if the layout has too many slots, perhaps just disable the warning if the layout ID is the default layout.

Guess if you really wanted to be fancy, 
```
if layoutId not null
  return layoutId 
if tool definition ID is a layout
  return tool definition ID
if requires anvil
  return anvil 5 slot layout
return tinker station 3 slot layout
```
If those end up looking bad, just define some new layouts for the sake of being the fallback. Sorta like how the anvil layout is not tool specific. Though hopefully the existing layouts will work fine so we don't need to do that.

paypur:
Wouldn't we have the same problem if a recipe had 6 ingredients for some reason?

KnightMiner:
6 ingredients is an unlikely edge case as its not supported in any of the tables. Thus, if you are going to add 6 ingredients, its on you to set the layout properly. They could also extend the tool building recipe to change the default layout in that case.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -54,18 +55,21 @@ public void setRecipe(IRecipeLayoutBuilder builder, ToolBuildingRecipe recipe, I
       recipe.getExtraRequirements().stream().map(ingredient -> Arrays.asList(ingredient.getItems()))).toList();
     List<LayoutSlot> layoutSlots = recipe.getLayoutSlots();
 
+    if (layoutSlots.isEmpty()) {
+      // fallback to tinker station or anvil
+      layoutSlots = StationSlotLayoutLoader.getInstance().get(TConstruct.getResource(recipe.requiresAnvil() ? "tinkers_anvil" : "tinker_station")).getInputSlots();
+    }
```
KnightMiner: I kinda want to leave this up to the recipe to handle. Perhaps instead of (or in addition to) `getSlotLayoutId` we have a method like `getLayoutSlots` which also caches the list in the recipe instance? Advantage of a getter there is you could subclass the recipe and swap out the default if you wish for a more complicated tool recipe.

The cache also helps as we need this list in 2 spots.

src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -0,0 +1,126 @@
...
+      this.anvil.draw(stack, 76, 44);
+    }
+
+    List<LayoutSlot> layoutSlots = recipe.getLayoutSlots();
+
+    int missingSlots = recipe.getAllToolParts().size() + recipe.getExtraRequirements().size() - layoutSlots.size();
+
+    // check layout slots for issues
+    if (missingSlots > 0) {
+      TConstruct.LOG.error(String.format("Tool part count is greater than layout slot count for %s!", recipe.getId()));
```
KnightMiner: Hopefully the last thing, isn't this error also printing in `setRecipe`? Little redundant to print in both, especial since IIRC draw runs every frame so this message will be printing a lot. Just keep the `setRecipe` copy.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/building/ToolBuildingRecipe.java
```diff
@@ -31,10 +31,15 @@
...
 
 import javax.annotation.Nullable;
+import java.util.ArrayList;
 import java.util.List;
 import java.util.Objects;
 import java.util.stream.IntStream;
 
+import static slimeknights.tconstruct.plugin.jei.ToolBuildingCategory.SLOT_SIZE;
+import static slimeknights.tconstruct.plugin.jei.ToolBuildingCategory.X_OFFSET;
+import static slimeknights.tconstruct.plugin.jei.ToolBuildingCategory.Y_OFFSET;
```
KnightMiner: You should not reference JEI code outside of JEI plugins, could cause a crash when JEI is not installed. Move these constants to this class (or copy if they existed before your PR)

PR #5269, opened 2024-10-15
add com.google.code.findbugs to remove '@Nullable' errors

Goldenlion5648:
When I loaded the project in Intellij, it complained about "@Nullable" for example in https://github.com/SlimeKnights/TinkersConstruct/blob/62dc631581963e32397dff774c477363b9c77a39/src/main/java/slimeknights/tconstruct/tables/block/TinkerStationBlock.java#L9

Also noted in the readme that java 18 must be used

### DIFF
--- a/README.md
+++ b/README.md
@@ -11,7 +11,7 @@ Anything that is not possible via IMC has to be integrated via Code through the 
 (old 1.7.10 IMCs can be found here: https://gist.github.com/bonii-xx/e46f9d9e81e29d796b1b)
 
 ## Setting up a Workspace/Compiling from Source
-Note: Git MUST be installed and in the system path to use our scripts.
+Note: Git MUST be installed and in the system path to use our scripts. Java 18 must be used.
 * Setup: Run [gradle]in the repository root: `gradlew[.bat] [setupDevWorkspace|setupDecompWorkspace] [eclipse|idea]`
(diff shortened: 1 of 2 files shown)


KnightMiner:
I don't see why this would be needed. We just setup a workspace using Gradle, and its always fetched that dependency for us just fine, has been doing so for years. That comes from javax, which is not google, its oracle.

As for the Java 18 requirement, its not a requirement? We use Java 17, as specified by [our build.gradle](https://github.com/SlimeKnights/TinkersConstruct/blob/1.19.2/build.gradle#L52). Our Jenkins server has been perfectly happy with that.

Please provide more information about why you think this change is needed. I assume you are just not setting up the workspace correctly.

----

If I had to guess, IDEA is not saying you need to add an additional project for those annotations. IDEA is just telling you it prefers its own annotations; you have to tell it we would rather use the built in ones. But thats not a build.gradle dependency, thats an IDEA profile setting. Show me the exact error message you got.

Goldenlion5648:
The output was 
image

Seems this was caused by having these two options set to intellij instead of gradle. (This was likely also the cause of IntelliJ telling me to use java 18 in the gradle output, or the other screenshot below) 
image

The gradle error might have been related to this setting
image

KnightMiner:
I was specifically asking about the error with nullable that prompted adding the dependency. It shouldn't be needed, but there is probably a better way to remove the warning.

PR #5270, opened 2024-10-18
Simplified Chinese Translation Update (1.19.2)

Cactusstudent:
Updating！


KnightMiner:
I'm going to go ahead and merge this into 1.19. Originally I was holding off in case I make other 1.19 changes, but its been long enough that I'd rather the PR be handled so someone else can help translate should we make other 1.19 changes.

Issue #5004, opened 2022-09-29
Incompatibility with Crafting Tweaks


Crystallyne:
### Minecraft Version

1.18.2

### Forge Version

40.1.68

### Mantle Version

Unknown

### Tinkers' Construct Version

3.5.2.40

### Describe your issue

[Crafting Tweaks](https://www.curseforge.com/minecraft/mc-mods/crafting-tweaks) adds buttons to the crafting window for common tasks. The Crafting Station for this version of TC does not show those buttons. They show up fine on standard crafting tables. I'm not sure which mod needs to be updated, so I'm reporting the issue on both mods.

### Other mods

[Crafting Tweaks](https://www.curseforge.com/minecraft/mc-mods/crafting-tweaks)

### Tried reproducing with just Tinkers?

No

### Performance Enchancers

None of the above

### Searched for known issues?

Searched open issues

KnightMiner:
Have you reported this to crafting tweaks? If so, link the issue here.

Crystallyne:
Yes, I posted a comment on the Crafting Tweaks page. You can find my post here: https://www.curseforge.com/minecraft/mc-mods/crafting-tweaks?comment=137

Edit: Didn't mean to mark this as closed.

KnightMiner:
A comment will get lost. Use the issues link for crafting tweaks to report a proper bug report on their github: https://github.com/ModdingForBlockheads/CraftingTweaks/issues

Looks like there is one, ModdingForBlockheads/CraftingTweaks#151

Crystallyne:
Oh perfect, thank you. Sorry, I didn't think to click the Issues button on the Crafting Tweaks page on Curse. Do you still need me to create a new post to report the issue? I would assume that one would suffice?

KnightMiner:
no reason to make duplicates

Crystallyne:
That's what I thought. Thank you for the help.

KnightMiner:
Crafting Tweaks support was added in #5277, which is in the latest 1.19.2 and 1.20.1 builds.

PR #5277, opened 2024-11-17
CraftingTweaks compatibility

paypur:
#5004

quick demo

https://github.com/user-attachments/assets/3ad5d33e-5dfc-4a5c-aa10-f7141fac9bda

### DIFF
--- a/src/main/java/slimeknights/tconstruct/plugin/craftingtweaks/TinkersCraftingGridProvider.java
+++ b/src/main/java/slimeknights/tconstruct/plugin/craftingtweaks/TinkersCraftingGridProvider.java
@@ -1,0 +1,32 @@
+package slimeknights.tconstruct.plugin.craftingtweaks;
+
+import net.blay09.mods.craftingtweaks.api.ButtonAlignment;
+import net.blay09.mods.craftingtweaks.api.CraftingGridBuilder;
+import net.blay09.mods.craftingtweaks.api.CraftingGridProvider;
+import net.minecraft.world.inventory.AbstractContainerMenu;
+import slimeknights.tconstruct.TConstruct;
+import slimeknights.tconstruct.tables.menu.CraftingStationContainerMenu;
+
+public class TinkersCraftingGridProvider implements CraftingGridProvider {
+
+  @Override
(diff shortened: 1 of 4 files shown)


src/main/java/slimeknights/tconstruct/plugin/craftingtweaks/TinkersCraftingGridProvider.java
```diff
@@ -0,0 +1,32 @@
...
+public class TinkersCraftingGridProvider implements CraftingGridProvider {
+
+  @Override
+  public String getModId() {
+    return TConstruct.MOD_ID;
+  }
+
+  @Override
+  public boolean handles(AbstractContainerMenu menu) {
+    return menu instanceof CraftingStationContainerMenu;
```
KnightMiner, line 19: Doesn't `menu` have a `getType()` method we can use instead of doing an `instanceof` check? Feels more appropriate to do so, and it would be faster.

paypur: Yes it does

KnightMiner: Turns out, crafting tweaks crashes if you use `getType` in this context on certain screens, guess not every menu supports that function (blame dumb mojang design). So guess the old code is required.

KnightMiner:
Tested and it works fine, though apparently my earlier suggestion is wrong due to how crafting tweaks works so will need to be reverted to what you had before. After that this should be ready to merge.

src/main/java/slimeknights/tconstruct/plugin/craftingtweaks/TinkersCraftingGridProvider.java
paypur: In which scenarios did it crash?

KnightMiner: Opening the inventory, might have been creative inventory specific

PR #5279, opened 2024-11-28
Render Type Registrations Move to Json

IACTU:
Removes all render type registrations for blocks from code and moves them into appropriate json model files.


KnightMiner:
Code looks good at a glance, though I'll want to test it in game before I merge it.

KnightMiner:
Looks like you missed the item forms of transparent glass panes, though probably faster if I just update those myself (probably will just toss them into datagen so I don't have to deal with 16 of them)

IACTU:
Mb, thanks for getting those

Issue #5281, opened 2024-12-17
Breaking changes for 1.20


KnightMiner:
To anyone discovering this issue looking for news about 1.20, see [the FAQ](https://slimeknights.github.io/faq/#what-versions-of-minecraft-do-you-support). This issue is simply a TODO list, not an indicator of update progress.

## API changes that can support migrating old behavior

These changes can be added without breaking addons.

* [ ] Look into using GameTest for automated tests, not sure how long our method will keep working as mojang changes bootstrap (plus world sensitive tests would be nice to have)
* [ ] Books: Remove all uses of named section transformers in favor of generic section data transformers
* [x] Replace tank block model property with a fluid and a capacity property
* [ ] Split castable tool casting from castable part swapping, allows us to do tool casting over a non-consumed cast and part swapping without a tool casting recipe.

### Tools

* [ ] Support global tool definition data loaders
* [ ] Support merging on tool definition data? May be redundant to above
* [ ] Migrate away from tool loot hook to standard global loot modifiers
* [ ] Consider - make tool building recipe contain parts list. Move part swapping logic to tool building recipe.
* [ ] Allow composable modifiers to merge modules from multiple data packs.

### Tools

* [x] Move modifier slot colors to Mantles resource color manager
* [x] Narrow return types on ToolStack methods to use NamespaceNBT instead of ModDataNBT when reasonable

### Modifiers

* [x] Migrate inventory modifier to modifier modules
* [x] Remove modifier serializers = use modifier modules with composable instead.
* [x] Remove typed check on dynamic modifiers, its no longer useful when every modifier is composable
* [x] Change projectile hit hook to use the new forge enum for canceling instead of a boolean.

### Recipes

* [x] Switch Tinker Station and Modifier Workstation to using LazyToolStack for results.

## New 1.20 Features

* [x] Duplicate armor trims in a casting table using molten diamond at a discount
* [x] Add trims to all relevant tinkers armors

KnightMiner:
Closing as this is all either completed or moved to the roadmap to complete as a non-breaking change or a 1.21 feature.

PR #5282, opened 2024-12-18
Change some recipes to use LazyToolStack for results

IACTU:
#5281

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
+++ b/src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
@@ -58,11 +58,11 @@ public interface ITinkerStationRecipe extends ICommonRecipe<ITinkerStationContai
 
   /**
    * Updates the input stacks upon crafting this recipe
-   * @param result  Result from {@link #assemble(ITinkerStationContainer)}. Generally should not be modified. TODO: switch parameter to LazyToolStack.
+   * @param result  Result from {@link #assemble(ITinkerStationContainer)}. Generally should not be modified.
    * @param inv     Inventory instance to modify inputs
-   * @param isServer  If true, this is on the serverside. Use to handle randomness, {@link IMutableTinkerStationContainer#giveItem(ItemStack)} should handle being called serverside only
+   * @param isServer  If true, this is on the serverside. Use to handle randomness, {@link IMutableTinkerStationContainer#giveItem(ItemStack)} should handle being called serverside onlys
    */
-  default void updateInputs(ItemStack result, IMutableTinkerStationContainer inv, boolean isServer) {
+  default void updateInputs(LazyToolStack result, IMutableTinkerStationContainer inv, boolean isServer) {
     // shrink all stacks by 1
     for (int index = 0; index < inv.getInputCount(); index++) {
       inv.shrinkInput(index, 1);
@@ -72,7 +72,7 @@ public interface ITinkerStationRecipe extends ICommonRecipe<ITinkerStationContai
 
   /* Deprecated */
 
-  /** @deprecated use {@link #updateInputs(ItemStack, IMutableTinkerStationContainer, boolean)} */
+  /** @deprecated use {@link #updateInputs(LazyToolStack, IMutableTinkerStationContainer, boolean)} */
   @Override
   @Deprecated
   default NonNullList<ItemStack> getRemainingItems(ITinkerStationContainer inv) {
--- a/src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
+++ b/src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
@@ -7,6 +7,7 @@ import net.minecraft.world.level.Level;
 import slimeknights.mantle.recipe.ICommonRecipe;
 import slimeknights.tconstruct.library.recipe.RecipeResult;
 import slimeknights.tconstruct.library.recipe.TinkerRecipeTypes;
+import slimeknights.tconstruct.library.tools.nbt.LazyToolStack;
 
 /**
  * Main interface for all recipes in the Tinker Station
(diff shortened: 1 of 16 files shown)


src/main/java/slimeknights/tconstruct/tables/block/entity/table/TinkerStationBlockEntity.java
```diff
@@ -132,9 +133,9 @@ public ItemStack calcResult(@Nullable Player player) {
         }
 
         // try for UI errors
-        RecipeResult<ItemStack> validatedResult = recipe.getValidatedResult(this.inventoryWrapper);
+        RecipeResult<LazyToolStack> validatedResult = recipe.getValidatedResult(this.inventoryWrapper);
         if (validatedResult.isSuccess()) {
-          result = validatedResult.getResult();
+          result = validatedResult.getResult().getStack();
```
KnightMiner: Resolving the lazy tool stack immediately defeats the purpose of using it in the first place.

Take a look at the modifier worktable block entity, that shows how to properly use this utility.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
```diff
@@ -40,12 +41,11 @@ default ItemStack assemble(ITinkerStationContainer inv) {
 
   /**
    * Gets the recipe result, or an object containing an error message if the recipe matches but cannot be applied.
-   * TODO 1.20: switch return type to {@code RecipeResult<LazyToolStack>}
    * @return Validated result
    */
-  default RecipeResult<ItemStack> getValidatedResult(ITinkerStationContainer inv) {
-    ItemStack result = assemble(inv);
-    if (result.isEmpty()) {
+  default RecipeResult<LazyToolStack> getValidatedResult(ITinkerStationContainer inv) {
```
KnightMiner, line 46: The key idea with the switch to LazyToolStack is we want to also pass it in instead of `ItemStack result` to `updateInputs` as that saves us needing to create the tool stack again just to figure out how to update inputs.

src/main/java/slimeknights/tconstruct/tables/recipe/TinkerStationPartSwapping.java
```diff
@@ -173,7 +174,8 @@ public RecipeResult<ItemStack> getValidatedResult(ITinkerStationContainer inv) {
           }
         }
         // everything worked, so good to go
-        return RecipeResult.success(tool.createStack(Math.min(inv.getTinkerableSize(), shrinkToolSlotBy())));
+        LazyToolStack lazyToolStack = LazyToolStack.from(tool, Math.min(inv.getTinkerableSize(), shrinkToolSlotBy()));
+        return RecipeResult.success(lazyToolStack);
```
KnightMiner: No need to split this over two lines, `LazyToolStack.from(tool, ` is not that much longer than `tool.createStack(`

KnightMiner:
Best to talk to me on discord before starting work on a feature from #5281; a few of them are better to do after the 1.20 port instead of before as neither API is significantly simpler than the other.

You have the right idea with this so far though and nothing stops us from doing this before the port. Just need to make sure you propagate usages of the lazy stack everywhere for it to be useful; take Modifier Worktable as a reference.

IACTU:
Yeah I just had a few minutes spare so I started on this. Very wip. Happy to coordinate on discord though if there's something else not too complicated which you'd prefer me to work on first.

KnightMiner:
This is fine right now, probably one of the more straight forward tasks on that list as its got a good reference. Mostly requires removing a lot of redundant `ToolStack.from` calls now that they can fetch the tool instance directly.

src/main/java/slimeknights/tconstruct/tables/block/entity/table/TinkerStationBlockEntity.java
IACTU: On account of the call later on to TooltipUtil.setDisplayName(ItemStack tool, String name) and the return type here, won't I have to resolve it in this method anyway? Otherwise, I can make a new method for LazyToolStack instead to do the same thing to the nbt but without changing the return type of calcResult I don't see the benefit

IACTU: Let me know your thoughts but the [commit](https://github.com/SlimeKnights/TinkersConstruct/pull/5282/commits/651ef6b42ce0687ef40979769fc0bec84ac7707b) I just pushed has that kind of idea. I can revert or adjust later if needed

src/main/java/slimeknights/tconstruct/library/tools/nbt/LazyToolStack.java
```diff
@@ -49,4 +50,13 @@ public ItemStack getStack() {
     }
     return stack;
   }
+
+  public void setDisplayName(String name) {
```
KnightMiner: This seems like a good approach, just make sure you get a javadoc in there.

Worth noting that both the stack and the tool share the same tag compound instance so setting either display name sets both.

src/main/java/slimeknights/tconstruct/library/recipe/tinkerstation/ITinkerStationRecipe.java
```diff
@@ -40,12 +41,11 @@ default ItemStack assemble(ITinkerStationContainer inv) {
...
+    LazyToolStack result = LazyToolStack.from(assemble(inv));
+    if (result.getStack().isEmpty()) {
       return RecipeResult.pass();
     }
     return RecipeResult.success(result);
```
KnightMiner: Probably better to do
```java
ItemStack result = assemble(inv);
if (result.isEmpty()) {
  return RecipeResult.pass()
}
return RecipeResult.success(LazyToolStack.from(result));
```
No sense creating a lazy tool stack if we are not returning it.

KnightMiner:
Looks pretty good overall, main thing that is missing is an update to the screen logic. Take a look at how the modifier worktable does it compared to the tinker station:

* https://github.com/SlimeKnights/TinkersConstruct/blob/1.19.2/src/main/java/slimeknights/tconstruct/tables/client/inventory/ModifierWorktableScreen.java#L193
* https://github.com/SlimeKnights/TinkersConstruct/blob/1.19.2/src/main/java/slimeknights/tconstruct/tables/client/inventory/ModifierWorktableScreen.java#L193

src/main/java/slimeknights/tconstruct/library/tools/nbt/LazyToolStack.java
```diff
@@ -34,6 +34,7 @@ public static LazyToolStack from(ToolStack tool) {
   }
 
   /** Gets the tool for this instance */
+  @NotNull
```
KnightMiner: I annotate methods as `Nonnull` at  a package level, you shouldn't need to set it on individual methods.

IACTU: Ah, for some reason Intellij didn't seem to recognize that. I'll undo that tomorrow

KnightMiner: You have to tell IntelliJ that I like javax annotations, just search `Nonnull` in settings, should be under Inspection > Configure Annotations. There is also a chance the package is missing an annotation (as it has to be added to every single package manually)

IACTU: The package is annotated correctly

PR #5308, opened 2025-01-13
New textures for ancient tools, fluid cannon, tough collar and a few other things

RCXcrafter:
As discussed on discord, here are the new textures for:
- Swasher
- Fluid Cannon

and edited textures for:
- Melting Pan
- Tough Collar
- Vein Hammer grip
- Slimewood Large Plates (the slime face got lost somewhere in an update so I added it back)
- Slime metal guards for swords
- Plate Armor items (small shading adjustments)

The swasher has ancient debris barrel textures as well as grayscale ones since I don't think it has been decided yet if the swasher will have two or three materials.

I couldn't quite figure out how you have the datagen set up for the charged states of tools so I didn't implement those and just included textures for them.

Screenshots:

output
image

image


src/main/java/slimeknights/tconstruct/library/data/AbstractToolItemModelProvider.java
```diff
@@ -91,6 +91,22 @@ protected void bow(IdAwareObject bow, JsonObject properties, boolean crossbow, S
     }
   }
 
+  /** Creates models for blocking, broken and charged for the given tool */
```
KnightMiner, line 94: nitpick, but worth a quick note on what "charged" means. Maybe just say "fully charged" to be more clear?

src/main/resources/assets/tconstruct/models/block/fluid_cannon_up.json
```diff
@@ -0,0 +1,67 @@
+{
+  "render_type": "minecraft:cutout",
```
KnightMiner, line 2: You know, I'm pretty sure this and the other fluid cannons uses the same model as the melter, save different textures. Recent Mantle reworks means the melter no longer lists items inside so its just a generic tank.

We can probably make a template for all three cannon directions and the melter to inherit from.

src/main/resources/assets/tconstruct/textures/item/tool/melting_pan/overlay_charged.png.mcmeta
KnightMiner, line 1: Isn't `overlay_charged.png` 5 frames? This animation only includes 4

KnightMiner:
Overall looks good apart from some nitpicks; can always handle the more minor of those post merge if needed.

I think the one thing I'd want that is missing is a fluid cannon for the nether, made of scorched stone and either gold or cobalt (whichever looks better). After thinking it over I am liking copper for that; slimesteel really only is needed for the upsidedown channel plans (which may benefit from the weird slimesteel block model)

KnightMiner:
I'm going to go ahead and merge this, as a couple of changes I am working on would conflict with this. Fluid cannon is not top priority so we can always add the nether textures later.

PR #5338, opened 2025-02-07
translate spanish

SrNadien:


Manwe03:
"pig iron" sould be translated to  "arrabio"

PR #5344, opened 2025-02-15
Update download link in README.md

zax71:


KnightMiner:
CurseForge team has been way more helpful over the years than Modrinth, notably they don't fire all their staff randomly. I trust it to stick around far longer than Modrinth will.

If you wanted a useful "update link" you'd link our projects site, but I wouldn't accept a PR for a minor readme change.

zax71:
The Previous link to CF was not valid - linked to the CF homepage instead of your page

KnightMiner:
Then why are you replacing it with a Modrinth link? A zero description PR that replaced the link with no mention of it being broken looks like to have a vendetta against CurseForge

PR #5351, opened 2025-02-26
Extract common code from bow and crossbow

lcy0x1:
- `BowAmmoModifierHook`: Change input parameter type from `Player` to `LivingEntity`

Restructure arrow shoot logic:
- User information (For player with bow: speed = 3, inaccuracy = 1, no infinity, do weapon durability cost)
- Shoot logic supporting multishot (common code)
  - Arrow creation logic. Use inheritance to allow crossbow to support rocket and mark arrow as shot from crossbow
  - Arrow modifier logic. Same for bow and crossbow.
  - Firing logic. This part bow differs from crossbow.
  - Shooting sound. Use inheritance to allow bow and crossbow to customize.
  - Durability cost. Same for bow and crossbow
- Individual code for weapon data manipulation


lcy0x1:
This PR is the first step toward allowing mob to use TConstruct bows and arrows. They can call `ModifiableLauncherWeapon.shootProjectiles` to simulate shooting.

src/main/java/slimeknights/tconstruct/library/tools/item/ranged/LauncherUserInfo.java
```diff
@@ -0,0 +1,58 @@
+package slimeknights.tconstruct.library.tools.item.ranged;
+
+import net.minecraft.world.entity.LivingEntity;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.level.Level;
+
+public interface LauncherUserInfo {
```
KnightMiner: This interface feels unnecessary; its just a bunch of getters. Any reason it can't just be parameters for the helper instead?

src/main/java/slimeknights/tconstruct/library/tools/item/ranged/ModifiableLauncherItem.java
```diff
@@ -372,8 +388,89 @@ public boolean onBlockStartBreak(ItemStack stack, BlockPos pos, Player player) {
...
+    }
+  }
+
+  public ProjectileData createProjectile(Level level, LivingEntity user, ItemStack ammo) {
+    ArrowItem arrowItem = ammo.getItem() instanceof ArrowItem a ? a : (ArrowItem) Items.ARROW;
+    AbstractArrow arrow = arrowItem.createArrow(level, ammo, user);
+    return new ProjectileData(arrow, 1f, 1);
+  }
+
+  public abstract Vector3f modifyShootAngle(LivingEntity user, Vec3 target, float angle);
```
KnightMiner, line 452: Every public or protected method must have either `@Override` or a JavaDoc.

src/main/java/slimeknights/tconstruct/library/modifiers/hook/ranged/BowAmmoModifierHook.java
```diff
@@ -183,7 +183,10 @@ static ItemStack findAmmo(IToolStackView tool, ItemStack bow, Player player, Pre
       if (needed > standardAmmo.getCount()) {
         // consume the whole stack
         resultStack.grow(standardAmmo.getCount());
-        player.getInventory().removeItem(standardAmmo);
+        standardAmmo.setCount(0);
+        if (user instanceof Player player) {
+          player.getInventory().removeItem(standardAmmo);
+        }
```
KnightMiner: I want to discuss this more; how is a non-player expected to remove the ammo from their inventory? Or are we just assuming they don't care about cleaning up empty stacks?

I know most mobs do infinite ammo, but imagine a dispenser like shooter.

src/main/java/slimeknights/tconstruct/library/modifiers/hook/ranged/BowAmmoModifierHook.java
```diff
@@ -100,21 +100,21 @@ private static ItemStack findMatchingAmmo(ItemStack bow, LivingEntity living, Pr
    * @param tool       Tool instance
    * @param bow        Bow stack instance
    * @param predicate  Predicate for valid ammo
-   * @param player     Player to search
+   * @param user       User to search
    * @return  Found ammo
    */
-  static ItemStack findAmmo(IToolStackView tool, ItemStack bow, Player player, Predicate<ItemStack> predicate) {
+  static ItemStack findAmmo(IToolStackView tool, ItemStack bow, LivingEntity user, Predicate<ItemStack> predicate) {
```
KnightMiner: Usually I call the entity using a tool `holder`, since it makes their role more clear in the grand scheme of tool usage.

src/main/java/slimeknights/tconstruct/library/modifiers/hook/ranged/BowAmmoModifierHook.java
lcy0x1: I would say the mob needs to be able to clean up empty stacks themselves. What harm will it do to not remove empty stacks? They are empty and `isEmpty()` returns true

src/main/java/slimeknights/tconstruct/library/tools/item/ranged/LauncherUserInfo.java
lcy0x1: It makes user side easier. Imagine one has a mob that should support both bow and crossbow. By making a record for the mob implementing this interface, the code becomes cleaner and more intuitive.

lcy0x1: Actually I feel like direction should be part of LauncherUserInfo as well: it's viewVector for player, and some calculated projection vector for mobs, based on target position

src/main/java/slimeknights/tconstruct/library/modifiers/hook/ranged/BowAmmoModifierHook.java
KnightMiner: Mostly just memory waste, which is why vanilla typically does it.

src/main/java/slimeknights/tconstruct/library/tools/item/ranged/LauncherUserInfo.java
KnightMiner: > It makes user side easier. Imagine one has a mob that should support both bow and crossbow. By making a record for the mob implementing this interface, the code becomes cleaner and more intuitive.

I disagree. If your mob does both, just make a helper method that call our method. We do it all the time in our API. And its just 4 values. Make a method for players to call that automatically sets the properties, and make a general method that other mods can call, making their own helper if they wish. Much cleaner than taking an single purpose interface they are expected to implement for a bunch of getters.

src/main/java/slimeknights/tconstruct/library/modifiers/hook/ranged/BowAmmoModifierHook.java
lcy0x1: Completed

src/main/java/slimeknights/tconstruct/library/tools/item/ranged/ModifiableLauncherItem.java
lcy0x1: Completed

src/main/java/slimeknights/tconstruct/library/tools/item/ranged/LauncherUserInfo.java
lcy0x1: I will change it then.

src/main/java/slimeknights/tconstruct/library/modifiers/hook/ranged/BowAmmoModifierHook.java
lcy0x1: I don't think there is a realistic method to clear mob inventory, as each mob would have different implementations. Sounds like it doesn't matter that much.

Though if you think this part should be part of the API, then maybe `LauncherUserInfo` would be a place to add it.

src/main/java/slimeknights/tconstruct/library/tools/item/ranged/ModifiableLauncherItem.java
```diff
@@ -386,8 +402,99 @@ public boolean onBlockStartBreak(ItemStack stack, BlockPos pos, Player player) {
...
+      projectile.shoot(targetVector.x(), targetVector.y(), targetVector.z(),
+        charge * velocity * data.speed() * speedFactor,
+        inaccuracy * inaccuracyFactor);
+      // finally, fire the projectile
+      holder.level().addFreshEntity(projectile);
+      playShotSound(holder, charge, angle, holder.getRandom());
+    }
+    if (damageWeapon) {
+      ToolDamageUtil.damageAnimated(tool, damage, holder, hand);
+    }
```
KnightMiner: Not sure on this point, but I do wonder whether it wouldn't be simpler to just `return damage` and let the caller handle it/ignore it, lets us drop 2 parameters from the method as hand is only needed for this method. Also gives me freedom to have say, a helmet shoot arrows with the helmet keybind.

lcy0x1: Maybe I can do, `shootProjectiles` and `shootProjectilesAndDamageWeapon`?

KnightMiner: Its not hard to write:
```java
int damage = shootProjectiles(...);
if (!creative) {
  ToolDamageUtil.damageAnimated(...);
}
```
so I think just `shootProjectiles` is sufficient

lcy0x1: It fool-proves addons who would have called `stack.hurtAndBreak`

KnightMiner: Hurt and break is fine for them to call, its just less efficient as it requires reparsing NBT.

And to be blunt, if an addon calling *Tinkers' Construct code* doesn't realize they should call *Tinkers' Construct methods* while using an internal API, there is not much I can do for them.

Point of the original comment was to minimize complexity of the API by separating responsibilities, not make it more complex by adding similar but slightly different methods.

lcy0x1: ok. I will remove the redundant method

lcy0x1:
Any remaining items to discuss?

KnightMiner:
I need to do a more in depth code review when I get a chance. I might have something after that.

lcy0x1:
Thanks

KnightMiner:
Apologizes, between the changes I pushed and the more recent update I spent most of the year on, this PR has gotten buried.

Is there anything notable missing from the living entity support for bows and crossbows?

I'm still not sure that firing the two should be sharing that much code since they work so differently.

PR #5381, opened 2025-04-13
Casting Tank

IACTU:
Adds the casting tank as described by: https://slimeknights.github.io/docs/roadmap/#casting-tank
Currently I haven't added a page in the book, but I could do that too if desired.


KnightMiner:
Can you include a few screenshots in the PR description? Its helpful for posterity, plus makes the pre-testing phase of the code review more fun.

src/main/java/slimeknights/tconstruct/smeltery/TinkerSmeltery.java
```diff
@@ -237,6 +239,7 @@ public final class TinkerSmeltery extends TinkerModule {
   public static final ItemObject<ChannelBlock> searedChannel = BLOCKS.register("seared_channel", () -> new ChannelBlock(SEARED_NON_SOLID), TOOLTIP_BLOCK_ITEM);
   public static final ItemObject<CastingBasinBlock> searedBasin = BLOCKS.register("seared_basin", () -> new CastingBasinBlock(SEARED_NON_SOLID, false), TOOLTIP_BLOCK_ITEM);
   public static final ItemObject<CastingTableBlock> searedTable = BLOCKS.register("seared_table", () -> new CastingTableBlock(SEARED_NON_SOLID, false), TOOLTIP_BLOCK_ITEM);
+  public static final ItemObject<CastingTankBlock> castingTank = BLOCKS.register("casting_tank", () -> new CastingTankBlock(SEARED_NON_SOLID, PushReaction.DESTROY), b -> new TankItem(b, ITEM_PROPS, true));
```
KnightMiner: Naming convention for the fields would make this `seared_casting_tank`/`searedCastingTank`. Not because we are definitely adding it, but to avoid needing to rename if we do.

Has the bonus of the block entity field can be just `castingTank`.

src/main/java/slimeknights/tconstruct/smeltery/data/SmelteryRecipeProvider.java
```diff
@@ -313,6 +313,15 @@ private void addSmelteryRecipes(Consumer<FinishedRecipe> consumer) {
                        .pattern("BBB")
                        .unlockedBy("has_item", has(TinkerSmeltery.searedBrick))
                        .save(consumer, location(folder + "lantern"));
+    ShapedRecipeBuilder.shaped(RecipeCategory.REDSTONE, TinkerSmeltery.castingTank.get())
+                       .define('#', TinkerSmeltery.searedBrick)
+                       .define('B', Tags.Items.GLASS)
+                       .define('C', Tags.Items.INGOTS_COPPER)
+                       .pattern("#C#")
+                       .pattern("BBB")
+                       .pattern("#B#")
```
KnightMiner: Lets make the recipe mirror the drain recipe a bit more, cross the drain and tank recipe. So

```
BGB
CGC
BGB
```
where `B` is bricks, `G` is glass, and `C` is copper ingots.

src/main/java/slimeknights/tconstruct/smeltery/block/CastingTankBlock.java
```diff
@@ -0,0 +1,151 @@
...
+import slimeknights.tconstruct.library.utils.NBTTags;
+import slimeknights.tconstruct.smeltery.block.entity.CastingTankBlockEntity;
+import slimeknights.tconstruct.smeltery.block.entity.CastingTankBlockEntity.ITankBlock;
+import slimeknights.tconstruct.smeltery.block.entity.ITankBlockEntity;
+
+import javax.annotation.Nullable;
+
+public class CastingTankBlock extends InventoryBlock implements ITankBlock, EntityBlock {
+  protected static final DirectionProperty FACING = BlockStateProperties.HORIZONTAL_FACING;
+  private final PushReaction pushReaction;
```
KnightMiner: The push reaction field is only used in tanks because we vary it. For this, just override the getter; no need for constructor parameter.

Though tbh, I have nothing against it being pushed, just let the vanilla default prevent it due to block entity (so mods like quark can still push it)

src/main/java/slimeknights/tconstruct/smeltery/block/CastingTankBlock.java
```diff
@@ -0,0 +1,151 @@
...
+  @Deprecated
+  @Override
+  public InteractionResult use(BlockState state, Level world, BlockPos pos, Player player, InteractionHand hand, BlockHitResult hit) {
+    if (player.isSuppressingBounce()) {
+      return InteractionResult.PASS;
+    }
+
+    BlockEntity te = world.getBlockEntity(pos);
+    if (te instanceof CastingTankBlockEntity) {
+      ((CastingTankBlockEntity) te).interact(player, hand, hit);
```
KnightMiner: use pattern matching instance of.

```java
if (te instanceof CastingTankBlockEntity castingTank) {
  castingTank.interact(player, hand, hit);
}
``

File: src/main/java/slimeknights/tconstruct/smeltery/block/CastingTankBlock.java [hunk 5]
```diff
@@ -0,0 +1,151 @@
...
+      return fluid.getFluid().getFluidType().getLightLevel(fluid);
+    }
+    return super.getLightEmission(state, world, pos);
+  }
+
+  @Override
+  public void setPlacedBy(Level worldIn, BlockPos pos, BlockState state, @Nullable LivingEntity placer, ItemStack stack) {
+    CompoundTag nbt = stack.getTag();
+    if (nbt != null) {
+      BlockEntityHelper.get(CastingTankBlockEntity.class, worldIn, pos).ifPresent(te -> te.updateTank(nbt.getCompound(NBTTags.TANK)));
```
KnightMiner: I'm inclined to just ditch `BlockEntityHelper` from our code, just do the standard if statement like the other methods; pattern matching makes it pretty clean:

```java
if (world.getBlockEntity(pos) instanceof CastingTankBlockEntity tank) {
  tank.updateTank(nbt.getCompound(NBTTags.TANK));
}
```

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,370 @@
...
+   * Gets the capacity for the given block
+   * @param block  block
+   * @return  Capacity
+   */
+  public static int getCapacity(Block block) {
+    if (block instanceof ITankBlock) {
+      return ((ITankBlock) block).getCapacity();
+    }
+    return DEFAULT_CAPACITY;
+  }
```
KnightMiner: We probably don't need to do the whole variable capacity thing; it just exists for the tank variants like lantern and ingot tanks. For the casting tank I don't plan to have variants, so probably just hardcode to 4000mb for now. Can always abstract it later if needed, but better than adding in abstractions for unknown future plans.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,370 @@
...
+    }
+
+    ItemStack input = getItem(INPUT);
+    ItemStack output = getItem(OUTPUT);
+    ItemStack held = player.getItemInHand(hand);
+    // if the held item can be placed inside, do so
+    if (canPlaceItem(INPUT, held)) {
+      setItem(INPUT, held.split(1));
+    // otherwise, interact with the tank
+    } else if (!FluidTransferHelper.interactWithTank(level, worldPosition, player, hand, hit)) {
```
KnightMiner: This is where the casting tank behavior gets a little funny. The items we might want to place inside happen to be the same items where the fluid transfer helper does anything useful. Ultimately, the casting tank is designed to be a block for automation; using it manually is going to just be more awkward than clicking a tank.

I think the best interaction would be:

* If there is an item in the output slot, remove it.
* If there is no item in the output, try dumping the item into the tank/filling it from the tank
* If interaction fails, then place the item inside.

Most of the time, users will be using this with automation, not manually. Assuming they are using it manually, we can save 3 steps by filling the item in their hand directly when possible.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,370 @@
...
+    return null;
+  }
+
+  /** Interface for blocks to return their capacity */
+  public interface ITankBlock {
+    /** Gets the capacity for this tank */
+    int getCapacity();
+  }
+
+  // TODO consider moving this to Mantle
```
KnightMiner, line 306: I'll look at it more in depth later, but is the only difference the transfer results here?

I do wonder about the alternative approach; can we get away with just playing the sound inside the method? Would be easier to merge into Mantle then; just a boolean "play sound" parameter that is false in the existing method (which is now an overload). Could also do it via a nullable level, pass in null for no sound.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,370 @@
...
+  /**
+   * Attempts to transfer fluid from the passed stack into a tank.
+   * @param teHandler  Tank handler
+   * @param stack      Input stack, may be modified
+   * @param direction  Determines whether we may empty the item, fill, or both
+   * @return  Transfer result from attempt, or null if no transfer happened
+   */
+  public static @Nullable IFluidContainerTransfer.TransferResult interactWithTankSlotForTransferResult(IFluidHandler teHandler, ItemStack stack, IFluidContainerTransfer.TransferDirection direction) {
```
KnightMiner: I put `@Nullable` on the line before as I use `javax` annotations.

KnightMiner:
I have not tested this yet, but a quick glance at the code looks pretty good apart from some comments. When I have more time I'll do an in game test and leave any feedback or mention if anything seems broken.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
IACTU: Yeah I tried to preserve the exact logic of Mantle's interactWithTankSlot(). Obviously up to you about how you want the sound played, but if there's any other sort of feedback you'd ever use that depends on the flow direction, I'd say returning a TransferResult like I currently have enables that the best.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
IACTU: With regard to step 2, would you say an empty bucket should be placeable in the input slot of an empty casting tank? Mantle's FluidTransferHelper.interactWithTank() returns true even if no fluid actually was moved, but I can work around if if that functionality is desired.

KnightMiner: Ah right, its setup for block interaction behavior where the client side doesn't know whether anything happened.

To be fully honest, its not that important that you can place items in the tank slot, its really made for hoppers. But if there is an easy workaround, lets do that. Alternatively, if we have to add a new boolean for whether to return true on "has valid fluid container" vs returning true on "moved fluid", we could.

IACTU: I won't resolve this for now, but I did make the logic change in the last commit when you have a chance to review

KnightMiner:
Some of the recent commits added in a new set of fluid helpers in mantle that make it easier to ask "did this actually transfer" question. Also had to make a change to render items a bit to make the fluid cannon work.

Should be able to take a look at this soon now that fluid cannon and the cake revision is done.

----

One thing I did for fluid cannon we should probably do here though is i just based dump item into tank vs place item into tank on clicking the top or bottom half of the block. This block is setup well to do the same, top half place item, bottom half fill fluid.

IACTU:
Sure. I'll be busy the next few days but next week I'll have some time to integrate those changes and fix the conflicts.

src/main/java/slimeknights/tconstruct/smeltery/data/SmelteryRecipeProvider.java
```diff
@@ -519,6 +528,10 @@ private void addSmelteryRecipes(Consumer<FinishedRecipe> consumer) {
                         .addByproduct(TinkerFluids.moltenGlass.result(FluidValues.GLASS_PANE))
                         .addByproduct(TinkerFluids.moltenIron.result(FluidValues.INGOT / 3))
                         .save(consumer, location(meltingFolder + "lantern"));
+    MeltingRecipeBuilder.melting(NoContainerIngredient.of(TinkerSmeltery.searedCastingTank), TinkerFluids.moltenCopper, FluidValues.INGOT, 2.5f)
+                        .addByproduct(TinkerFluids.searedStone.result(FluidValues.BRICK * 4))
+                        .addByproduct(TinkerFluids.moltenGlass.result(FluidValues.GLASS_BLOCK * 4))
```
KnightMiner: This should be 2 ingots copper, 3 blocks glass now, since the above recipe changed.

src/main/java/slimeknights/tconstruct/common/data/render/RenderItemProvider.java
```diff
@@ -92,6 +92,11 @@ protected void addEntries() {
       itemBuilder.center( 4, 12,  4).build(),
       itemBuilder.center(12, 12,  4).build()
     ));
+    // casting tank
+    itemBuilder = RenderItem.builder().center(8, 15.5f, 8).size(14).x(270).y(180).transform(TinkerItemDisplays.CASTING_TABLE);
+    entry(TinkerSmeltery.searedCastingTank, List.of(
+      itemBuilder.build()
```
KnightMiner: This should have two items in here, just render both items overlapping like the casting table. Practically it should never happen, but it lets you reuse more existing code.

I believe the new `TankInventoryBlockEntityRenderer` would be perfect for your needs, and you could probably also use the casting table template here instead of copying the item list. Though the item handler instance in that class might be hard to get outside a capability call, so if its too much effort don't worry about code reuse.

src/main/java/slimeknights/tconstruct/common/data/tags/BlockTagProvider.java
```diff
@@ -503,6 +503,7 @@ private void addHarvest() {
     // melting pan blacklist, basically anything that feels gross due to unsupported melting recipe
     tagBlocks(MINEABLE_MELTING_BLACKLIST, TinkerSmeltery.searedMelter, TinkerSmeltery.smelteryController, TinkerSmeltery.foundryController, TinkerSmeltery.searedLantern, TinkerSmeltery.scorchedLantern, TinkerSmeltery.searedFluidCannon, TinkerSmeltery.scorchedFluidCannon);
     tagBlocks(MINEABLE_MELTING_BLACKLIST, TinkerSmeltery.searedTank, TinkerSmeltery.scorchedTank);
+    tagBlocks(MINEABLE_MELTING_BLACKLIST, TinkerSmeltery.searedCastingTank);
```
KnightMiner: Just add this to the end of the line with scorched fluid cannon. Reason there are two calls is the first is over standard `ItemObject` while the second is over `EnumObject`

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,367 @@
...
+   * Gets the capacity for the given item
+   * @param item  item
+   * @return  Capacity
+   */
+  public static int getCapacity(Item item) {
+    return DEFAULT_CAPACITY;
+  }
+
+  public CastingTankBlockEntity(BlockPos pos, BlockState state) {
+    this(pos, state, (ITankBlock) state.getBlock());
```
KnightMiner: No need for an unchecked cast, just do what I did this in `TankBlockEntity`.

Sure, it shouldn't happen, but I've seen people do hacky stuff with block entities giving them the wrong block.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,367 @@
...
+    super.setItem(slot, itemstack);
+    // if we are setting the input item and it actually changed, try to process it
+    if (slot == INPUT && !ItemStack.matches(currentInput, itemstack)) {
+      tryToProcessItem();
+    }
+  }
+
+  public void setInputItemWithoutTryingToProcess(ItemStack stack) {
+    super.setItem(INPUT, stack);
+  }
```
KnightMiner: Just make this `private void `setInputItem`; avoids risk of public calls to it.

If there is a reason you need public, give it a shorter name and a java doc explaining why it exists; probably best to mark it `@Internal` in that case.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,367 @@
...
+
+  @Override
+  public boolean canPlaceItem(int pIndex, ItemStack pStack) {
+    if (pIndex == INPUT) {
+      return getItem(INPUT).isEmpty() && getItem(OUTPUT).isEmpty() && !pStack.isEmpty() && (
+        // check the various options for some sort of fluid-containing stack
+        FluidContainerTransferManager.INSTANCE.mayHaveTransfer(pStack)
+          || pStack.getCapability(ForgeCapabilities.FLUID_HANDLER_ITEM).isPresent()
+          || pStack.getItem() instanceof BucketItem
```
KnightMiner: Lets ditch bucket items from this list; I'd rather not deal with spawning fish in this block.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,367 @@
...
+      }
+    }
+  }
+
+  @Override
+  public void setItem(int slot, ItemStack itemstack) {
+    ItemStack currentInput = getItem(INPUT);
+    super.setItem(slot, itemstack);
+    // if we are setting the input item and it actually changed, try to process it
+    if (slot == INPUT && !ItemStack.matches(currentInput, itemstack)) {
```
KnightMiner: Since we can't transfer empty, validate that the new stack is not empty before checking matches, since its way faster than item NBT comparison.

For readability, I normally call the parameter `newStack` and the original `oldStack` (or in this case, `oldInput`), as "current" changes mid method.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
KnightMiner: So, in regards to this transfer logic, the new method you will want is `FluidTransferHelper#interactWithContainer`. Don't bother with `BucketItem`, fish buckets are not something we need to automatically empty; let dispensers do that. Method will automatically play sounds for you.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
```diff
@@ -0,0 +1,367 @@
...
+    }
+  }
+
+  @Override
+  public AbstractContainerMenu createMenu(int pContainerId, Inventory pPlayerInventory, Player pPlayer) {
+    return null;
+  }
+
+  /** Interface for blocks to return their capacity */
+  public interface ITankBlock {
```
KnightMiner: This seems redundant, just reuse the one from `TankBlockEntity`

src/main/resources/assets/tconstruct/lang/en_us.json
```diff
@@ -1310,6 +1310,8 @@
   "block.tconstruct.seared_chute": "Seared Chute",
   "block.tconstruct.seared_chute.tooltip": "Smeltery item input and output",
 
+  "block.tconstruct.seared_casting_tank": "Casting Tank",
+  "block.tconstruct.seared_casting_tank.tooltip": "Automatically empties or fills an item.\nRetains liquid when broken.",
```
KnightMiner: > Automatically empties or fills inserted items.

Make it clear the purpose of this is to insert items over and over again.

KnightMiner:
I just pushed some commits related to the Proxy Tank (called tinker tank on the roadmap). There probably will not be much overlap, but you might find that a good reference for some of the weirder fluid interactions.

Feel free to either write the book page or leave it unwritten. If you write it, it would go in puny melting at the end of the casting section.

----

Since proxy tank was fresh in my mind I took another pass at the code and left a few comments related to how the two behaviors are the same.

After testing a few variants, I think the fluid cannon style of interaction is definitely the best choice here: interact with item if you click the top half, and tank on bottom half (though it might not be exactly half, I think your model was 7 pixels to 9 or something like that). The conditional "place item if failed to transfer" makes a bit less sense in this context as we have two reasonable transfer targets.

KnightMiner:
Did a quick in game test and this looks pretty good apart from the remaining comments. I'll leave some followup on the remaining ones with some additional specifics

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
KnightMiner: Did some testing, and I think the best behavior is to just go off location clicked like I did with the fluid cannon. So in the block class:
```java
  public InteractionResult use(BlockState state, Level world, BlockPos pos, Player player, InteractionHand hand, BlockHitResult hit) {
    if (world.getBlockEntity(pos) instanceof CastingTankBlockEntity tank) {
      tank.interact(player, hand, hit.getLocation().y - pos.getY() < 0.6875);
      return InteractionResult.SUCCESS;
    }
    return InteractionResult.FAIL;
  }
```
and in the block entity:

```java
 public void interact(Player player, InteractionHand hand, boolean clickedTank) {
    // skip client side
    if (level == null || level.isClientSide) {
      return;
    }

    // clicked tank? move fluid
    if (clickedTank) {
      if (!FluidTransferHelper.interactWithContainer(level, worldPosition, tank, player, hand).didTransfer()) {
        FluidTransferHelper.interactWithFilledBucket(level, worldPosition, tank, player, hand, getBlockState().getValue(FACING));
      }
    } else {
      // did not click tank? swap items around
      ItemStack input = getItem(INPUT);
      ItemStack output = getItem(OUTPUT);
      ItemStack held = player.getItemInHand(hand);
      // if there is an item in the output slot, take it
      if (!output.isEmpty()) {
        setItem(OUTPUT, ItemStack.EMPTY);
        ItemHandlerHelper.giveItemToPlayer(player, output, player.getInventory().selected);
      // next try to take the item from the input slot
      } else if (!input.isEmpty()) {
        setItem(INPUT, ItemStack.EMPTY);
        ItemHandlerHelper.giveItemToPlayer(player, input, player.getInventory().selected);
      // if no item in the tank, try to place a held item in the input
      } else if (!held.isEmpty() && canPlaceItem(INPUT, held)) {
        setItem(INPUT, held.split(1));
      }
    }
  }
```

src/main/java/slimeknights/tconstruct/common/data/render/RenderItemProvider.java
KnightMiner: Took a look at this and I believe that `TankInventoryBlockEntityRenderer` works perfectly. Just swap from `ITankBlockEntity` to `ITankInventoryBlockEntity` on the block entity and you have everything you need for that renderer (Mantle already implements the additional method added by the new interface).

Requires having a separate renderer for each item, but we want that anyways. If there is an item in both slots render both.

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
IACTU: Here, I need a method that doesn't require a player since this is potentially automated. Unless you're intending for fake players to be used or something.

src/main/java/slimeknights/tconstruct/common/data/render/RenderItemProvider.java
IACTU: I can modify the renderer I made to potentially render the two items, but I don't think TankInventoryBlockEntityRenderer is exactly right due the tankFluidModel config option switching to a render which doesn't match the casting table's tank shape

src/main/java/slimeknights/tconstruct/smeltery/block/entity/CastingTankBlockEntity.java
KnightMiner: Didn't consider that; just leave it for now and I'll merge something into mantle after the PR.

Move any of the areas with player access to the new methods.

src/main/java/slimeknights/tconstruct/common/data/render/RenderItemProvider.java
KnightMiner: That probably means you have a misconfigured block model. Its unfortunately a little redundant, but that runs off the block model JSON instead of Mantles fluid cuboid list.

Maybe I should redirect the former to the latter, but thats a 1.21 change.

Edit: found it here: https://github.com/SlimeKnights/TinkersConstruct/pull/5381/files#diff-cb09ad9cbf564afae4a878dda2ad65c467584b992cd20af43bfa81d1547b1100R147-R151

PR #5398, opened 2025-05-15
Cobalt texture updates

moon-st0ne:
- New shape and purplish color palette for Raw Cobalt
- New purplish color palette for Raw Cobalt Block
- New shape for Cobalt Nugget

image


moon-st0ne:
image

Altered raw ore block

PR #5399, opened 2025-05-18
French Translation for 1.20.1

XtraBe:


Mrthomas20121 (changes requested):
your translation is very inconsistent like why do translate skyslime but not enderslime or blazewood? either translate or don't. don't do both.

and also king lemming is a person. not something you translate

votre traduction est pas très consistent, pourquoi traduire skyslime et pas enderslime ou blazewood? soit vous les traduisers tous les matériels, soit vous les traduisers pas, pas les deux. 

ah oui aussi King Lemming c'est un Nom Propre, tu n'es pas senser le traduire

XtraBe:
I am deleting this pull request in favor of pull request #5401, which consolidates the translation of books in the code. This is also to ensure better consistency between the invented words used in the game and those that are too close to real-world terms.

I am using a glossary in the new version, so I will also revise this single fr_fr file accordingly.

PR #5401, opened 2025-05-21
French Translation for 1.20.1

XtraBe:
Change comit #5399 for this with book transaltion.


XtraBe:
add fr_fr.json

XtraBe:
Final update of FR translation txt.

PR #5411, opened 2025-06-02
minor bug in json file fr_fr (, or space)

XtraBe:


KnightMiner:
I would recommend at some point deleting your 1.20.1 branch then repulling our branch so that future PRs do not show all the commits that are already merged.

PR #5450, opened 2025-07-04
Simplified Chinese Translation Update (1.20.1)

Don-Trueno:
This pull request contains translation to features since 77696af7d982ee72d4fb31bdc0435289c0110737.

Any reviews and suggestions are welcomed.


KnightMiner:
Going to mention @Cactusstudent and @ChuijkYahus on this as the two of them have done the most work on the Chinese Translations. I recommend communicating with them if you aren't already to ensure there is not duplicate work.

Cactusstudent:
Thanks for reminding. I reviewed this Pull Request last week, and most of translation is great. Also made some correction based on our translation.
In fact, I haven't maintained the Simplified Chinese transaltion over the past month due to some reasons. So, probably @Don-Trueno will take my place.
The translation he made was in good quality, which has both consistency and accuracy. I'll also give assistance to keep the translation quality.

KnightMiner:
Going to go ahead and merge this as I am pushing for an update. Though as one comment based on a commit description:

The material "ancient" is meant to be "ancient material". It is repaired with ancient debris but not crafted with it. The lore description is we just don't know what its made from or how to quite make it, so its just described as old.

Its name is really an adjective rather than a noun as I liked how it combined with tool names, e.g. "Ancient Pickaxe" or "Ancient Battle Sign". If that doesn't work in Chinese though then try to find an appropriate noun for "old thing" I guess.

PR #5459, opened 2025-07-19
Simplified Chinese Translation Update (1.20.1)

Don-Trueno:
Mainly to keep step with the newest release, since commit 428701e9e7973a9dbcb55552bc41adcb90d1caae.

Any reviews and suggestions are welcomed.


src/main/resources/assets/tconstruct/book/tinkers_gadgetry/zh_cn/exploration/knightmetal_shard.json
```diff
@@ -0,0 +1,10 @@
+{
+  "title": "Knightmetal Shards",
```
KnightMiner: Title should be translated too, use the same name from the main language file. It's not used for anything technical, just the display name in the table of contents.

Qikahome:
请修改"modifier.tconstruct.glowing.description"，它自2021年5月29的提交就不再需要潜行
Please modify "modifier.tconstruct.glowing.description", as it has not required stealth since the update on May 29, 2021.

PR #5505, opened 2025-10-02
Add Simplified Chinese translation for the "Extract Enchantment" in "Puny Smelting"

Niangao-YX:


KnightMiner:
This is already covered by #5459, which is far more complete. Please communicate with the author of that PR if you think their translation can be improved.

Niangao-YX:
ok thx, i think he's translation is good

PR #5519, opened 2025-11-22
Add two Simple page buttons for Tinker Station

Qikahome:
image
I added two page-turning buttons to Tinker Station. 
These buttons will not appear unless there are more than 48 Station Layouts.
These changes seem not to break anything.
The style and placement of the button still need to be carefully considered.


KnightMiner:
Ah nice, that will help out a common problem people with too many addons have. Certainly better than my quick fix that would be in the next update. I'll test this out when I get a chance but code look good beyond a few stylistic changes.

I do wonder if its worth the option to keep the armor stand visible with the page buttons and just having fewer rows, could be a config option. But not essential as more rows is probably better.

src/main/java/slimeknights/tconstruct/tables/client/inventory/widget/SideButtonsWidgetPaged.java
```diff
@@ -0,0 +1,157 @@
...
+
+
+import net.minecraft.network.chat.Component;
+
+import net.minecraft.client.gui.components.Button;
+import slimeknights.mantle.client.screen.MultiModuleScreen;
+import lombok.Getter;
+import net.minecraft.client.gui.GuiGraphics;
+
+public class SideButtonsWidgetPaged<T extends Button> extends SideButtonsWidget<T> {
```
KnightMiner, line 19: Can you add a quick javadoc here? I generally like anything new that is public/protected (particularly classes and methods) to have a javadoc comment unless its a method override. Anything private doesn't need to have javadocs.

Same for the public static constants, just helps the future maintainer remember what the constants are used for.

src/main/java/slimeknights/tconstruct/tables/client/inventory/widget/SideButtonsWidgetPaged.java
```diff
@@ -0,0 +1,157 @@
...
+
+    public static final int SPACING = 4;
+
+    public static final int MAX_ROWS = 8;
+
+    public final MultiModuleScreen<?> parent;
```
KnightMiner: Any reason this and `page` below needs to be public? at a glance I don't see its usage.

Reason for lower visibility is once a version becomes stable, I try to avoid any breaking changes; things that are public that could be private can make it harder to change later.

Qikahome: I'm just used to making everything public.
it's a bad habit

PR #5527, opened 2025-11-28
A handfull of suggested entity type tags for entity melting recipes

Sarinsa:
Based on previous discussion, here are the tags I suggest to add. Personally, I will be utilizing most of them.
Recipes that were using vanilla or forge tags have not been changed.


src/main/java/slimeknights/tconstruct/common/TinkerTags.java
```diff
@@ -599,6 +599,26 @@ private static void init() {}
     public static final TagKey<EntityType<?>> SLIMES = common("slimes");
     public static final TagKey<EntityType<?>> BACON_PRODUCER = local("bacon_producer");
 
+    // Here goes a big block of tags used in entity melting recipes.
+    public static final TagKey<EntityType<?>> MELTABLE_BEES = local("melting/meltable_bees");
```
KnightMiner: These names are a little redundant. I would compress `melting` and `meltable` into one thing. Maybe use `meltable/bees`, or `melting/bees`, or something similar.

src/main/java/slimeknights/tconstruct/common/TinkerTags.java
```diff
@@ -599,6 +599,26 @@ private static void init() {}
     public static final TagKey<EntityType<?>> SLIMES = common("slimes");
     public static final TagKey<EntityType<?>> BACON_PRODUCER = local("bacon_producer");
 
+    // Here goes a big block of tags used in entity melting recipes.
```
KnightMiner: I would be more specific on the why instead of the what. We use these tags in melting recipes to make it easier for mods that add vanilla-like mobs to add support.

src/main/java/slimeknights/tconstruct/common/TinkerTags.java
Sarinsa: I see what you mean. I'll fix that

src/main/java/slimeknights/tconstruct/common/TinkerTags.java
Sarinsa: Oh wow. yeah that comment isn't very helpful, I'll revisit that

Sarinsa:
I realize using forge and/or minecraft tags for some recipes and melting tags for others is a bit janky; entirely different scopes. For example, I think it better to use a "melting/skeletons" tag over "minecraft:skeletons", but this would obviously break existing "compat". What is your stance on this?

KnightMiner:
I don't see much point in switching tags. These are highly specific usecase tags, you want to tag your mobs to work in our recipes.

Skeletons you can already solve that with the general tag. Other mobs just lack such general tags.

Sarinsa:
The reason I thought of this is that some mobs tagged as "forge:whatever" may not thematically fit the recipe result or have drops that reflect the result. 

After giving it a bit more thought, I realize if I am so invested in that stuff I am better off just taking the time to make the recipes I want and include it as default data on my end. 

I will revert the removal of forge tag uses in the recipes and push that.

KnightMiner:
If such a case happens, it can always be fixed when we encounter it.

KnightMiner:
Looks like some of my other tag changes conflicted with this PR, so I'll have to put it off until the inevitable 3.11.1 update. Once you get that resolved I can go ahead and merge this in.

Sarinsa:
This PR is now up to date! I added back the "melting/meltable_skeletons" tag as well for consistency. By default it contains "#minecraft:skeletons" and "minecraft:skeleton_horse". Let me know if you want me to change anything else.

KnightMiner:
Something is not right with rebased pull request; it now shows as every commit for the last 7 months is part of this PR. Make sure you rebased with respect to the latest 1.20 branch, not simply merged.

I could also just directly commit the tags if that is easier.

Sarinsa:
The latter option is likely quicker, seen as my Git knowledge is poor

PR #5568, opened 2026-01-02
Add BlockItemProvider capability to allow alternative sources of blocks for modifiers like Exchanging

ChiefArug:
This adds a capability `BlockItemProviderCapability` that can be added to ItemStacks to let them provide block items. This is intended for compatibility with other mods like Botania that add items that place blocks at the cost of things such as Mana, or hold large numbers of a single block.

BlockItem rather than Block was chosen as the current exchanging logic works off BlockItem, however looking into other places that use `instanceof BlockItem` there are alternatives for placing the block 'manually' which may be added to Exchanging and the capability swapped to directly use Blocks.

Additionally a new modifier hook has been added to provide a blockitem (at the cost of durability), which glowing uses to provide glows.

TODO:
- [x] Investigate updating Exchanging to be able to place non BlockItem blocks.
- [x] Look into support for PlaceBlockFluidEffect
- [x] Do a final docs pass


src/main/java/slimeknights/tconstruct/library/tools/capability/inventory/BlockItemProviderCapability.java
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  /** Registers the capability with the event bus */
+  private static void register(RegisterCapabilitiesEvent event) {
+    event.register(BlockItemProviderCapability.class);
+  }
+
+  /** Event listener to attach the capability */
+  private static void attachCapability(AttachCapabilitiesEvent<ItemStack> event) {
+    if (event.getObject().getItem() instanceof BlockItem block) {
```
KnightMiner: Attaching this to every block item feels a bit wasteful, especially with how 1.20 capabilities are implemented.

Why not just have a fallback if the stack lacks the capabilty?

ChiefArug: Its a lot cleaner, as the alternative of special casing for everything that uses this capability (probably only two things) is quite ugly imo (and can lead to issues such as not being able to override the block a BlockItem provides if a consumer of the capability mistakenly orders the checks the wrong way around).
It also acts as an example to other implementers, though thats a very weak reason.

KnightMiner:
> BlockItem rather than Block was chosen as the current exchanging logic works off BlockItem, however looking into other places that use instanceof BlockItem there are alternatives for placing the block 'manually' which may be added to Exchanging and the capability swapped to directly use Blocks.

This could always be done by putting the capability in charge of placing the block to some degree. Default impl runs block item placement, but you could do an impl for placement from other sources.

----

Related, if we are going to place in exchanging here, we should place from ichors fluid effect in the same way. We do not, however, need to use this capability for the fixed block forms of fluid effects, just the one that selects block from the offhand.

src/main/java/slimeknights/tconstruct/library/tools/capability/inventory/BlockItemProviderCapability.java
KnightMiner: In your experience, how notable is the cost of such attachments? I know firing the event is expensive, but it might be once you need to fetch caps anyways that cost is unavoiable.

ChiefArug: I haven't done any specific profiling, but I rarely see cap stuff come up in spark profiles and when it does its usually Curios capability providers which do expensive stuff for some reason (can't remember the specifics).

In this particular case its just allocating a single object, which should be fine especially given the overhead of the event anyway. Some caching could be implemented but I doubt that is worth it.

ChiefArug:
> This could always be done by putting the capability in charge of placing the block to some degree. Default impl runs block item placement, but you could do an impl for placement from other sources.

After implementing this I have realised that theres nowhere in Tinkers (or in the compat I had planned for Botania) where this would actually be used, as they all consume from the inventory anyway. This means I can't test it without some annoying additional useless code. I've [committed and pushed this code to my fork on a different branch](https://github.com/ChiefArug/TinkersConstruct/commit/fa4c41624cf1e9fc01974110f21717493ccdf1e7) in case its useful to anyone, but will leave it out of this PR.

ChiefArug:
Should be ready to go now.

If the default BlockItem cap does end up being a performance issue then additional context could be added to the `getBlockItem` and `consume` methods to allow a single instance to be reused between all `BlockItem`s.

ChiefArug:
Actually turns out that the Botania integration I am thinking of requires some extra context like that, so time to rewrite it a bit.

ChiefArug:
One thing to consider here is if tools such as those with glowing should implement this capability to provide glows.

src/main/java/slimeknights/tconstruct/library/modifiers/ModifierHooks.java
```diff
@@ -124,6 +126,16 @@ public float getRepairAmount(IToolStackView tool, ModifierEntry modifier, Materi
   /** Hook running while the tool is in the inventory */
   public static final ModuleHook<InventoryTickModifierHook> INVENTORY_TICK = register("inventory_tick", InventoryTickModifierHook.class, InventoryTickModifierHook.AllMerger::new, (tool, modifier, world, holder, itemSlot, isSelected, isCorrectSlot, stack) -> {});
 
+  /** Hook for providing a BlockItem via the {@link slimeknights.tconstruct.library.tools.capability.BlockItemProviderCapability}*/
+  public static final ModuleHook<ToolBlockItemProviderHook> BLOCK_ITEM_PROVIDER = register("block_item_provider", ToolBlockItemProviderHook.class, new ToolBlockItemProviderHook() {
+    @Override
+    public @Nullable BlockItem getBlockItem(IToolStackView tool, ModifierEntry modifier, @Nullable LivingEntity entity, @Nullable ItemStack stack) {
```
KnightMiner: This is minor, but I have always preferred `@Nullable` above the method rather than next to the return type. It just looks weird there.

src/main/java/slimeknights/tconstruct/library/modifiers/fluid/block/PlaceBlockFluidEffect.java
```diff
@@ -55,104 +57,150 @@ public float apply(FluidStack fluid, EffectLevel level, FluidEffectContext.Block
...
-          for (InteractionHand hand : InteractionHand.values()) {
-            ItemStack held = entity.getItemInHand(hand);
-            if (!held.isEmpty() && held.getItem() instanceof BlockItem blockItem) {
-              block = blockItem.getBlock();
-              stack = held;
-              useHand = hand;
-              break;
-            }
+        ItemStack stack = new ItemStack(block);
+        if (!context.placeRestricted(stack)) {
```
KnightMiner: For the case of `block != null`, lets simplify this call to just `player == null || player.mayBuild()` since the `placeRestricted` helper is really just after the item stack NBT (always null in this branch since you split them)

I suppose we could add a version of `placeRestriced` to context that does not take a stack, but code seems pretty simple here.

Also, I feel like we are missing a `return 0` in the "else" for this method. If we have a block set and its unable to be placed we should give up, not try to search the entity for one.

src/main/java/slimeknights/tconstruct/library/modifiers/fluid/block/PlaceBlockFluidEffect.java
```diff
@@ -55,104 +57,150 @@ public float apply(FluidStack fluid, EffectLevel level, FluidEffectContext.Block
...
-          placeContext = blockItem.updatePlacementContext(placeContext);
-          if (placeContext == null) {
-            return 0;
+
+      if (entity != null) {
+        // either hand is fine, allows using the tool from offhand or mainhand
+        // iterate in reverse order so that we prefer the offhand
+        for (InteractionHand hand : new InteractionHand[]{InteractionHand.OFF_HAND, InteractionHand.MAIN_HAND}) {
+          ItemStack held = entity.getItemInHand(hand);
+          var cap = getBlockProvider(held);
```
KnightMiner: I am not a fan of `var`. Just call it what it is so the code is more readable on github.

src/main/java/slimeknights/tconstruct/library/modifiers/fluid/block/PlaceBlockFluidEffect.java
```diff
@@ -55,104 +57,150 @@ public float apply(FluidStack fluid, EffectLevel level, FluidEffectContext.Block
...
+          if (cap == null) continue;
+          BlockItem blockItem = cap.getBlockItem(held, entity);
+          if (blockItem == null) continue;
+
+          ItemStack stack = new ItemStack(blockItem);
+          if (context.placeRestricted(stack)) continue;
```
KnightMiner: This should really be called on the original stack `held`, not on the result from the block item provider. Reason being its just doing NBT checks and things not changing with respect to the entity.

Could probably do this first before you even query the capability.

src/main/java/slimeknights/tconstruct/library/modifiers/fluid/block/PlaceBlockFluidEffect.java
```diff
@@ -55,104 +57,150 @@ public float apply(FluidStack fluid, EffectLevel level, FluidEffectContext.Block
...
-        if (!state.canSurvive(world, clicked) || !world.isUnobstructed(state, clicked, player == null ? CollisionContext.empty() : CollisionContext.of(player))) {
-          return 0;
+      } else {
+        ItemStack held = context.getStack();
+        var cap = getBlockProvider(held);
+        if (cap == null) return 0;
+        BlockItem blockItem = cap.getBlockItem(held, entity);
+        if (blockItem == null) return 0;
+        ItemStack stack = new ItemStack(blockItem);
+        if (context.placeRestricted(stack)) return 0;
```
KnightMiner: same as above, this should use `held`, no NBT on the stack you just made up.

src/main/java/slimeknights/tconstruct/library/modifiers/fluid/block/PlaceBlockFluidEffect.java
```diff
@@ -55,104 +57,150 @@ public float apply(FluidStack fluid, EffectLevel level, FluidEffectContext.Block
...
-        // simulate note: we don't ask the block item for its state for placement as that method is protected, this notably affects signs/banners (unlikely need)
-        BlockState state = block.getStateForPlacement(placeContext);
-        if (state == null) {
-          return 0;
-        }
-        // simulate note: we don't call BlockItem#canPlace as its protected, though never overridden in vanilla
-        if (!state.canSurvive(world, clicked) || !world.isUnobstructed(state, clicked, player == null ? CollisionContext.empty() : CollisionContext.of(player))) {
-          return 0;
+      } else {
+        ItemStack held = context.getStack();
```
KnightMiner: All the code after this line is the same as the entity search; is there a good way to extract it into a common method or to a common section below the if statement?

I suppose we would need to have a 3 state return for the place method, so we can tell the difference between skip this stack and tried to place and fail (for the hand iteration).

If it ends up messier to separate it then probably best to just leave it as is.

src/main/java/slimeknights/tconstruct/library/tools/capability/BlockItemProviderCapability.java
```diff
@@ -0,0 +1,105 @@
...
+import net.minecraftforge.common.util.LazyOptional;
+import net.minecraftforge.event.AttachCapabilitiesEvent;
+import net.minecraftforge.eventbus.api.EventPriority;
+import net.minecraftforge.fml.javafmlmod.FMLJavaModLoadingContext;
+import slimeknights.tconstruct.TConstruct;
+
+import javax.annotation.Nullable;
+
+/**
+ * A capability that provides block items to things that place blocks, such as
```
KnightMiner: I tend to not break javadocs into multiple lines mid sentence unless its quite a bit longer than these sentences.

src/main/java/slimeknights/tconstruct/library/tools/capability/BlockItemProviderCapability.java
```diff
@@ -0,0 +1,105 @@
...
+ */
+public interface BlockItemProviderCapability {
+
+  /** Capability ID */
+  ResourceLocation ID = TConstruct.getResource("block_provider");
+  /** Capability type */
+  Capability<BlockItemProviderCapability> CAPABILITY = CapabilityManager.get(new CapabilityToken<>() {});
+
+  /** Registers this capability */
+  static void register() {
```
KnightMiner, line 39: Mark this as `@Internal`; just to minimize chances of some addon calling it.

src/main/java/slimeknights/tconstruct/library/tools/capability/BlockItemProviderCapability.java
```diff
@@ -0,0 +1,105 @@
...
+
+  /** Event listener to attach default implementation(s) of the capability */
+  private static void attachCapability(AttachCapabilitiesEvent<ItemStack> event) {
+    if (event.getObject().getItem() instanceof BlockItem) {
+      event.addCapability(SimpleBlockItem.ID, SimpleBlockItem.INSTANCE);
+    }
+  }
+
+  /**
+   * @return The block provider for this stack, or null if this stack cannot provide blocks
```
KnightMiner: I believe if you use `{@return ...}` it will automatically use this for both description and return tag.

src/main/java/slimeknights/tconstruct/library/tools/capability/BlockItemProviderCapability.java
```diff
@@ -0,0 +1,105 @@
...
+   */
+  @SuppressWarnings("DataFlowIssue")
+  @Nullable
+  static BlockItemProviderCapability getBlockProvider(ItemStack stack) {
+    return stack.getCapability(CAPABILITY).orElse(null);
```
KnightMiner: `LogicHelper.orElseNull` from Mantle saves you the annotation here if you want.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * Consume a block from this provider. For example may decrease a contained stacks size or remove fluid from the stack's tank.
+     */
+    void consume(IToolStackView tool, ModifierEntry modifier, @Nullable LivingEntity entity, @Nullable ItemStack stack);
+
+    @Nullable
+    static BlockItemProviderCapability getHookAsCapability(IToolStackView tool) {
+        for (ModifierEntry entry : tool.getModifiers()) {
+            ToolBlockItemProviderHook provider = entry.getModifier().getHooks().getOrNull(ModifierHooks.BLOCK_ITEM_PROVIDER);
+            if (provider != null) return new CapabilityImpl(provider, tool, entry);
```
KnightMiner: This implementation may cause problems when multiple modifiers both can provide blocks. If the first provider returns null we should query the second provider instead of giving up entirely. As an example, consider a glowing tool with a "exchanging bag" modifier; if the bag has priority but is empty I would want glowing to be used. I don't think merging on a single modifier is needed (that is what the merger in the hook is for), but merging between multiple modifiers is essential.

I think its probably best if the CapabilityImpl iterates the modifiers in both methods. For `consume`, either use a nonnull return on `getBlockItem()` to determine when to stop iteration, or have `consume` return a boolean so you know it consumed.

src/main/java/slimeknights/tconstruct/shared/TinkerCommons.java
```diff
@@ -101,7 +101,7 @@ public final class TinkerCommons extends TinkerModule {
   /*
    * Blocks
    */
-  public static final RegistryObject<GlowBlock> glow = BLOCKS.registerNoItem("glow", () -> new GlowBlock(builder(MapColor.NONE, SoundType.WOOL).noCollission().pushReaction(PushReaction.DESTROY).replaceable().strength(0.0F).lightLevel(s -> 14).noOcclusion()));
+  public static final ItemObject<GlowBlock> glow = BLOCKS.register("glow", () -> new GlowBlock(builder(MapColor.NONE, SoundType.WOOL).noCollission().pushReaction(PushReaction.DESTROY).replaceable().strength(0.0F).lightLevel(s -> 14).noOcclusion()), BLOCK_ITEM);
```
KnightMiner: Seems there are addons using this field, and changing the field type is a binary break.

We might just want to use a new field name for the `ItemObject` version (`glowBlock` perhaps?), and then call `RegistryObject.create` on the existing name `glow` (marked deprecated for removal).

src/main/java/slimeknights/tconstruct/tools/data/ModifierProvider.java
```diff
@@ -317,7 +320,8 @@ protected void addModifiers() {
       .addModule(ShowOffhandModule.DISALLOW_BROKEN)
       .addModule(new PlaceGlowModule(5))
       .addModule(new GlowWalkerModule(new LevelingValue(2, 1), 3, 5))
-      .addModule(new ProjectilePlaceGlowModule(5, true, true));
+      .addModule(new ProjectilePlaceGlowModule(5, true, true))
+      .addModule(new BlockItemProviderModule((BlockItem) TinkerCommons.glow.asItem(), 5,ModifierCondition.ANY_TOOL.with(ToolStackPredicate.tag(TinkerTags.Items.HELD))));
```
KnightMiner: missing space between `5,` and `ModifierCondition`

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/tool/ExchangingModifier.java
```diff
@@ -39,14 +42,32 @@ public int getPriority() {
...
     }
+    BlockItemProviderCapability blockProvider = BlockItemProviderCapability.getBlockProvider(item);
+    BlockItem blockItem = blockProvider == null ? null : blockProvider.getBlockItem(item, entity);
+    // if the thing in our offhand cannot provide at all or cannot currently provide then check
+    // the mainhand next (this tool), in case we have glowing or a similar modifier to provide blocks.
+    if (blockItem == null) {
+      item = context.getLiving().getMainHandItem();
+      // skip forges cap system and go to the tinkers hook because we know this is a tinkers tool
+      blockProvider = ToolBlockItemProviderHook.getHookAsCapability(tool);
+      if (blockProvider == null)
```
KnightMiner: my preference when it comes to no `{}` if statements is only if they are on the same line; leaving them out with just indentation can lead to misleading code especially when tabs vs spaces fails us. So either of these is fine:
```java
if (blockProvider == null) return null;

if (blockProvider == null) {
  return null;
}
```

src/main/java/slimeknights/tconstruct/library/modifiers/modules/behavior/BlockItemProviderModule.java
```diff
@@ -0,0 +1,48 @@
...
+import slimeknights.tconstruct.library.modifiers.modules.ModifierModule;
+import slimeknights.tconstruct.library.modifiers.modules.util.ModifierCondition;
+import slimeknights.tconstruct.library.module.ModuleHook;
+import slimeknights.tconstruct.library.tools.capability.ToolBlockItemProviderHook;
+import slimeknights.tconstruct.library.tools.helper.ToolDamageUtil;
+import slimeknights.tconstruct.library.tools.nbt.IToolStackView;
+
+import java.util.List;
+
+public record BlockItemProviderModule(BlockItem item, int damage, ModifierCondition<IToolStackView> condition) implements ModifierModule, ToolBlockItemProviderHook, ModifierCondition.ConditionalModule<IToolStackView> {
```
KnightMiner: Make sure to add javadocs to new classes. I think this one also belongs among the modules in `library` as I can expect addons to want to use it.

KnightMiner (changes requested):
Apart from cosmetics, just have three concerns:

* If two different modifiers implement the new hook, the merging will simply ignore the second. It should try each hook until it finds the first with a nonnull block item.
* Place block fluid effect has some issues with how it handles `placeRestricted`.
* Seems you need a rebase, as when you implemented the glowing changes it conflicted with an older feature related to the tooltips.

Otherwise looks pretty good as a way to bring back that behavior with exchanging.

src/main/java/slimeknights/tconstruct/tools/modifiers/BlockItemProviderModule.java
```diff
@@ -0,0 +1,48 @@
...
+    }
+
+    @Override
+    public @Nullable BlockItem getBlockItem(IToolStackView tool, ModifierEntry modifier, @Nullable LivingEntity entity, @Nullable ItemStack stack) {
+        return !tool.isBroken() && condition.matches(tool, modifier) ? item : null;
+    }
+
+    @Override
+    public void consume(IToolStackView tool, ModifierEntry modifier, @Nullable LivingEntity entity, @Nullable ItemStack stack) {
+        ToolDamageUtil.damage(tool, damage, entity, stack);
```
KnightMiner: Thinking about it, we probably want to show the break animation here via the entity method. Rather than passing slot to call it, we can use the passed stack to search the 6 equipment slots (if it's none of the 6, can't show the animation anyways). Damage returns a boolen when it breaks, so if that's true and the entity/stack are non null can search

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
KnightMiner: As a followup idea, perhaps we want to pass the block item as a parameter to consume so the consume hook knows what we are trying to consume. Helps on the chance we have multiple blocks to choose from to avoid consuming the wrong thing; could he neat for a provider that randomizes choices sorta like the quark trowel.

For modifiers, block item can help you find which modifier to call consume on.

----

Still inclined to say we handle this for modifiers via consume with a boolean return for "did consume", as that lets us optimize things like inventory searches.

src/main/java/slimeknights/tconstruct/library/modifiers/ModifierHooks.java
ChiefArug: I'm the same, this was just automatic by ij and i didn't notice it.
Though unfortunately now with people moving to jspecify and that only going on types will probably end up having to be on the return type eventually, but that will be quite a few years away.

src/main/java/slimeknights/tconstruct/library/modifiers/fluid/block/PlaceBlockFluidEffect.java
ChiefArug: This makes me consider that in the case of the cap being backed by an actual item stack of a block (ie the blockitem impl, or a theoretical block bag) it should use the tags on that, so might need to add a `@Nullable ItemStack getBackingStack()` or similar to the cap. Thoughts?

src/main/java/slimeknights/tconstruct/library/tools/capability/BlockItemProviderCapability.java
ChiefArug: Id personally avoid that, what's the point in stating the exact same thing twice
image

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
ChiefArug: I have been avoiding a boolean for consume as we don't know if the block placement was successful till after we have actually placed it. We can make a good guess, but we either have to consume an item without placing a block in rare case, or place blocks for free in rare cases. I've opted for the second as we have more control over the capability and can be a little bit more sure that when we say we can consume (return non-null from getBlockItem), we can actually consume, rather than placing a block where any old mod could be adding weird checks to setBlock.

src/main/java/slimeknights/tconstruct/library/modifiers/fluid/block/PlaceBlockFluidEffect.java
KnightMiner: Yeah, I keep thinking we might want something like that as well. Also allows you to return NBT with your stack, e.g. to place retexturable blocks from tinkers. I would also recommend making it nonnull and using empty to indicate "I could have a block item but currently have nothing". Be clear in the docs the stack should contain a block item (though still probably should validate it in usages, perhaps in a helper method).

For the module implementation, we can stick to block item for simplicity.

For the consume hook, pass both the original stack and the stack returned by the cap. This allows them to ensure they shrink the proper thing for bags.

src/main/java/slimeknights/tconstruct/library/tools/capability/BlockItemProviderCapability.java
KnightMiner: Mostly just looks wrong when a javadoc has just a returns and no description.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
KnightMiner: The boolean is the return for consume in the modifier hook version (no need on the capability version). When you call consume on the capability for a tool, you iterate each nested modifier and ask "are you in charge of consuming?" If it returns false, you keep iterating. If it returns true, you found the right modifier.

This is not a boolean parameter on getting the block. This is a boolean return on consume, just for the modifier hook.

```java
boolean consume(ItemStack stack, ...);

void consume(ItemStack stack, ...) {
  for (ModifierEntry entry : tool.getModifiers()) {
    if (entry.getHook(BLOCK_PROVIDER).consume(stack, ...)) {
      return;
    }
  }
  // probably log some warning about being unable to consume the block, should never reach here
}
```

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
```diff
@@ -0,0 +1,77 @@
...
+    /**
+     * @return the {@link BlockItem} that this provides, or {@code null} if this cannot provide more block items (for example if the stack has been depleted)
+     */
+    @Nullable
+    BlockItem getBlockItem(IToolStackView tool, ModifierEntry modifier, @Nullable LivingEntity entity, @Nullable ItemStack stack);
+
+    /**
+     * Consume a block from this provider. For example may decrease a contained stacks size or remove fluid from the stack's tank.
+     */
+    void consume(IToolStackView tool, ModifierEntry modifier, @Nullable LivingEntity entity, @Nullable ItemStack stack);
```
KnightMiner: Something that just occured to me that I forgot to mention: modifier hook methods names should clearly reference back to the hook name as a module may implement many hooks (and lots of similarly named methods get confusing where each comes from, not to mention the chance of conflict).

In this case, something like `consumeBlockItem` should be clear enough. No need to change the capability.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
ChiefArug: I remember the modifier that provided the hook (as I'm already allocating an instance of the cap to cache the tool stack view to prevent reparsing from NBT so might as well store the modifier and hook instance as well). I think this should handle this.

src/main/java/slimeknights/tconstruct/library/tools/capability/BlockItemProviderCapability.java
```diff
@@ -70,11 +72,21 @@ static BlockItemProviderCapability getBlockProvider(ItemStack stack) {
   BlockItem getBlockItem(ItemStack stack, @Nullable LivingEntity entity);
 
   /**
+   * Get the stack backing the provided BlockItem. The returned stack will be not be modified as it is copied immediately.
+   * The returned stack is primarily used to determine placement state and placement permissions (for adventure mode players).
+   * @return {@link ItemStack#EMPTY} if there is no backing item, otherwise an {@link ItemStack} instance holding at least one of {@code item}.
+   */
+  ItemStack getBackingStack(ItemStack capStack, BlockItem item, @Nullable LivingEntity entity);
```
KnightMiner: Seems it might be worth this being a default method to return empty, since that is the easiest way to do a no NBT block item. Though less sure on whether this is useful on the capability implement.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
KnightMiner: That is okay valid if you precheck that the hook has a block item return. There is a small performance concern there as it may require NBT parsing to make that decision, causing that parsing to happen up to 3 times, though this isn't a particularly hot spot.

Remember, test case is we have two modifiers on the same stack that are both providers. Conditions one modifier means it currently returns nothing. Thus, next modifier should be given a chance.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
```diff
@@ -17,12 +17,26 @@ public interface ToolBlockItemProviderHook {
...
+    void consumeBlockItem(IToolStackView tool, ItemStack toolStack, ModifierEntry modifier, ItemStack backingStack, @Nullable LivingEntity entity);
+
+    /**
+     * Get the stack backing the provided BlockItem. The returned stack will be not be modified as it is copied immediately.
+     * The returned stack is primarily used to determine placement state and placement permissions (for adventure mode players).
+     * @param tool The tool that this hook is attached to, as a tool stack view
+     * @param toolStack The tool that this hook is attached to
+     * @return {@link ItemStack#EMPTY} if there is no backing item, otherwise an {@link ItemStack} instance holding at least one of {@code item}.
+     */
+    ItemStack getBackingStack(IToolStackView tool, ItemStack toolStack, ModifierEntry modifier, BlockItem item, @Nullable LivingEntity entity);
```
KnightMiner: Having this be a separate method from the block item one does raise a potential concern when getting the block item requires parsing NBT, as weat have to do all that parsing twice. This mostly becomes a risk if using something in the fluid cannon. Might just mean it's best to hold off on a exchanging bag modifier until 1.21.

It may be worth having this default to returning empty as both our current implementation want that.

src/main/java/slimeknights/tconstruct/shared/TinkerCommons.java
```diff
@@ -101,7 +103,12 @@ public final class TinkerCommons extends TinkerModule {
   /*
    * Blocks
    */
-  public static final ItemObject<GlowBlock> glow = BLOCKS.register("glow", () -> new GlowBlock(builder(MapColor.NONE, SoundType.WOOL).noCollission().pushReaction(PushReaction.DESTROY).replaceable().strength(0.0F).lightLevel(s -> 14).noOcclusion()), BLOCK_ITEM);
+  public static final ItemObject<GlowBlock> glowBlock = BLOCKS.register("glow", () -> new GlowBlock(builder(MapColor.NONE, SoundType.WOOL).noCollission().pushReaction(PushReaction.DESTROY).replaceable().strength(0.0F).lightLevel(s -> 14).noOcclusion()), BLOCK_ITEM);
+  /**
+   * @deprecated Use {@link #glowBlock}
+   */
+  @Deprecated(forRemoval = true)
+  public static final RegistryObject<GlowBlock> glow = RegistryObject.create(TConstruct.getResource("glow"), ForgeRegistries.BLOCKS);
```
KnightMiner: Should be able to just use glowBlock.getId() here, or whatever I called it.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
ChiefArug: Yeah I suspect 1.21 can rework this a little bit with not having to pass around IToolStackView everywhere either due to data components and a few related things.
Have made it default to `EMPTY` though.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
```diff
@@ -17,51 +19,68 @@ public interface ToolBlockItemProviderHook {
...
-    record CapabilityImpl(ToolBlockItemProviderHook base, IToolStackView tool, ModifierEntry modifier) implements BlockItemProviderCapability {
+    record CapabilityImpl(IToolStackView tool) implements BlockItemProviderCapability {
 
+        @Nullable
         @Override
-        public @Nullable BlockItem getBlockItem(ItemStack capStack, @Nullable LivingEntity entity) {
-            return base.getBlockItem(tool, capStack, modifier, entity);
+        public BlockItem getBlockItem(ItemStack capStack, @Nullable LivingEntity entity) {
+            for (ModifierEntry entry : tool.getModifiers()) {
+                ToolBlockItemProviderHook provider = entry.getModifier().getHooks().getOrNull(ModifierHooks.BLOCK_ITEM_PROVIDER);
```
KnightMiner: Just do `entry.getHook(ModifierHooks.BLOCK_ITEM_PROVIDER).getBlockItem(tool, entry, entity)`. Its why the default instance exists.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
```diff
@@ -17,51 +19,68 @@ public interface ToolBlockItemProviderHook {
...
+            TConstruct.LOG.warn("Could not find a modifier to consume {} from after providing it from ToolBlockItemProviderHook. This is likely causing a duplication glitch! Stack nbt: {}", BuiltInRegistries.ITEM.getKey(item), backingStack.getTag());
         }
 
         @Override
         public ItemStack getBackingStack(ItemStack capStack, BlockItem item, @Nullable LivingEntity entity) {
-            return base.getBackingStack(tool, capStack, modifier, item, entity);
+            for (ModifierEntry entry : tool.getModifiers()) {
+                ToolBlockItemProviderHook provider = entry.getModifier().getHooks().getOrNull(ModifierHooks.BLOCK_ITEM_PROVIDER);
+                if (provider != null) {
+                    ItemStack backingStack = provider.getBackingStack(tool, entry, item, entity);
```
KnightMiner: There is some concern here that the modifier returning the block item and the one returning the item stack are different modifiers.

With consuming it doesn't matter a ton as as long as someone consumes that item, its properly consumed. With block items though, we might be providing a block from one modifier and then another decides to go and add a stack to it unexpectedly.

Is there any reason to not just have `getBlockItem` return an item stack which we validate with a quick `instanceof` in the caller? Or alternatively, return a record with the block item + item stack pair?

I think I like the record overall, we can even cache the record instance in a module as its the same each time.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
```diff
@@ -80,11 +99,7 @@ public <T> LazyOptional<T> getCapability(IToolStackView tool, Capability<T> cap)
             if (cap == BlockItemProviderCapability.CAPABILITY) {
                 LazyOptional<BlockItemProviderCapability> lo = lazy;
                 if (lo == null) {
-                    // Iterate modifiers to see if any of them actually provide this hook so we don't provide the cap unless we actually have the hook
-                    BlockItemProviderCapability capability = getHookAsCapability(tool);
-                    if (capability == null)
-                        return LazyOptional.empty();
-                    lo = lazy = LazyOptional.of(() -> capability);
+                    lo = lazy = LazyOptional.of(() -> new CapabilityImpl(tool));
```
KnightMiner: We could initialize this in the constructor using the context passed by the place you registered this. Seems my other impls store the supplier but no reason you can't do `LazyOptional.of(() -> new CapabilityImpl(tool.get())` in the constructor (the `LazyOptional` ensures that we don't immediately resolve the tool supplier, and its the same instance we get in this method)

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
ChiefArug: Hmm another edge case I hadn't considered.
I was hoping to not have to rely on runtime checks as those lead to either eagerly crashing (annoying for users) or bugs not getting reported (users don't read the log), but it is the most sensible way to get around this issue and in this case there will be non-working behaviour in game anyway. I'll just have it log a warning and treat it as ItemStack.EMPTY.

Note sure how the caching you mentioned would work, would it need another module to hold ToolBlockItemProviderHook.Provider? Not sure how important that is given it is just a holder and making a new one is just allocating a few objects and fields.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
ChiefArug: Ah, I hadn't realized that for a given stack with a given provieder the tool stack instance won't change, it just refreshes its fields, so I was working off the assumption that I needed to handle clearing the cache in case the tool nbt had changed.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
KnightMiner: If you want to avoid runtime checks, do a record `BlockStack(BlockItem block, ItemStack stack)`. Method then returns `@Nullable BlockStack` instead of `ItemStack`.

In `BlockItemProviderModule`, add a field `BlockStack` which in the main constructor is set equal to `new BlockStack(blockItem, ItemStack.EMPTY)`. This cache is not at a capability level, its just handled in the one module that has the ability to cache it (same item returned every time).

----

If you want to avoid extra allocation for the cases that do not always return the same block, then we just return `ItemStack`. Can cache a `ItemStack` in one of the fields on `BlockItemProviderModule`, just make sure we copy it (have to decide if we copy on return or we copy before we mutate it).

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
KnightMiner: Yeah, right now we just clear its cache whenever the capability is fetched. Is a little bit wasteful if a stack persists for awhile, but seems unlikely to me.

src/main/java/slimeknights/tconstruct/library/tools/capability/ToolBlockItemProviderHook.java
```diff
@@ -0,0 +1,81 @@
...
+                }
+            }
+            return ItemStack.EMPTY;
+        }
+
+        @Override
+        public void consume(ItemStack capStack, ItemStack backingStack, @Nullable LivingEntity entity) {
+            for (ModifierEntry entry : tool.getModifiers()) {
+                ToolBlockItemProviderHook provider = entry.getModifier().getHooks().getOrNull(ModifierHooks.BLOCK_ITEM_PROVIDER);
+                if (provider != null && provider.consumeBlockItem(tool, capStack, entry, backingStack, entity)) {
```
KnightMiner, line 62: `if (entry.getHook(ModifierHooks.BLOCK_ITEM_PROVIDER).consumeBlockItem(...))`

src/main/java/slimeknights/tconstruct/tools/modifiers/ability/tool/ExchangingModifier.java
```diff
@@ -39,12 +44,42 @@ public int getPriority() {
...
-    // must have blocks in the offhand
-    ItemStack offhand = context.getLiving().getOffhandItem();
+    // We check the offhand first
+    ItemStack item = context.getLiving().getOffhandItem();
     BlockState state = context.getState();
     Level world = context.getWorld();
     BlockPos pos = context.getPos();
-    if (offhand.isEmpty() || !(offhand.getItem() instanceof BlockItem blockItem)) {
+    LivingEntity entity = context.getLiving();
+    if (item.isEmpty()) return null;
```
KnightMiner, line 53: This causes an issue where empty offhand blocks us from using mainhand's glowing.

KnightMiner (changes requested):
Just had a chance to test this in game and it overall looks good. Just one small bug I discovered when testing it. Should be good to merge after fixing that.

Also seems I accidently gave you another merge conflict due to the reusable projectile logic.

PR #5573, opened 2026-01-06
Add Dynamic ToolStatType to datapack

Qikahome:
It seems work well;
TEST THINGPACK: 
[testpack.zip](https://github.com/user-attachments/files/24451661/testpack.zip)


Qikahome:
I know nothing about the test, so it cant build now, but it can run client

src/main/java/slimeknights/tconstruct/library/materials/IMaterialRegistry.java
```diff
@@ -170,6 +171,16 @@ default boolean canRepair(MaterialStatsId statsId) {
   default void registerStatType(MaterialStatType<?> type, @Nullable MaterialStatsId traitFallback) {
     registerStatType(type);
   }
+  /**
+   * Clears all dynamic stat types..
+   */
+  void clearDynamicStatTypes();
```
KnightMiner: This should not be publicly exposed. We only need to clear then before stats load.

src/main/java/slimeknights/tconstruct/library/materials/IMaterialRegistry.java
```diff
@@ -170,6 +171,16 @@ default boolean canRepair(MaterialStatsId statsId) {
...
+  /**
+   * Registers a dynamic stat type. Dynamic stat types are stat types that reloads when datapacks reload.
+   * Fallback is not supported for dynamic stat types.
+   * @param type  Stat type
+   */
+  void registerDynamicStatType(FlexMaterialStatType type);
```
KnightMiner: This does not seem necessary. Its loaded in JSON, why must it be registered?

If it has to be registered, a JSON only addon can't use it.

src/main/java/slimeknights/tconstruct/library/materials/MaterialRegistry.java
```diff
@@ -101,6 +103,10 @@ public MaterialRegistry() {
       materialsLoaded = true;
       checkAllLoaded();
     });
+    materialStatTypesLoader = new MaterialStatTypesLoader(() -> {
+      statsLoaded = true;
+      checkAllLoaded();
+    });
```
KnightMiner: Why do we need this? Just put the loading at the start of the material stats loader. You may need to override a method for that. Also ensures stat types are guaranteed to load before we need them instead of being non-deterministic.

Also, setting the same boolean as another loader is just wrong. But since this is the wrong approach don't worry about that part of the comment.

src/main/java/slimeknights/tconstruct/library/materials/stats/IMaterialRegister.java
```diff
@@ -0,0 +1,5 @@
+package slimeknights.tconstruct.library.materials.stats;
+
+public record IMaterialRegister() {
```
KnightMiner: Why is there an empty record named like an interface here? This seems like an accidental commit.

New classes and public methods should have javadocs explaining how they are used.

src/main/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManager.java
```diff
@@ -55,6 +60,12 @@ public class MaterialStatsManager extends MergingJsonDataLoader<Map<ResourceLoca
   @Getter
   private final IdAwareComponentRegistry<MaterialStatType<?>> statTypes = new IdAwareComponentRegistry<>("Unknown Material Stat Type");
 
+  /**
+   * Another registry, cleared on reload.
+   */
+  @Getter
+  private IdAwareComponentRegistry<FlexMaterialStatType> dynamicStatTypes = null;
```
KnightMiner: This probably does not need to be publicly exposed as the other is only exposed so you can register to it.

Part of me thinks it would be better to make this a `Map` instead of the registry as I am not sure its saving much using that class. Feel free to justify though where its making the code easier.

src/main/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManager.java
```diff
@@ -71,10 +82,30 @@ public <T extends IMaterialStats> void registerStatType(MaterialStatType<T> type
     statTypes.register(type);
   }
 
-  /** Gets a lit of all material stat IDs */
+  public void clearAndRecreateDynamicStatTypes() {
```
KnightMiner: again, does not need to be public. Just do this on reload before parsing stats.

src/main/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManager.java
```diff
@@ -71,10 +82,30 @@ public <T extends IMaterialStats> void registerStatType(MaterialStatType<T> type
...
+    dynamicStatTypes = new IdAwareComponentRegistry<>("Unknown Dynamic Material Stat Type");
+  }
+
+    /**
+   * Registers a new material stat type
+   * @param type   Type object
+   */
+  public <T extends IMaterialStats> void registerDynamicStatType(FlexMaterialStatType type) {
```
KnightMiner: Does not need to be registered externally. Load it immediately into the the registry.

src/main/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManager.java
```diff
@@ -71,10 +82,30 @@ public <T extends IMaterialStats> void registerStatType(MaterialStatType<T> type
...
+  }
+
+  /** Gets a list of all material stat IDs */
   public Collection<ResourceLocation> getAllStatTypeIds() {
-    return statTypes.getKeys();
+    Collection<ResourceLocation> newCollection = new ArrayList<>(statTypes.getKeys());
+    if(dynamicStatTypes != null) {
+      newCollection.addAll(dynamicStatTypes.getKeys());
+    }
+    return newCollection;
```
KnightMiner: This method is supposed to be efficient, so cache the result as an immutable collection and clear it when the dynamic collection changes.

src/main/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManager.java
```diff
@@ -84,7 +115,11 @@ public Collection<ResourceLocation> getAllStatTypeIds() {
   @SuppressWarnings("unchecked")
   @Nullable
   public <T extends IMaterialStats> MaterialStatType<T> getStatType(MaterialStatsId id) {
-    return (MaterialStatType<T>) statTypes.getValue(id);
+    MaterialStatType<T> type=(MaterialStatType<T>) statTypes.getValue(id);
+    if(type==null && dynamicStatTypes != null) {
+      type=(MaterialStatType<T>) dynamicStatTypes.getValue(id);
```
KnightMiner: Minor style: add spaces around operators:
```java
if (type == null && dynamicStatTypes != null) {
  type = dynamicStatTypes.getValue(id);
}
```
You can also simplify this by removing the cast until the final return. Make the variable `MaterialStatType<?>` until then.

src/main/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManager.java
```diff
@@ -186,7 +221,8 @@ public UpdateMaterialStatsPacket getUpdatePacket() {
                             .collect(Collectors.toMap(
                               Map.Entry::getKey,
                               entry -> entry.getValue().values()));
-    return new UpdateMaterialStatsPacket(networkPayload);
+    Collection<FlexMaterialStatType> types = dynamicStatTypes.getValues();
```
KnightMiner: just pass this directly into the constructor instead of needing a local variable.

src/main/java/slimeknights/tconstruct/library/materials/stats/UpdateMaterialStatsPacket.java
```diff
@@ -24,12 +27,18 @@ public class UpdateMaterialStatsPacket implements IThreadsafePacket {
...
   public UpdateMaterialStatsPacket(FriendlyByteBuf buffer) {
     this(buffer, MaterialRegistry.getInstance().getStatTypeLoader());
   }
 
   public UpdateMaterialStatsPacket(FriendlyByteBuf buffer, Loadable<MaterialStatType<?>> statTypeLoader) {
+    int statTypeCount = buffer.readInt();
+    IMaterialRegistry instance=MaterialRegistry.getInstance();
+    instance.clearDynamicStatTypes();
+    for(int i = 0;i < statTypeCount;i++)
+      instance.registerDynamicStatType(FlexMaterialStatType.decode(buffer));
```
KnightMiner: this should not be calling a register method. Create the collection or registry and pass the whole collection/registry into the manager.

Probably best to pass it at the same time as the collection of material stats.

src/main/java/slimeknights/tconstruct/library/materials/stats/UpdateMaterialStatsPacket.java
```diff
@@ -39,7 +48,8 @@ public UpdateMaterialStatsPacket(FriendlyByteBuf buffer, Loadable<MaterialStatTy
       for (int j = 0; j < statCount; j++) {
         try {
           MaterialStatType<?> statType = statTypeLoader.decode(buffer);
-          statList.add(statType.getLoadable().decode(buffer, TypedMapBuilder.builder().put(MaterialStatType.CONTEXT_KEY, statType).build()));
+          statList.add(statType.getLoadable().decode(buffer,
+              TypedMapBuilder.builder().put(MaterialStatType.CONTEXT_KEY, statType).build()));
```
KnightMiner, line 60: Don't make unrelated formatting changes

src/main/java/slimeknights/tconstruct/library/materials/stats/UpdateMaterialStatsPacket.java
```diff
@@ -39,7 +48,8 @@ public UpdateMaterialStatsPacket(FriendlyByteBuf buffer, Loadable<MaterialStatTy
       for (int j = 0; j < statCount; j++) {
         try {
           MaterialStatType<?> statType = statTypeLoader.decode(buffer);
```
KnightMiner: this needs to be updated to support the dynamic ones.

src/main/java/slimeknights/tconstruct/library/materials/stats/UpdateMaterialStatsPacket.java
```diff
@@ -60,11 +72,13 @@ public void encode(FriendlyByteBuf buffer) {
...
-   * @param buffer  Buffer instance
-   * @param stat    Stat to encode
+   *
+   * @param buffer Buffer instance
+   * @param stat   Stat to encode
    */
   @SuppressWarnings("unchecked")
-  private <T extends IMaterialStats> void encodeStat(FriendlyByteBuf buffer, IMaterialStats stat, MaterialStatType<T> type) {
+  private <T extends IMaterialStats> void encodeStat(FriendlyByteBuf buffer, IMaterialStats stat,
+      MaterialStatType<T> type) {
```
KnightMiner: again, unnecessary formatting change. Disable autoformat when making PRs.

I do not care about max line length on this repo.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+
+import com.google.common.base.Supplier;
+
+import net.minecraft.network.chat.Component;
+import slimeknights.tconstruct.library.materials.stats.IMaterialStats;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatType;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.Stat;
+import slimeknights.tconstruct.library.tools.stat.ModifierStatsBuilder;
+
+public class FlexMaterialStat implements IMaterialStats {
```
KnightMiner: Missing Javadoc comment.

We also do not use "Flex" as our naming convention. If we have a JSON thing and a non-JSON thing, Dynamic is the best word, so `DynamicMaterialStats` is a decent name.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStatType.java
```diff
@@ -0,0 +1,183 @@
...
+import net.minecraft.world.item.Tiers;
+import slimeknights.mantle.data.loadable.field.RecordField;
+import slimeknights.mantle.data.loadable.record.RecordLoadable;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatType;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatsId;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.FloatStat;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.Stat;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.TierStat;
+
+public class FlexMaterialStatType extends MaterialStatType<FlexMaterialStat> {
```
KnightMiner: DynamicMaterialStatType

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+
+	public FlexMaterialStat(MaterialStatType<?> type, Map<String, Supplier<Stat<?, ?>>> stats, Object... values) {
```
KnightMiner: Why the suppliers? This object is just created from JSON so everything should be loaded by then.

I also don't fancy the `Object...` values, seems like unchecked casts for no real benefit. I think its better to store stats and values together, see `StatBoostModule`.

For the constructor, I would recommend a `List<DynamicStat>`, where `DynamicStat<T>(DynamicStatField<T> field), T value` is a record with interface `DynamicStatField` that handles any important methods (I think that is what your `Stat` is trying to be, but its not named clearly). We can have a `SetStatField`, plus various options for the other float operations (either `MultiplyStatField` and alike or a generic `NumericStatField` with the existing enum from `StatBoostModule`). It will probably need a JSON registry.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+			return list;
+		} catch (Exception e) {
+			return List.of(Component.literal("Error: " + e.getMessage()));
+		}
+	}
+
+	@Override
+	public void apply(ModifierStatsBuilder builder, float scale) {
+		stats.forEach((name, stat) -> {
+			stat.apply(builder, name, scale);
```
KnightMiner: only doing apply is so limited, what about add, multiply, percent, and boost?

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+		});
+		for (int i = 0; i < values.length; i++) {
+			var stat = map.values().toArray(new Stat<?, ?>[0])[i];
+			stat.setValue(values[i]);
+		}
+		this.stats = map;
+	}
+
+	private final MaterialStatType<?> type;
+	private final Map<String, Stat<?, ?>> stats;
```
KnightMiner: Fields go before constructors.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+import slimeknights.tconstruct.library.materials.stats.IMaterialStats;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatType;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.Stat;
+import slimeknights.tconstruct.library.tools.stat.ModifierStatsBuilder;
+
+public class FlexMaterialStat implements IMaterialStats {
+
+	public FlexMaterialStat(MaterialStatType<?> type, Map<String, Supplier<Stat<?, ?>>> stats, Object... values) {
+		this.type = type;
+		var map = new LinkedHashMap<String, Stat<?, ?>>();
```
KnightMiner: We don't use var, it makes the code less readible on github for no gain.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+			var stat = map.values().toArray(new Stat<?, ?>[0])[i];
+			stat.setValue(values[i]);
+		}
+		this.stats = map;
+	}
+
+	private final MaterialStatType<?> type;
+	private final Map<String, Stat<?, ?>> stats;
+
+	public Object get(String name) {
```
KnightMiner: what is the purpose of this getter? New public methods that are not `@Override` need javadocs.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+	public MaterialStatType<?> getType() {
+		return type;
+	}
+
+	@Override
+	public List<Component> getLocalizedInfo() {
+		try {
+			List<Component> list = new ArrayList<>();
+			stats.forEach((name, stat) -> {
+				list.add(stat.format(name));
```
KnightMiner: `stat.format` is limiting as it only lets us do stat sets (see below). Make sure the interface I proposed lets you specify how to format the stat, as we want multipliers to be a thing and percentages.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStat.java
```diff
@@ -0,0 +1,75 @@
...
+			});
+			return list;
+		} catch (Exception e) {
+			return List.of(Component.literal("Error: " + e.getMessage()));
```
KnightMiner: What types of exceptions are you expecting here? Some comments explaining why we need to catch errors in a valid JSON would be useful.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStatType.java
```diff
@@ -0,0 +1,183 @@
...
+
+    @Override
+    public RecordLoadable<FlexMaterialStat> getLoadable() {
+        var fields = new ArrayList<RecordField<Object, FlexMaterialStat>>();
+        stats.forEach((str, sup) -> {
+            fields.add((RecordField) sup.get().getLoadable(str));
+        });
+        switch (fields.size())// 1~15
+        {
+            case 1:
```
KnightMiner: yeah, this is really messy. Just don't use the record loadable. Make a custom implementation of record loadable that takes a list of fields and calls the constructor wanting a list of stats.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/StatTypes.java
```diff
@@ -0,0 +1,200 @@
...
+import slimeknights.tconstruct.library.tools.stat.INumericToolStat;
+import slimeknights.tconstruct.library.tools.stat.IToolStat;
+import slimeknights.tconstruct.library.tools.stat.ModifierStatsBuilder;
+import slimeknights.tconstruct.library.tools.stat.ToolStatId;
+import slimeknights.tconstruct.library.tools.stat.ToolStats;
+import slimeknights.tconstruct.library.tools.stat.ToolTierStat;
+
+public class StatTypes {
+    public enum Operator {
+        UPDATE, ADD, PERCENT, MULTIPLY, MULTIPLY_ALL;
```
KnightMiner: Hardcoded enum is not a good design. We need addons to be able to add their own stat types.

Define a list of all operations a field in the JSON needs to do in an interface. Things like update the stat on the tool, get the description, and parse. Then implement that for various types.

You can have an enum for the numeric implementation. Also make a generic set implementation for things like mining tier.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypesLoader.java
```diff
@@ -0,0 +1,73 @@
...
+      counter--;
+      log.error("Error parsing material stat type {}: {}", id, e.getMessage());
+      builder.shouldBuild(false);
+    }
+  }
+
+  @Override
+  protected void finishLoad(Map<ResourceLocation,MaterialStatTypeBuilder> map, ResourceManager manager) {
+    // Take the final structure and actually load the different material stats. This drops all invalid stats
+    IMaterialRegistry materialRegistry = MaterialRegistry.getInstance();
```
KnightMiner: As I stated earlier, this all belongs in the top of the material stats manager. That is why I told you to put the field there.

Right now, we have no idea when this will load. Putting it at the top gives us a guarantee. You might have to override a new method to do it.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/NoSuchToolStatException.java
```diff
@@ -0,0 +1,9 @@
+package slimeknights.tconstruct.library.materials.stats.types;
+
+import slimeknights.tconstruct.library.tools.stat.ToolStatId;
+
+class NoSuchToolStatException extends RuntimeException {
```
KnightMiner: What is this exception for? Needs javadocs.

We really don't use custom exceptions though without good reason. Just use json parse or whatever other related exception is most accurate.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/StatTypes.java
```diff
@@ -0,0 +1,200 @@
...
+                case PERCENT:
+                    return 2;
+                case MULTIPLY:
+                    return 3;
+                case MULTIPLY_ALL:
+                    return 4;
+            }
+            return -1;
+        }
+        public static Operator fromIndex(byte index) {
```
KnightMiner: Network buffer has helpers for reading and writing enums, no need for these.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/StatTypes.java
```diff
@@ -0,0 +1,200 @@
...
+                case 3:
+                    return MULTIPLY;
+                case 4:
+                    return MULTIPLY_ALL;
+            }
+            throw new IndexOutOfBoundsException("No operator with index " + index);
+        }
+    }
+
+    public static abstract class Stat<T, S extends IToolStat<T>> {
```
KnightMiner: This should really be two things:

`DynamicStatField`: interface containing logic for parsing the stat. Has a loader registry where you can register field types, such as numeric, harvest tier, and alike. As I think about it, might be best to separate multiplier, boost, and alike into their own impls as they all have different validation ranges and different ways they want to do tooltips.
`DynamicStat`: record that holds a `DynamicStatField` and its value. Used in the constructor for `DynamicStatType`

Note these names are not final, you can name them something else if its clear.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/StatTypes.java
```diff
@@ -0,0 +1,200 @@
...
+        public T value;
+
+        public final Operator operator;
+
+        public T getValue() {
+            return value;
+        }
+
+        public abstract String getStatType();
```
KnightMiner: And again, javadocs on all new public or protected methods.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStatType.java
```diff
@@ -0,0 +1,183 @@
...
+import slimeknights.tconstruct.library.materials.stats.MaterialStatType;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatsId;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.FloatStat;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.Stat;
+import slimeknights.tconstruct.library.materials.stats.types.StatTypes.TierStat;
+
+public class FlexMaterialStatType extends MaterialStatType<FlexMaterialStat> {
+
+    public FlexMaterialStatType(MaterialStatsId id, boolean canRepair,
+            LinkedHashMap<String, Supplier<Stat<?, ?>>> stats) {
```
KnightMiner: This should take a `List<DynamicStatField>` in the constructor.

Also, `canRepair` should be computed from that list, not passed in separately.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FlexMaterialStatType.java
```diff
@@ -0,0 +1,183 @@
...
+        super(id, (FlexMaterialStat) null, null);
```
KnightMiner: I really don't like passing in null to super here, we should be able to create the default from the `List<DynamicStatField>`, by giving each field a default value.

src/test/java/slimeknights/tconstruct/fixture/MaterialRegistryFixture.java
```diff
@@ -96,6 +97,15 @@ public void registerStatType(MaterialStatType<?> type) {
...
   }
 
+  @Override
+  public void clearDynamicStatTypes() {
+  }
+  
+  @Override
+  public void registerDynamicStatType(FlexMaterialStatType type) {
+    throw new UnsupportedOperationException("No registration possible in test mock");
+  }
```
KnightMiner: These need not even exist in our tests. Entirely internal to the stats manager

src/main/java/slimeknights/tconstruct/library/materials/stats/types/ToolStatTypeNotMatchException.java
```diff
@@ -0,0 +1,9 @@
+package slimeknights.tconstruct.library.materials.stats.types;
+
+import slimeknights.tconstruct.library.tools.stat.ToolStatId;
+
+class ToolStatTypeNotMatchException extends RuntimeException {
```
KnightMiner: Again, why the custom exception? Looks to me like another json parse exception.

src/test/java/slimeknights/tconstruct/library/materials/stats/UpdateMaterialStatsPacketTest.java
```diff
@@ -35,7 +37,8 @@ void testGenericEncodeDecode() {
       MATERIAL_ID, List.of(MaterialStatsFixture.MATERIAL_STATS)
     );
 
-    UpdateMaterialStatsPacket packetToDecode = sendAndReceivePacket(materialToStats);
+    Collection<FlexMaterialStatType> dynamic = new HashSet<>();// I dont know how to test so just an empty set
```
KnightMiner: The way you test is you write JSON for custom stat types, and then use them to load stats with those types.

From the packet, make sure:
* The stat type you send in comes back out the same
* stats using the stat type you send in come back out the same

But it also needs tests for JSON parsing:
* make sure you can parse a stat type as expected.
* make sure stats using the custom stat type parse as expected.

KnightMiner (changes requested):
This has the start of something that could work, but it does seem you made some design decisions that will make this very hard to maintain. I left some comments proposing a cleaner design. Feel free to ping me on discord and we can discuss it more in depth.

Qikahome:
I wrote the code for the test part, but it seems that ToolStats wasn't registered in the test, which caused my code to throw a JsonParseException.

Qikahome:
[newtestpack.zip](https://github.com/user-attachments/files/24529758/newtestpack.zip)

Qikahome:
Tests are completed

src/test/java/slimeknights/tconstruct/fixture/MaterialStatTypesFixture.java
```diff
@@ -1,23 +1,32 @@
...
-  public static final MaterialStatsId STATS_TYPE = new MaterialStatsId("test", "mat_stat_type_1");
+  public static final MaterialStatsId STATS_TYPE = new MaterialStatsId("test", "dynamic");
 
+  public static final FloatToolStat TestFloatStat = ToolStats.register(new FloatToolStat(new ToolStatId(TConstruct.MOD_ID, "test_stat"), 0xFF47CC47, 1, 1, Integer.MAX_VALUE, TinkerTags.Items.DURABILITY));
+  public static final ToolTierStat TestTierStat = ToolStats.register(new ToolTierStat(new ToolStatId(TConstruct.MOD_ID, "test_tier_stat")));
+  
   public static final List<DynamicStatField<?>> statFields = List.of(
-    new FloatDynamicStatField("test1",ToolStats.DURABILITY,1f,FloatDynamicStatField.Operation.PERCENT, "desc1", "info1"),
-    new FloatDynamicStatField("test2",ToolStats.MINING_SPEED,2f,FloatDynamicStatField.Operation.UPDATE, "desc2", "info2"),
-    new TierDynamicStatField("test3",ToolStats.HARVEST_TIER,Tiers.STONE,"desc3")
```
KnightMiner: Worth noting, we do use these stats in other tests. You might just need to ensure static init is run before the deserialize method runs.

But yeah, fake stats are fine for the tests. I sometimes use `"test"` as a namespace for them to distinguish them.

src/test/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManagerTest.java
```diff
@@ -13,12 +18,16 @@
...
+      .makeType(STATS_ID_SIMPLE);
   private static final MaterialStatsId STATS_ID_DONT_CARE = new MaterialStatsId("dont", "care");
 
-  private final MaterialStatsManager materialStatsManager = new MaterialStatsManager(() -> {});
+  private final MaterialStatsManager materialStatsManager = new MaterialStatsManager(() -> {
+  });
   private final MergingJsonFileLoader<?> fileLoader = new MergingJsonFileLoader<>(materialStatsManager);
-  private final MergingJsonFileLoader<?> anoFileLoader = new MergingJsonFileLoader<>(materialStatsManager.getStatTypesLoader());
+  private final MergingJsonFileLoader<?> anoFileLoader = new MergingJsonFileLoader<>(
+      materialStatsManager.getStatTypesLoader());
```
KnightMiner: Again, turn off your auto formatter. You should not be reformatting lines you did not change.

You should revert any formatting changes that are committed to minimize noise in the review.

Qikahome: Reverted, and all unrelated formatting changes removed.

Qikahome:
May I ask how this PR is going?

KnightMiner:
This is not the only thing I have to work on nor the only PR I have to review right now. You don't have to keep commenting on it requesting an update.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStat.java
```diff
@@ -0,0 +1,69 @@
...
+
+	/**
+	 * Gets a stat by name.
+	 * @see DynamicStatField need this to encode and serialize
+	 * @param name  Stat name
+	 * @return  Stat, or null if not found
+	 */
+	@Nullable
+	public DynamicStat getStat(String name)
+	{
```
KnightMiner: Minor, but be consistent on formatting. In this case, brackets go on same line, not newline.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStat.java
```diff
@@ -0,0 +1,69 @@
...
+
+
+	/**
+	 * Gets a stat by name.
+	 * @see DynamicStatField need this to encode and serialize
+	 * @param name  Stat name
+	 * @return  Stat, or null if not found
+	 */
+	@Nullable
+	public DynamicStat getStat(String name)
```
KnightMiner: I understand what this method does from the javadoc, but not why it is needed. Does any other class use it or is it just used internally?

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStat.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	@Override
+	public MaterialStatType<?> getType() {
+		return type;
+	}
+
+	@Override
+	public List<Component> getLocalizedInfo() {
+		return stats.values().stream().map(stat->stat.getLocalizedInfo()).collect(Collectors.toList());
```
KnightMiner: Probably worth caching these two lists as part of the constructor since they are not changing.

Also, you can use `DynamicStat::getLocalizedInfo` for the map call, IDEA has a warning in place if you don't do that so I tend to do it.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStat.java
```diff
@@ -0,0 +1,69 @@
...
+	}
+
+	@Override
+	public List<Component> getLocalizedDescriptions() {
+		return stats.values().stream().map(stat->stat.getLocalizedDescription()).collect(Collectors.toList());
+	}
+
+	@Override
+	public void apply(@Nonnull ModifierStatsBuilder builder, float scale) {
+		stats.values().forEach(stat -> {
```
KnightMiner: Similarly to above, IDEA is configured to warn of single line lambda statements; can make this compact as `stat -> stat.apply(builder, scale)`

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatRecord.java
```diff
@@ -0,0 +1,48 @@
...
+import slimeknights.mantle.data.loadable.record.RecordLoadable;
+import slimeknights.mantle.util.typed.TypedMap;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatType;
+import slimeknights.tconstruct.library.materials.stats.types.DynamicStatField.DynamicStat;
+
+/**
+ * A record class that represents a dynamic material stat.
+ */
+public record DynamicMaterialStatRecord(MaterialStatType<?> type, List<DynamicStatField<?>> statFields)
+        implements RecordLoadable<DynamicMaterialStat> {
```
KnightMiner: There is not a line length limit for this repository; keep implements on the same like as the class declaration unless its seriously long (in which case I tend to prefer splitting the record constructor instead).

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatRecord.java
```diff
@@ -0,0 +1,48 @@
...
+import com.google.gson.JsonObject;
+
+import net.minecraft.network.FriendlyByteBuf;
+import slimeknights.mantle.data.loadable.record.RecordLoadable;
+import slimeknights.mantle.util.typed.TypedMap;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatType;
+import slimeknights.tconstruct.library.materials.stats.types.DynamicStatField.DynamicStat;
+
+/**
+ * A record class that represents a dynamic material stat.
```
KnightMiner: This name is pretty vague. Yes, its a record, but a record is just a type of object (in JSON context, it means it serializes to a JSON object).

The important word in `RecordLoadable` is `Loadable`, or in some contexts we say `Loader`. So `DynamicMaterialStatsLoader`.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStat.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraft.network.chat.Component;
+import slimeknights.tconstruct.library.materials.stats.IMaterialStats;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatType;
+import slimeknights.tconstruct.library.materials.stats.types.DynamicStatField.DynamicStat;
+import slimeknights.tconstruct.library.tools.stat.ModifierStatsBuilder;
+
+/**
+ * A material stat that has dynamic stat fields.
+ */
+public class DynamicMaterialStat implements IMaterialStats {
```
KnightMiner: Interface is `IMaterialStats`, and this contains multiple stats, so `DynamicMaterialStats` is my recommendation.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicStatField.java
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Gets the decoder for the given ID, or throws an JsonParseException if not
+     * found
+     * 
+     * @param id ID of the decoder
+     * @return Decoder for the ID
+     */
+    @NonNull
+    public static DynamicStatDecoder<?> getDecoderOrThrow(ResourceLocation id) {
```
KnightMiner: This should not be necessary. We have methods on `IdAwareCOmponentRegistry` for this sort of thing.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicStatField.java
```diff
@@ -0,0 +1,188 @@
...
+
+    /**
+     * Deserialize the stat field from JSON
+     * 
+     * @param json JSON object
+     * @return Deserialized stat field
+     */
+    @NonNull
+    public static DynamicStatField<?> deserializeSelf(JsonObject json, ResourceLocation path) {
+        String statType = json.get("type").getAsString();
```
KnightMiner: `GsonHelper.getAsString`, throws the proper exception type.

Though like above, just use the existing helpers.

src/main/java/slimeknights/tconstruct/library/materials/stats/dynamic/DynamicStatField.java
```diff
@@ -0,0 +1,188 @@
...
+import lombok.NonNull;
+import net.minecraft.network.FriendlyByteBuf;
+import net.minecraft.network.chat.Component;
+import slimeknights.mantle.data.registry.IdAwareComponentRegistry;
+import slimeknights.mantle.registration.object.IdAwareObject;
+import slimeknights.tconstruct.TConstruct;
+import slimeknights.tconstruct.library.tools.stat.ModifierStatsBuilder;
+import net.minecraft.resources.ResourceLocation;
+
+public interface DynamicStatField<S extends DynamicStatField.DynamicStat> {
```
KnightMiner: I feel like a lot of this is recreating features we already have with Loadable. You can create a `GenericLoaderRegistry` for `DynamicStat` to parse it instead of needing to recreate the whole loader system from scratch.

We basically have two things here:
* The `DynamicStat` object which contains a stat, operation, and alike that are loaded.
* Loaders for `DynamicStat`, those can just be `RecordLoadable<T extends DynamicStat>`

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatRecord.java
```diff
@@ -0,0 +1,48 @@
...
+ */
+public record DynamicMaterialStatRecord(MaterialStatType<?> type, List<DynamicStatField<?>> statFields)
```
KnightMiner: Name should be `DynamicMaterialStatLoader`. Saying its a record is not descriptive and does not distinguish it from `DynamicMaterialStats`. But saying its a loader (the important part of `RecordLoadable` does.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStat.java
KnightMiner: Interface is `IMaterialStats`, this should be `DynamicMaterialStats`

KnightMiner: Interface is `IMaterialStats`, so this should be called `DynamicMaterialStats`

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatRecord.java
KnightMiner: `Record` is not the informative part of `RecordLoadable`, its `Loadable`. Current class name does nothing to distinguish it from `DynamicMaterialStats` which is also a record functionally.

`DynamicMaterialStatsLoader` or `DynamicMaterialStatsLoadable.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatType.java
```diff
@@ -0,0 +1,83 @@
...
+    /**
+     * Constructs a dynamic material stat type.
+     * 
+     * @param id          The material stats ID.
+     * @param canRepair   Whether the tool with this material stat type can be repaired.
+     * @param statFields  The dynamic stat fields.
+     */
+    public DynamicMaterialStatType(MaterialStatsId id, boolean canRepair, List<DynamicStatField<?>> statFields) {
+        super(id, new DynamicMaterialStat(null, new LinkedHashMap<>()), new DynamicMaterialStatRecord(null, statFields));
+            // I have no way to make all these parameters nonnull.
```
KnightMiner: Parameters are nonnull by default. Copy the package info to new packages.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatType.java
```diff
@@ -0,0 +1,83 @@
+package slimeknights.tconstruct.library.materials.stats.types;
```
KnightMiner: Package should really be `materials.stats.dynamic`, this is not where all types live and most of these classes are not types.

src/main/java/slimeknights/tconstruct/library/materials/stats/dynamic/DynamicStatField.java
KnightMiner: This whole class seems to just be recreating Loadable in a way that requires writing serializers in the redundant error prone way. Just use GenericLoaderRegistry<DynamicStat here.

Your DynamicStatField is almost just a RecordLoadable<DynamicStat>. I would recommend doing `record DynamicStatField<T extends DynamicStat>(RecordLoadable<T> loader, String name)` to save a lot of duplicate serializer code.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FloatDynamicStatField.java
```diff
@@ -0,0 +1,167 @@
...
+    @AllArgsConstructor
+    public static class FloatDynamicStat implements DynamicStat {
+
+        private final FloatToolStat stat;
+        @Getter
+        private final float value;
+        private final Operation operation;
+        @Getter
+        private final Component localizedDescription;
+        private final String localizedInfoPrefix;
```
KnightMiner: This is a key for a language file. That is by definition unlocalized.

Call it something like `tooltipKeyPrefix` or simply `tooltipKey`; the name `localizedInfo` is kinda vague in retrospective.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/FloatDynamicStatField.java
```diff
@@ -0,0 +1,167 @@
...
+                case MULTIPLY_ALL -> stat.multiplyAll(builder, value);
+            }
+        }
+
+        @Override
+        public Component getLocalizedInfo() {
+            if (operation != Operation.PERCENT) {   
+                return stat.formatValue(value);
+            }
+            return IToolStat.formatColoredPercentBoost(localizedInfoPrefix, value);
```
KnightMiner: Not a fan of how `localizedInfoPrefix` is just unused for non-precents here. Just make it check if its empty or not; let any stat override it.

If a prefix is not provided, use `formatValue`. If its provided choose the appropriate type of `formatColored` for the operation chosen.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+import net.minecraft.resources.ResourceLocation;
+import slimeknights.tconstruct.library.materials.stats.MaterialStatsId;
+
+
+public class MaterialStatTypeBuilder {
+
+    private final ResourceLocation registryName;
+    private List<DynamicStatField<?>> fields = new ArrayList<>();
+    @Setter
+    private boolean canRepair = false;
```
KnightMiner: This should not be a setter. A material is only repariable if it implements `IRepairableMaterialStats` and thus has a method to return `durability` as an int. The getter was just a shortcut to instanceof as its a common query.

To really do that, you basically need to create two versions of `DynamicMaterialStats`, one that implements `IRepariableMaterialStats` and one that does not. One way to do that is setting a string here, which if not empty creates an add durability field for you that is returned by that method.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatRecord.java
```diff
@@ -0,0 +1,48 @@
...
+        implements RecordLoadable<DynamicMaterialStat> {
+
+    @Override
+    public DynamicMaterialStat decode(FriendlyByteBuf buffer, TypedMap context) {
+        Map<String, DynamicStat> stats = new LinkedHashMap<>();
```
KnightMiner: I generally prefer immutable datastructures for the underlying data for these sorts of things. So `ImmutableMap.Builder` is preferable.

That said, I've not seen anything here that could not be done as a list.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStatRecord.java
```diff
@@ -0,0 +1,48 @@
...
+        statFields.forEach(field -> stats.put(field.getName(), field.decode(buffer)));
+        return new DynamicMaterialStat(type, stats);
+    }
+
+    @Override
+    public void encode(FriendlyByteBuf buffer, DynamicMaterialStat value) {
+        for (DynamicStatField<?> field : statFields) {
```
KnightMiner: For encoding, iterate the two lists together. Whatever is at index `i` in the fields list will match what is at index `i` in the values list.

No need to give the entire `DynamicMaterialStats` object to the field. Just give it the one `DynamicStat` it cares about.

You may need to deal with some generics casting to make that work. Feel free to message me in `#coding` to figure out how to write that.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+    @Getter
+    @Setter
+    @Accessors(fluent = true)
+    private boolean shouldBuild = true;
```
KnightMiner: These fields with `@Setter` need javadocs explaining what they do. I really don't know what `shouldBuild` means for instance. Also unsure why this is `@Getter`

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+        this.registryName = registryName;
+    }
+
+    /**
+     * Creates a new material stat type builder.
+     * 
+     * @param registryName  The registry name of the material stat type.
+     * @return  The created material stat type builder.
+     */
+    public static MaterialStatTypeBuilder begin(
```
KnightMiner: We don't use `begin` here. We use `builder()` or if its taking a single named parameter something like `name` is a good way to start.

You can use `@RequiredArgsConstructor(staticName="name")` to generate this and the constructor above automatically.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+            ResourceLocation registryName) {
+        return new MaterialStatTypeBuilder(registryName);
+    }
+
+    /**
+     * Adds a stat field to the material stat type.
+     * 
+     * @param field The stat field to add.
+     */
+    public void addField(DynamicStatField<?> field) {
```
KnightMiner: Builders should return themselves for method chaining.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+
+public class MaterialStatTypeBuilder {
+
+    private final ResourceLocation registryName;
+    private List<DynamicStatField<?>> fields = new ArrayList<>();
+    @Setter
+    private boolean canRepair = false;
+    @Getter
+    @Setter
+    @Accessors(fluent = true)
```
KnightMiner: Builders should chain, use `chain=true`.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypeBuilder.java
```diff
@@ -0,0 +1,57 @@
...
+    }
+    
+    /**
+     * Builds the material stat type.
+     * 
+     * @param resource The resource location to use for the material stat type.
+     * @return The built material stat type.
+     */
+    public DynamicMaterialStatType build() {
+        return new DynamicMaterialStatType(new MaterialStatsId(registryName), this.canRepair, fields);
```
KnightMiner: Might as well make the field be `MaterialStatsId` instead of waiting to do so. Add `name` overload for `ResourceLocation` if you wish.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypesLoader.java
```diff
@@ -0,0 +1,82 @@
...
+ * first one encountered will be used, and the second one will be skipped.
+ * (e.g. having a 'Laser' stat type, and there are 2 mods who add Laser stat
+ * types to the iron material)
+ * <p>
+ * The location inside datapacks is "materials/stat_types".
+ * So if the material's mod name is "foobar", the location for your material's
+ * stat types is "data/foobar/materials/stat_types".
+ */
+@Log4j2
+public class MaterialStatTypesLoader extends MergingJsonDataLoader<MaterialStatTypeBuilder> {
```
KnightMiner: Loaders should not be merging, we do not need multiple sources of what a stat type looks like.

Take a look at `MaterialManager` for reference.

src/main/java/slimeknights/tconstruct/library/materials/stats/MaterialStatsManager.java
```diff
@@ -154,9 +169,17 @@ public void updateMaterialStatsFromServer(Map<MaterialId, Collection<IMaterialSt
...
           )))
       );
+    allStatTypeIds = new ArrayList<>(statTypes.getKeys());
+    if (statTypesLoader.getStatTypes() != null) {
+      allStatTypeIds.addAll(statTypesLoader.getStatTypes().keySet());
+    }
     onLoaded.run();
   }
 
+  public void setDynamicStatTypes(Map<MaterialStatsId, DynamicMaterialStatType> dynamicStatTypes) {
```
KnightMiner, line 183: Needs a javadoc. I assume this is set from the packet? If so, an `@Internal` is a good idea too.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/MaterialStatTypesLoader.java
```diff
@@ -0,0 +1,82 @@
...
+  public static final String FOLDER = "tinkering/materials/stat_types";
+
+  /** Map of material stat types, keyed by material ID */
+  @Getter
+  @Setter
+  private Map<MaterialStatsId, DynamicMaterialStatType> statTypes;
```
KnightMiner: This being nullable is just leading to a bunch of redundant checks. Just make it default to `Map.of()` (empty map).

KnightMiner (changes requested):
My main comment on this is you seem to be reinventing the wheel in how `DynamicStatField` works. You can simplify that a lot if you take advantage of the `Loadable` API we already use elsewhere.

Feel free to talk to me on `#coding` on our discord if you have questions about how to do that.

src/main/java/slimeknights/tconstruct/library/materials/stats/types/DynamicMaterialStat.java
Qikahome: In network transmission or serialization, we need specific property values, but it's not feasible to create a getter for every DynamicStat.
So I get the value from the name.

Qikahome:
After some research, I found that RecordLoadable is much simpler than reinventing the wheel. I have refactored my code and am now using RecordLoadable.

PR #5609, opened 2026-01-25
Russian Localization Update (1.20.1)

deaddiesel:
- Updated Russian localization (ru_ru.json file).
- Reworked and updated book localization.
- Added Russian book covers.


src/main/resources/assets/tconstruct/book/materials_and_you/ru_ru/appearance.json
```diff
@@ -0,0 +1,10 @@
+{
+  "coverTexture": "tconstruct:book/materials_and_you/ru_ru/images/materials_and_you_cover.png",
```
KnightMiner: This path is wrong, it should be `tconstruct:book/images/covers/ru_ru/materials_and_you_cover.png` like I told the other person wanting to do the same. I would much rather all images are kept together than scattering them.

Same comment for the other 5 books.

deaddiesel: I moved the covers and changed the paths.

PR #5663, opened 2026-04-08
Skip circle attack on extra attacks

### LINKED ISSUE #5662: circle attack will attack more mobs
### Minecraft Version

1.20.1

### Modloader

Forge

### Modloader Version

47.4.0

### Mantle Version

1.11.104

3.11.2.201

### Describe your issue

Circle attack doesn't check extra attack, so it will run multiple times.

### Steps to reproduce

Watch the code.

### Crash Report

_No response_

Yes

### Performance Enchancers

None of the above

### Other mods

_No response_

I did not search

utafrali:
The circle attack was triggering on extra attack procs (like from enchantments), which caused it to hit way more mobs than intended. This just adds a check to skip the circle attack entirely when it's an extra attack.

Fixes #5662

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/tools/definition/module/weapon/CircleWeaponAttack.java
+++ b/src/main/java/slimeknights/tconstruct/library/tools/definition/module/weapon/CircleWeaponAttack.java
@@ -37,6 +37,7 @@ public record CircleWeaponAttack(float diameter) implements MeleeHitToolHook, To
 
   @Override
   public void afterMeleeHit(IToolStackView tool, ToolAttackContext context, float damage) {
+    if (context.isExtraAttack()) return;
     // no need for fully charged for scythe sweep, easier than sword sweep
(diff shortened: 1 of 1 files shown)


gjhihihi:
More scenarios need to be considered. For example, bonking is always an extra attack, so SweepWeaponAttack.java needs similar change.

KnightMiner:
Bonking was never meant to do AOE attacks on scythes, which is why it was marked as an extra attack. The scythe sweep comes from you swinging it, bonk is a whack. Or from a gameplay perspective, why would it damage mobs that are not knocked back?

gjhihihi:
My point is that the current circle attack can be triggered by bonking, which is another reason to skip the extra attack; and I think the same issue applies to the sweep attack.

utafrali:
Good point on the circle attack. I'll defer to KnightMiner's call though - bonking's meant to work differently.

utafrali:
Yeah, fair point about the circle attack - but KnightMiner's right that bonk's meant to work different from sweep.

PR #5668, opened 2026-04-28
Fix issues with TextData lists that aren't prefixed

paypur:
~~3~~ 6 lines


paypur:
maybe these should be moved into a helper function

KnightMiner:
tbh, the margin top should probably be a CSS class. But yeah, helper function is also reasonable.

paypur:
or maybe `HtmlElement` with a string or span descendant should just automatically wrap that with `p` when you serialize it

KnightMiner:
I'd rather not hardcode that. Its a style preference to wrap it.

PR #5677, opened 2026-05-30
render traveller accessories from Baubles slots when Tinker tab is disabled

Eldrinn-Elantey:
When the Tinker inventory tab is disabled (default after #279), accessories go into Baubles-Expanded slots but weren't rendered on the player because `renderArmorExtras` still read from `armorExtended.inventory`. Fixed by reading from Baubles inventory when the tab is disabled, keeping the original path when it's enabled.


KnightMiner:
We have no inventory tab, and we don't do Baubles. I assume this is another PR meant for one of the old version forks.

Eldrinn-Elantey:
> We have no inventory tab, and we don't do Baubles. I assume this is another PR meant for one of the old version forks.

Sorry. This is my first time using the gh command-line tool to create a pull request.

PR #5700, opened 2026-07-30
3D Tool Model with modifiers

Qikahome:
[testResourcePack.zip](https://github.com/user-attachments/files/30549061/testResourcePack.zip)
It replaces the standard pickaxe model with a block_tool_model to help evaluate the visual behavior of the following modifiers:
tconstruct:diamond
tconstruct:emerald
tconstruct:haste
tconstruct:gilded
tconstruct:netherite
tconstruct:tank
For these modifiers, I've added some basic/placeholder models so you can easily test and check the rendering logic in-game. Feel free to use this pack to validate those specific modifier interactions.


src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+import slimeknights.tconstruct.library.materials.definition.MaterialVariantId;
+import slimeknights.tconstruct.library.modifiers.ModifierEntry;
+import slimeknights.tconstruct.library.modifiers.ModifierId;
+import slimeknights.tconstruct.library.recipe.worktable.ModifierSetWorktableRecipe;
+import slimeknights.tconstruct.library.tools.nbt.IToolStackView;
+import slimeknights.tconstruct.library.tools.nbt.MaterialIdNBT;
+import slimeknights.tconstruct.library.tools.nbt.ModifierNBT;
+import slimeknights.tconstruct.library.tools.nbt.ToolStack;
+
+@SuppressWarnings({ "null" })
```
KnightMiner: It should not be necessary to suppress warning for nullability; this usually indicates something is not properly doing null checks and should be doing them.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+@RequiredArgsConstructor(access = AccessLevel.PRIVATE)
+@Log4j2
```
KnightMiner: I've been moving away from the logger annotation in favor of just using `TConstruct.log`.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+
+  public static BlockToolModel deserialize(JsonObject json, JsonDeserializationContext context) {
+    SimpleBlockModel model = SimpleBlockModel.deserialize(json, context);
+    List<Set<String>> parts = PARTS.getIfPresent(json, "parts");
+    // modifier root fetching
+    List<ResourceLocation> modifierModels = List.of();
+    if (json.has("modifier_maps")) {
+      modifierModels = JsonHelper.parseList(json, "modifier_maps", Loadables.RESOURCE_LOCATION);
+    }
+    boolean showTraits = GsonHelper.getAsBoolean(json, "show_traits", false);
```
KnightMiner: Show traits should not be needed anymore - the modifier map system replaces that with the far cleaner constant models.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+  @Override
+  public BakedModel bake(IGeometryBakingContext owner, ModelBaker baker, Function<Material, TextureAtlasSprite> spriteGetter, ModelState transform, ItemOverrides overrides, ResourceLocation location) {
+    BakedModel baked = model.bake(owner, baker, spriteGetter, transform, overrides, location);
+    // load modifier models
+    ModifierModelMap modifierModels = BlockModifierModelMapManager.INSTANCE.getModelsForTool(spriteGetter, modelLocation);
+    boolean particleRetextured = parts.stream().anyMatch(set -> set.contains("particle"));
+    return new BakedBlockTool(baked, owner, model, transform, parts, particleRetextured, overrides, modifierModels, showTraits);
+  }
+
+  private static BakedModel bakeInternal(Function<Material, TextureAtlasSprite> spriteGetter, BakedBlockTool baked, ModifierModelMap modifierModels, boolean showTraits, MaterialIdNBT materials, @Nullable IToolStackView tool) {
```
KnightMiner: This method relies far too much on `BakedBlockTool` to not just be an instance method there. Would save passing in any parameters that are a part of that model.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+    Transformation transforms = Transformation.identity();
+    ItemLayerPixels pixels = new ItemLayerPixels();
```
KnightMiner: Pixels are only relevant to item layers, they should not be needed for a block model.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+    SimpleBakedModel.Builder modelBuilder = baked.getBuilderWithMaterials(materials);
+    if (tool != null && !modifierModels.isEmpty()) {
+      // reversed order to ensure the pixels is updated correctly, not required for 3d tools then
```
KnightMiner: This comment is confusing. 3D tools don't need reverse order so why bring it up? You should go in order for modifier models as we don't need to reverse quad order to handle z-fighting.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+                List<BakedQuad> quads = new ArrayList<>();
+                model.addQuads(tool, entry, spriteGetter, transforms, false, 0, quads::addAll, pixels);
+                return quads;
+              }).forEach(modelBuilder::addUnculledFace);
+            }
+          }
+        }
+      }
+      // iterate the constant models in order for simplicity, you can just reverse the order yourself I guess
+      for (ModifierModel model : modifierModels.constant().values()) {
```
KnightMiner: Constant modifiers should run first. They are last on the tool model because the order of quads is later reversed.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+        Set<ModifierId> hidden = ModifierSetWorktableRecipe.getModifierSet(tool.getPersistentData(), TConstruct.getResource("invisible_modifiers"));
+        for (int i = modifiers.size() - 1; i >= 0; i--) {
+          ModifierEntry entry = modifiers.get(i);
+          ModifierId modifier = entry.getModifier().getId();
+          if (!hidden.contains(modifier)) {
+            IBakedModifierModel model = modifierModels.get(modifier);
+            if (model != null) {
+              baked.getCachedModifierModel(model.getCacheKey(tool, entry), () -> {
+                List<BakedQuad> quads = new ArrayList<>();
+                model.addQuads(tool, entry, spriteGetter, transforms, false, 0, quads::addAll, pixels);
```
KnightMiner: This is not stable. The fact it takes the tool instance as a parameter means this cache will be wrong the next time we encounter a tool with the same modifier.

Its probably best to just not cache the modifier model quads here. Just pull them directly as needed.

src/main/java/slimeknights/tconstruct/library/client/model/tools/BlockToolModel.java
```diff
@@ -0,0 +1,415 @@
...
+        }).forEach(modelBuilder::addUnculledFace);
+      }
+    }
+    return modelBuilder.build(SimpleBlockModel.getRenderTypeGroup(baked.owner));
+  }
+
+  private static class BakedBlockTool extends DynamicBakedWrapper<BakedModel> {
+    /** Dynamic overrides that rebake the model with material and modifier data from the ItemStack */
+    // extends NestedOverrides so an override hit (e.g. placed model switching) also re-resolves the target model's
+    // own overrides (guarded by ignoreNested), applying its material/modifier baking instead of dropping it
```
KnightMiner: Regular comments should not be between javadocs and the javadoc target. Either make it a multiline javadoc and include them, or move them to a more relevant location.

src/main/java/slimeknights/tconstruct/library/client/modifiers/block/model/BlockModifierModel.java
```diff
@@ -0,0 +1,62 @@
...
+import slimeknights.mantle.data.loadable.record.SingletonLoader;
+import slimeknights.mantle.data.registry.GenericLoaderRegistry;
+import slimeknights.mantle.util.ItemLayerPixels;
+import slimeknights.tconstruct.library.client.modifiers.model.ModifierModel;
+import slimeknights.tconstruct.library.json.TinkerLoadables;
+import slimeknights.tconstruct.library.modifiers.ModifierEntry;
+import slimeknights.tconstruct.library.tools.nbt.IToolStackView;
+
+/** Represents a 3d model defined for the given tool */
+public interface BlockModifierModel extends ModifierModel {
```
KnightMiner: This should not extend `ModifierModel`. That is going to cause confusion thinking this works on item layer tools when it should be specialized to the needs of the 3D tool.

src/main/java/slimeknights/tconstruct/library/json/TinkerLoadables.java
```diff
@@ -91,4 +103,57 @@ public static <B, T> StringLoadable<T> instance(StringLoadable<B> loadable, Clas
...
     }, t -> (B)t);
   }
+
+  
+  /** Models */
+  public static final Loadable<BlockElement> BLOCK_ELEMENT = new GsonLoadable<>(new GsonBuilder().registerTypeAdapter(BlockElement.class, new BlockElement.Deserializer()).registerTypeAdapter(BlockElementFace.class, new BlockElementFace.Deserializer()).registerTypeAdapter(BlockFaceUV.class, new BlockFaceUV.Deserializer()).create(), BlockElement.class);
+  public static final Loadable<Transformation> TRANSFORMATION = new CodecLoadable(Transformation.CODEC);
+
+  /** Generic */
+  public abstract static class EitherLoadable<A, B> implements Loadable<Either<A, B>> {
```
KnightMiner: You should not be implementing an entire class in `TinkerLoadables`. This is meant for single 1 line implementations. Just make this its own class.

src/main/java/slimeknights/tconstruct/library/json/TinkerLoadables.java
```diff
@@ -91,4 +103,57 @@ public static <B, T> StringLoadable<T> instance(StringLoadable<B> loadable, Clas
       throw error.create(errorMsg);
     }, t -> (B)t);
   }
+
+  
+  /** Models */
+  public static final Loadable<BlockElement> BLOCK_ELEMENT = new GsonLoadable<>(new GsonBuilder().registerTypeAdapter(BlockElement.class, new BlockElement.Deserializer()).registerTypeAdapter(BlockElementFace.class, new BlockElementFace.Deserializer()).registerTypeAdapter(BlockFaceUV.class, new BlockFaceUV.Deserializer()).create(), BlockElement.class);
```
KnightMiner: `BlockElement` is a client side class. This is going to crash servers. Make a separate class to hold client side loadables if they are needed among many classes.

src/main/java/slimeknights/tconstruct/library/client/modifiers/block/BlockModifierModelMapManager.java
```diff
@@ -0,0 +1,236 @@
...
+
+
+  /** Gets a map of modifier models for the given tool */
+  public ModifierModelMap getModelsForTool(Function<Material, TextureAtlasSprite> spriteGetter, List<ResourceLocation> options) {
+    // quick exit: no options
+    if (options.isEmpty()) {
+      return ModifierModelMap.EMPTY;
+    }
+    // fetch options, filter to just those that exist
+    List<ModifierModelMap> maps = options.stream().map(id -> this.models.getOrDefault(id, ModifierModelMap.EMPTY)).filter(NOT_EMPTY_MAP).toList();
```
KnightMiner: This should use a new class that is designed for the 3D models instead of trying to reuse the standard modifier model class per the above discussion about making a unique interface for the block model needs.

KnightMiner:
The zip file linked in the main post seems to be corrupted, I cannot extract any of the JSON files in it due to an unknown error.

KnightMiner:
I feel like this needs a decent bit of work before it can be merged. You seem to have gotten stuck trying to reuse the existing modifier model system instead of creating something specific to the needs of the 3D tools. The individual models seem to be doing way too much work just to adhere to the interface meant for item layer models; this should not need to be this complex jus to write a model; it should be something very approachable by addons.

The base 3D modifier model should look a lot like the parameters to `SimpleBlockModel#bakePart`. Take a `SimpleBakedModel.Builder` parameter and just add the baked elements to it directly like `SimpleBlockModel` does. No need to go through all the extra work with transforms, just use the transforms from the parent where relevant. It also should pass in any other parameters needed to bake a `BlockElement`, such as the quad transformer and the resource location (you should not be passing in null and suppressing warnings, look at how we do it on `SimpleBlockModel`)

For handling textures, we can wrap `IGeometryBakingContext` in a custom class like `ModifierBakingContext`. It should have the `IToolStackView` as a field, along with the `ModifierEntry` so it can call a tool/entry sensitive `getMaterial(IToolStackView tool, ModifierEntry entry, String name)` method to fetch textures; this will prevent them from conflicting with the textures from other modifiers.

I'm not convinced we need multiple distinct modifier model managers. A single manager that just lazy loads the models from the other folder seems simpler to me. You could start by fetching a map of all JSONs using the helper from `SimpleJsonResourceReloadListener`, which you lazily parse into a map as you encounter models with the given name.

It may be worth doing a markdown writeup of the plan for this instead of jumping into code so we can ensure its reasonable before everything gets coded. We could discuss that further on Discord if you wish.

Qikahome:
[testResourcePack.zip](https://github.com/user-attachments/files/31693941/testResourcePack.zip)
new test pack
I've already refactored the code based on the discussion on Discord.

src/main/java/slimeknights/tconstruct/library/json/TinkerLoadables.java
Qikahome: separated to ClientLoadables

KnightMiner:
I will review this again when I get a chance, though I will say I am planning to make this next release without it as I suspect for such a major feature it will need more iteration and discussion.

Mostly mentioning this to ensure you don't feel like you need to rush to update this when I next leave comments. I expect after this release to have another release in a month or so with bug fixes.

PR #5701, opened 2026-08-02
Add Unarmed Logic to Monster Melee

### LINKED ISSUE #5695: Only players can use unarmed attack
### Minecraft Version

1.20.1

### Modloader

Forge

### Modloader Version

36.1.0

### Mantle Version

1.11.104

3.11.2.166

### Describe your issue

When I give a chestplate with high attack damage and fiery modifier to a mob, not only does the attack damage not increase, but also the target isn't on fire.

### Steps to reproduce

As the description.

### Crash Report

_No response_

Yes

### Performance Enchancers

None of the above

### Other mods

_No response_

I did not search

gjhihihi:
Fix #5695


src/main/java/slimeknights/tconstruct/tools/logic/ToolEvents.java
```diff
@@ -290,6 +290,18 @@ static void livingHurt(LivingHurtEvent event) {
           for (ModifierEntry entry : tool.getModifiers()) {
             originalDamage = entry.getHook(ModifierHooks.MONSTER_MELEE_DAMAGE).getMeleeDamage(tool, entry, meleeContext, baseDamage, originalDamage);
           }
+        } else if (weapon.isEmpty()) {// unarmed
+          // don't consider adding non-chest armor to the unarmed tag
+          ItemStack chestplate = living.getItemBySlot(EquipmentSlot.CHEST);
+          if (!chestplate.isEmpty() && chestplate.is(TinkerTags.Items.UNARMED)) {
+            IToolStackView tool = ToolStack.from(chestplate);
+            // already know the player is null
+            ToolAttackContext meleeContext = ToolAttackContext.attacker(living, null).target(entity).applyAttributes().build();
```
KnightMiner: This should be calling `toolAttributes()` instead of `applyAttributes()` to ensure we are using the chestplate stats, not the empty held item stats (notably the chestplate damage is not added as an entity attribute as that would boost non-unarmed).

This also needs to call `slot()` so the modifier knows the chestplate triggered the modifier.

Same comments at the other callsite below.

src/main/java/slimeknights/tconstruct/tools/logic/ToolEvents.java
```diff
@@ -290,6 +290,18 @@ static void livingHurt(LivingHurtEvent event) {
           for (ModifierEntry entry : tool.getModifiers()) {
             originalDamage = entry.getHook(ModifierHooks.MONSTER_MELEE_DAMAGE).getMeleeDamage(tool, entry, meleeContext, baseDamage, originalDamage);
           }
+        } else if (weapon.isEmpty()) {// unarmed
```
KnightMiner: This is minor, but I don't like redundant condition checks. Just split the `&&` above into an `if (!weapon.isEmpty) { if (weapon.is(...)) {`

src/main/java/slimeknights/tconstruct/tools/logic/ToolEvents.java
```diff
@@ -290,6 +290,18 @@ static void livingHurt(LivingHurtEvent event) {
...
+          // don't consider adding non-chest armor to the unarmed tag
+          ItemStack chestplate = living.getItemBySlot(EquipmentSlot.CHEST);
+          if (!chestplate.isEmpty() && chestplate.is(TinkerTags.Items.UNARMED)) {
+            IToolStackView tool = ToolStack.from(chestplate);
+            // already know the player is null
+            ToolAttackContext meleeContext = ToolAttackContext.attacker(living, null).target(entity).applyAttributes().build();
+            float baseDamage = originalDamage;
```
KnightMiner, line 307: This needs to sum in chestplate stats. You should be able to do `baseDamage = context.getBaseDamage()`.

However, that discards the original damage from the event. To account for that, I think in the builder it will be necessary to manually set the base damage to `originalDamage + toolDamage - entity.getAttribute(ATTACK_DAMAGE)` where `toolDaamge` comes from `toolAttributes()`. Find the cleanest way to do that logic (maybe you make a new attributes helper in the builder for damage override maybe you call `addDamage(originalDamage - entity.getAttribute(ATTACK_DAMAGE)` with a new `addDamage` method. (doing it in the builder ensures modifiers know the correct base damage)

KnightMiner: Thinking about this a bit more today, we can probably ignore main hand attributes on the chestplate as its super niche; we basically just use it for flint's trait which isn't an armor modifier. Similarly, seems unlikely a chestplate would add a knockback mainhand melee attribute.

So to simplify, could replace `applyAttributes()` with `setBaseDamage(originalDamage + tool.getStats().get(ATTACK_DAMAGE)`.

gjhihihi: Doesn't strength modifier change the main hand's attribute? It's a chestplate modifier.

gjhihihi: I can't get `originalDamage` at the other callsite below. Should I replace it with `amount` directly?

KnightMiner: Hmm, for that case lets just replace it with the entity damage attribute plus the chestplate damage stat. Thats pretty close to accurate; doesn't take into account the type of attack they did but its good enough for that hook given I doubt many modifiers are conditioning on "original damage" that late (plus that is the value we used in the above version).

src/main/java/slimeknights/tconstruct/library/tools/context/ToolAttackContext.java
```diff
@@ -262,6 +262,13 @@ public Builder toolAttributes(IToolStackView tool) {
       return this;
     }
 
+    /** Change the base damage for some special reason. */
+    public Builder addAttributes(float damage, float knockback) {
```
KnightMiner: based on how you are using this (and the fact its normally coming from tool stats), seems like `addBaseDamage` is simpler, leave out the knockback part.

src/main/java/slimeknights/tconstruct/tools/logic/ToolEvents.java
KnightMiner: This is still missing being set `baseDamage = context.getBaseDamage()`. We also probably need to add the above difference into `originalDamage` or we are still ignoring the chestplate melee damage.

So basially:
```java
originalDamage = context.getBaseDamage();
baseDamage = originalDamage;
```

PR #5705, opened 2026-08-14
Add some missing modifier textures for ranged tools and armor

CreepingCreeper:
Add some missing modifier textures for ranged tools and armor to Tinkers' Construct.
Now including crossbow, longbow, fishing rod and plate armor.
Most of these previously appeared in[ Fresh Tinker Grain](https://modrinth.com/project/AB7yH3nT).
Some textures are still missing, pending artwork from the artist.

屏幕截图_15-8-2026_215836_discord com


src/main/java/slimeknights/tconstruct/library/data/AbstractModifierModelMapProvider.java
```diff
@@ -167,6 +167,20 @@ public Builder modifier(ModifierId id, ModifierModel model, ModifierModel... mod
       return this;
     }
 
+    public Builder empty(ModifierId id) {
```
KnightMiner, line 174: New public or protected methods should get javadocs. Does not need to be anything too long, see others in this class for examples.

For empty, we really need to describe the usecase. This is to override a "lower" priority modifier map that sets a modifier, such as removing modifier textures on the hook for a cast fishing rod.

KnightMiner:
This PR adds `overforce`, `reinforced`, and `unbreakable` to plate armor's modifier sprite map, but does not include those textures. Did you forget to upload them?

I am also a little concerned about how much of the 16x16 space is used by a few of the armor modifiers even if it is transparent. Going to potentially cover up a lot of other modifiers, but I guess that is less of an issue with modifier invisibility in the worktable.

CreepingCreeper:
> This PR adds `overforce`, `reinforced`, and `unbreakable` to plate armor's modifier sprite map, but does not include those textures. Did you forget to upload them?
> 
> I am also a little concerned about how much of the 16x16 space is used by a few of the armor modifiers even if it is transparent. Going to potentially cover up a lot of other modifiers, but I guess that is less of an issue with modifier invisibility in the worktable.

Plate armor does not have these textures; I added them by mistake.

PR #5706, opened 2026-08-17
Fix soul glass being washed away by water

### LINKED ISSUE #5667: Soul glass gets washed away by water (unlike seared/scorched)
### Minecraft Version

1.20.1

### Modloader

Forge

### Modloader Version

47.4.20

### Mantle Version

1.11.104

3.11.2.166

### Describe your issue

When you place a block of soul glass in an infinite water source, it immediately breaks and drops itself.
However, when you place seared soul glass or scorched soul glass, they place perfectly fine.

### Steps to reproduce

1. Make a 2x2 infinite water source in the ground
2. Place a soul glass block in one of the corners
3. It drops as an item

### Crash Report

_No response_

Yes

### Performance Enchancers

None of the above

### Other mods

_No response_

Checked pinned issues, Searched open issues, Searched closed issues, Checked the FAQ

ryancircelli:
Placing soul glass where flowing water can reach it destroys the block and drops it as an item. Seared soul glass and scorched soul glass are unaffected.

soul_glass is missing forceSolidOn() in its block properties. It has noCollission() like its seared/scorched siblings, but without forceSolidOn() to compensate it reads as non-solid, so water treats the space as free to flow into. The seared and scorched variants build off structureNonSolid(), which already calls forceSolidOn(), so they're fine. This adds forceSolidOn() to soul_glass to match.

Tested in game, same rig on both builds, water flowing into the block rather than setblock (which would skip the bug entirely):

before: the glass pops as an item and water flows straight through
before

after: water just stops beside it
after

Also checked you still sink through it same as before — forceSolidOn() only changes whether the block reads as solid, not its collision shape. Soul glass panes were never affected by this bug in the first place since they already waterlog instead of getting destroyed.

Fixes #5667


KnightMiner:
For the future, I don't like letting Github autoclose issues when a PR is merged. Instead of saying "Fixes ####", say "Resolves ####" or simply say "(####)". I only want the issues closed once the fix is released; not simply on a branch.

Issue #5682, opened 2026-06-12
Spilling and Spitting have different action in creative gamemode


gjhihihi:
### Minecraft Version

1.20.1

### Modloader

Forge

### Modloader Version

36.1.0

### Mantle Version

1.11.97

### Tinkers' Construct Version

3.11.2.166

### Describe your issue

Spitting will consume fluid in creative, but spilling won't.

### Steps to reproduce

1. Get a swasher and filll some fluid.
2. Attack some mobs with it.
3. Spit with it.

### Can you reproduce with just Tinkers?

Yes

### Performance Enchancers

None of the above

### Searched for known issues?

I did not search

KnightMiner:
Fixed in 3.12.0.220.

PR #5707, opened 2026-08-17
Fix spitting draining the tank in creative mode

ryancircelli:
Spitting a tool with a tank (e.g. the Swasher) drains the tank in creative mode, even though spilling (melee) doesn't.

SpittingModule.spit() drains the tank unconditionally. It's missing the creative check that SpillingModule and the other tank-draining modules (SlurpingModule, SplashingModule, etc.) already have. The first commit guards the drain with the same player == null || !player.isCreative() check the other modules use. Durability was already creative-safe on its own, so that part is untouched.

The second commit changes no behavior: since that same check was written out by hand in six places, it now lives in one helper, ModifierUtil.consumesResources, the way ToolDamageUtil is already the one place deciding whether durability gets spent. Spilling, spitting, slurping, splashing and use fluid on hit all call it now, and ToolDamageUtil.directDamage does too; FireballModule keeps its own check, since that one skips durability whenever there is no player rather than only for creative players. Happy to drop that commit if you'd rather keep the fix minimal.

Tested in game with an Iron Swasher, tank filled through the game's own UI:

before: three creative spits, tank draining each time
before

after: same spits, tank stays put
after

Survival still drains normally, a mob holding a spitting tool still drains too (the guard is player == null || ..., not player != null && ...), and spilling was already unaffected either way. I re-ran the same checks after the refactor: creative spitting and creative spilling still leave the tank alone, both still drain in survival, and slurping (which also moved onto the helper) is unchanged in creative and still drains in survival.

(#5682)

### DIFF
--- a/src/main/java/slimeknights/tconstruct/library/tools/helper/ModifierUtil.java
+++ b/src/main/java/slimeknights/tconstruct/library/tools/helper/ModifierUtil.java
@@ -91,7 +91,17 @@ public final class ModifierUtil {
    * @return  True if resources should be consumed.
    */
   public static boolean consumesResources(@Nullable LivingEntity entity) {
-    return !(entity instanceof Player player && player.isCreative());
+    return !(entity instanceof Player player) || consumesResources(player);
+  }
+
+  /**
+   * Checks if the given player pays the resource costs of using a tool, such as draining the tank or consuming ammo.
+   * Creative players get their resources for free, matching {@link ToolDamageUtil#directDamage(IToolStackView, int, LivingEntity, ItemStack)} skipping durability for them.
+   * @param player  Player using the tool. Null consumes, as only creative players get their resources for free.
+   * @return  True if resources should be consumed.
+   */
+  public static boolean consumesResources(@Nullable Player player) {
+    return player == null || !player.isCreative();
   }
 
   /**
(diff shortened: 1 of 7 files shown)


src/main/java/slimeknights/tconstruct/library/tools/helper/ModifierUtil.java
```diff
@@ -84,6 +84,16 @@ public static Player asPlayer(@Nullable Entity entity) {
     return null;
   }
 
+  /**
+   * Checks if the given entity pays the resource costs of using a tool, such as draining the tank or consuming ammo.
+   * Creative players get their resources for free, matching {@link ToolDamageUtil#directDamage(IToolStackView, int, LivingEntity, ItemStack)} skipping durability for them.
+   * @param entity  Entity using the tool. Anything that is not a creative player consumes, including mobs and null.
+   * @return  True if resources should be consumed.
+   */
+  public static boolean consumesResources(@Nullable LivingEntity entity) {
```
KnightMiner: This should have an overload for `Player` to save an `instanceof` check if the entity is already player. The overload would just do a null check then.

ryancircelli: Added `consumesResources(Player)` as just the null check, and the `LivingEntity` overload now extracts the player and hands off to it so the rule is still written once. Five of the seven callers already hold a `Player` and pick the overload up through overload resolution with no edit; spitting and `directDamage` stay on the `LivingEntity` version since neither has anything narrower. Also reworded the first commit's message to drop the auto-close line, per your note on the soul glass PR.

Issue #5699, opened 2026-07-29
Holding a cleaving tool while killing a mob by other means will drop the additional loot as if you killed it with the cleaving tool.


Chubbachoo:
### Minecraft Version

1.20.1

### Modloader

Forge

### Modloader Version

47.4.10

### Mantle Version

1.11.104

### Tinkers' Construct Version

3.11.2.166

### Describe your issue

Blowing up mobs while holding a cleaving tool drops their additional loot. Found this out when I accidentally blew myself up.

<img width="111" height="86" alt="Image" src="https://github.com/user-attachments/assets/3b6ec078-701c-4ec8-850b-b3b80327e88b" />

### Steps to reproduce

Spawn a ton of mobs.
ignite tnt near them
hold a cleaver when the tnt blows up

### Can you reproduce with just Tinkers?

I have not tried

### Performance Enchancers

Rubidium/Embeddium

### Searched for known issues?

Searched open issues, Searched closed issues

KnightMiner:
That is just how vanilla loot tables work, its is not fixable.

We can fix it for our specific damage sources and do so with our bows and a few other tools notably. Any fix for things like explosions relies on a heuristic that may be wrong. I suppose we can at least setup a damage source blacklist or a "causing entity" whitelist to ensure only our projectiles count for severing.

Chubbachoo:
Ok.

KnightMiner:
Looking into this issue a little deeper, I think the fix is to patch ModifierLootModifier with a damage source whitelist. Can likely reuse the melee damage sources tag we used for one of our other features. Core idea is modifiers should only modify loot drops if the held tool was involved in the kill. Probably best to create a new tag for this even if it sources from an existing tag we have.

Currently undecided if this should include purely melee damage sources, or anything in melee range is fine (e.g. thorns). For now lets stick to purely melee as we can always tag more stuff later. May also be worth checking `DamageSource#isDirect`; its possible that is sufficient for most of our needs in which case we would just want a damage source tag blacklist for anything that is direct but shouldn't be (e.g. not sure where explosions land off the top of my head)

For projectiles, those are already handled via the projectile capability, so no need to special case their damage sources. Will also fix vanilla projectiles working severing.

KnightMiner:
Fixed in 3.12.0.220.

PR #5708, opened 2026-08-17
Fix indirect damage applying loot modifiers from the held tool

ryancircelli:
Killing a mob with something other than your held tool (TNT you lit, for example) still applies the tool's loot modifiers, like severing heads, as if the tool made the kill.

Went with the damage source tag approach from the issue comment. ModifierLootModifier only falls back to the killer entity's held item when the damage source is in a new tag, tconstruct:loot_modifier_whitelist. It pulls in #tconstruct:protection/melee and adds piercing and the melee halves of the fluid damage pairs, since those are damage the held tool dealt itself. The projectile branch is untouched.

Thorns and spiny are deliberately out of the tag, going with purely melee for now. Bleeding is out too, so a mob that dies to the lacerating bleed tick no longer drops a severing head. That's a behavior change rather than something I wanted to quietly tag around; adding tconstruct:bleeding later restores it.

Tested in game, cleaver pushed to guaranteed severing chance, six zombies ringed around a block of TNT:

before: TNT goes off, six heads on the ground
before

after: same explosion, no heads
after

after: the bleed-tick kill, no head either
bleed

Three melee kills in the same session still gave three heads, so severing on direct hits is unaffected.

(#5699)

### DIFF
--- a/src/main/java/slimeknights/tconstruct/common/data/tags/DamageTypeTagProvider.java
+++ b/src/main/java/slimeknights/tconstruct/common/data/tags/DamageTypeTagProvider.java
@@ -59,6 +59,7 @@ import static slimeknights.tconstruct.common.TinkerDamageTypes.WATER;
 import static slimeknights.tconstruct.common.TinkerTags.DamageTypes.BLAST_PROTECTION;
 import static slimeknights.tconstruct.common.TinkerTags.DamageTypes.FALL_PROTECTION;
 import static slimeknights.tconstruct.common.TinkerTags.DamageTypes.FIRE_PROTECTION;
+import static slimeknights.tconstruct.common.TinkerTags.DamageTypes.IS_MELEE;
 import static slimeknights.tconstruct.common.TinkerTags.DamageTypes.LOOT_MODIFIER_WHITELIST;
 import static slimeknights.tconstruct.common.TinkerTags.DamageTypes.MAGIC_PROTECTION;
 import static slimeknights.tconstruct.common.TinkerTags.DamageTypes.MELEE_PROTECTION;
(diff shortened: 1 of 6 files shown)


ryancircelli:
missed new comments on 5699, will reconsider approach

ryancircelli:
Reworked to the damage source tag approach from the issue comment.

ryancircelli:
One call I'd rather leave to you: ExplosionFluidEffect uses explosion_melee / mob_explosion_melee on its non-projectile path, so an explosive spill that finishes a mob on your own swing won't drop heads with the tag as written. The damage source entity there is still the attacker, so tagging those two would credit the right tool, and vanilla explosions like TNT stay out either way. Left them out for now since you weren't sure where explosions land — happy to add them if you want them in.

src/main/java/slimeknights/tconstruct/common/data/tags/DamageTypeTagProvider.java
```diff
@@ -87,6 +88,9 @@ protected void addTags(Provider pProvider) {
 
     // modifiers
     tag(MODIFIER_WHITELIST).add(MOB_ATTACK, MOB_ATTACK_NO_AGGRO);
+    // loot modifiers come from the held tool, so limit them to melee damage the tool is responsible for
+    // projectiles are not needed here, they use the modifiers stored on the projectile instead
+    tag(LOOT_MODIFIER_WHITELIST).addTag(MELEE_PROTECTION).add(PIERCING, FLUID_FIRE.melee(), FLUID_COLD.melee(), FLUID_MAGIC.melee(), WATER.melee());
```
KnightMiner: This should not include the melee protection tag. Its possible someone adds something that is thematically melee to that tag but is not melee from the perspective of this modifier.

You could make both this and melee protection share a common local tag though.

src/main/java/slimeknights/tconstruct/tools/modifiers/ModifierLootModifier.java
```diff
@@ -66,7 +67,9 @@ protected ObjectArrayList<ItemStack> doApply(ObjectArrayList<ItemStack> generate
       }
 
       // not a projectile causing it, fetch the killer entity directly from loot context
-      if (context.getParamOrNull(LootContextParams.KILLER_ENTITY) instanceof LivingEntity living) {
+      // requires a melee damage source, the held tool is not responsible for kills it did not make such as explosions
+      DamageSource damageSource = context.getParamOrNull(LootContextParams.DAMAGE_SOURCE);
+      if (damageSource != null && damageSource.is(TinkerTags.DamageTypes.LOOT_MODIFIER_WHITELIST) && context.getParamOrNull(LootContextParams.KILLER_ENTITY) instanceof LivingEntity living) {
```
KnightMiner: It occurs to me this hook is also run for harvest tools breaking blocks; its not strictly a melee tool. Based on that, may be best to just patch this directly in the severing logic rather than a global change; minimizes the risk of breaking an addon.

KnightMiner:
> One call I'd rather leave to you: ExplosionFluidEffect uses explosion_melee / mob_explosion_melee on its non-projectile path, so an explosive spill that finishes a mob on your own swing won't drop heads with the tag as written. The damage source entity there is still the attacker, so tagging those two would credit the right tool, and vanilla explosions like TNT stay out either way. Left them out for now since you weren't sure where explosions land — happy to add them if you want them in.

Explosions caused by your weapon should inherit its loot behavior I think. I believe as setup explosions caused by projectiles will get the same behavior as we check the projectile itself for the modifier.

src/main/java/slimeknights/tconstruct/common/data/tags/DamageTypeTagProvider.java
ryancircelli: Done as suggested — both tags now source a new `tconstruct:is_melee` holding just the damage types that are an actual melee attack, and `protection/melee` is that tag plus cramming and the optional Twilight Forest sources, so protection keeps exactly the members it had. The practical change is that cramming and the Twilight Forest sources are no longer in the loot whitelist, which is the narrowing your comment was after. I also added `explosion_melee` and `mob_explosion_melee` to the loot whitelist per your answer on the other thread about explosions from your weapon inheriting its loot behavior.

KnightMiner: mobs can technically speaking use looting, you should add the TF mobs to it

ryancircelli: Added — the nine moved onto `is_melee` rather than duplicated onto the whitelist, since every one is an attack a mob makes and that's what the shared tag is for. `protection/melee` still resolves to the same 16 members, and cramming stays protection-only as it's still not an attack.

Issue #4881, opened 2022-05-26
Fuel displayed in the Smeltery GUI is wrong when one Fuel container runs out of fuel


Insane96:
### Minecraft Version

1.18.2

### Forge Version

40.1.20

### Mantle Version

1.9.20

### Tinkers' Construct Version

3.5.0.17

### Describe your issue

Having multiple fuel container (with the same fuel) confuses the smeltery GUI, showing the wrong amount of fuel left when one fuel container runs out of fuel.
In the video one scorched tank run out of lava, at first the Smeltery showed "No fuel", then it showed double the fuel that was actually in the smeltery. Re-joining the world / reloading chunks fixes this.

https://www.youtube.com/watch?v=Md4MGPfKjcM

### Other mods

None

### Tried reproducing with just Tinkers?

Yes

### Performance Enchancers

None of the above

### Searched for known issues?

Checked pinned issues, Searched open issues, Searched closed issues, Checked the FAQ, I did not search

KnightMiner:
Don't have multiple fuels in the smeltery.

KnightMiner:
Commit pushed fixes the bug with fluid info doubling and a couple other related issues. Still need to determine a good fix for "no fuel found" showing briefly, likely a sync issue with the current fuel.

KnightMiner:
Fixed in 3.12.220 for 1.20.1. No immediate plans to backport.

PR #5709, opened 2026-08-19
Fix smeltery fuel gauge ignoring other fuel tanks when the current one empties

ryancircelli:
This is the "no fuel found" half of #4881, the part left open after the fluid doubling fix. The guess on the issue was a sync issue with the current fuel; it is not, server and client agree on the current tank the whole time. It is display logic in `MultitankFuelModule#getFuelInfo()`.

That method resolves a main tank, asks the superclass for its contents, then walks the other tanks to sum them into the gauge. The sum is behind one check, `MultitankFuelModule.java:262`:

```java
    FuelInfo info = super.getFuelInfo();
    // add extra fluid display
    if (!info.isEmpty()) {
```

`super.getFuelInfo()` only reads the main tank, and `FuelInfo.of` short circuits to the shared empty instance the moment that tank reads empty (`FuelModule.java:266-268`). So an empty main tank skips the aggregation entirely and every other tank is dropped, however full. `GuiFuelModule` then draws nothing and the tooltip falls to `gui.tconstruct.melting.fuel.empty`.

The main tank reads empty while the smeltery is still burning because fuel is banked. `findFuel` drains a whole charge at once and stores a tick countdown, and `SmelteryBlockEntity.heat()` only looks for fuel again when that countdown hits zero (`SmelteryBlockEntity.java:72-73`). For the whole burn-down of the last charge, `lastPos` still points at the tank that is now empty, correctly, and the gauge concludes there is no fuel. When the charge runs out `findFuel` moves on and the gauge fixes itself, which is why it shows briefly, and a rejoin clears it because that resets `lastPos`.

The fix promotes the first tank that still holds fluid when the main tank reads empty, then aggregates as before. Nothing is written back, `fluidHandler` and `lastPos` are untouched and the promotion is recomputed per frame, so there is no protocol or server side change.

One thing worth your call: the scan promotes the first tank with fluid, not the first tank with valid fuel, matching the TODO already in that method about a valid tank being more effort to find. The promoted tank still goes through one `findRecipe` for the temperature, so a non-fuel fluid gets temperature 0 and the tooltip already renders that as "Invalid melting fuel!". The visible cost is a structure holding some non-fuel fluid in a tank ahead of the real fuel tank, where the gauge would point at the wrong one. Happy to make it require valid fuel if you would rather.

Tested on a tiny smeltery with two seared fuel tanks, one holding a single 50 mB lava charge and the other four buckets, melting raw iron so the charge is actually consumed. On this size the banked charge is 400 ticks, so the wrong display lasts about twenty seconds.

before, the last charge has drained the first tank and the gauge is empty while the other tank holds four buckets:
before

after, same moment, the gauge shows the tank that still has fuel:
after

with every tank empty it still reads no fuel found, and a main tank that does have fluid renders exactly as before:
empty

`./gradlew build test` passes, 187 tests.

Report: #4881

### DIFF
--- a/src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MultitankFuelModule.java
+++ b/src/main/java/slimeknights/tconstruct/smeltery/block/entity/module/MultitankFuelModule.java
@@ -14,7 +14,6 @@ import net.minecraftforge.fluids.capability.IFluidHandler;
 import net.minecraftforge.fluids.capability.templates.EmptyFluidHandler;
 import slimeknights.mantle.block.entity.MantleBlockEntity;
 import slimeknights.mantle.util.WeakConsumerWrapper;
-import slimeknights.tconstruct.library.recipe.fuel.MeltingFuel;
 import slimeknights.tconstruct.library.utils.Util;
 
 import javax.annotation.Nonnull;
@@ -22,7 +21,6 @@ import javax.annotation.Nullable;
 import java.util.LinkedHashMap;
 import java.util.List;
 import java.util.Map;
-import java.util.Map.Entry;
 import java.util.function.Supplier;
 
 /** Fuel module that supports multiple tanks, selecting just one for the fuel result */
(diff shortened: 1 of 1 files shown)


KnightMiner:
I think this code is going to be too inefficient in a large smeltery given the fact it runs every frame.

I think optimization that we should consider is moving the tank search to before the supercall so it can properly be cached. There are two cases we should do the search:

1. If the main tank is null
2. If the main tank contains an empty fluid (note this is not an empty lazy optional)

If the main tank is null, instead of selecting the first tank, we should probably select the first tank that contains fluid. During that search, cache the fluid handler.

If the main tank contains an empty fluid before doing the super call, we can search other fluid handlers to find a new tank with fluid.

In either case, if we find no tanks with fluid, we could cache an empty lazy optional, but that may cause inaccurate display if you add fuel when there previously was none. If we don't cache, we are back to the same problem of it running the search every frame but it at least only happens when all tanks are empty (which is probably rare with a large set of tanks?).

Also need to consider whether the search should be "find a fluid" or "find a fuel". Former is way more efficient, and tbh, there is not much usecase for putting non-fuels in the smeltery walls so perhaps we skip the lookup as an optimization.

----

I guess we do loop over the fluid handlers below where your loop is located so its already doing a bit of work every frame. So I suppose the recipe lookup is really the biggest concern.

So I propose skip the recipe lookup, cache the "new main tank", but if all tanks contain no fuel then don't bother caching. Means we have a loop over all tanks in either case, same efficiency.

Aside, for the local varible problem, I'd say just ditch the `isPresent` call in favor of an `if` statement.

ryancircelli:
Done as proposed — the search runs before the super call, caches into the existing `fluidHandler` field (same `clearLastListener`/`addListener` shape `tryFuelPosition` uses), re-searches when the cached tank reads an empty fluid, and caches nothing when every tank is empty. One correction: the first commit already searched by fluid rather than fuel, but it did call `findRecipe` once for a temperature — that's gone, the super call now does the single lookup it always did. The `ifPresent` is a plain `if`.

Re-verified in game: the banked-burn window shows the other tank, all-empty still reads "No fuel found", and filling a tank with the GUI open brings the gauge back the next frame — the case that would have broken had the empty result been cached.

PR #5714, opened 2026-08-28
Use newer JEI convenience functions

mezz:
No logic changes here, just updating to use some newer JEI methods.
All these methods are available within the current minimum-jei-version for TC.

Most of these are to avoid having to specify verbose basic things like `VanillaTypes.ITEM_STACK` or `RecipeIngredientRole.INPUT`

### DIFF
--- a/src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
+++ b/src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
@@ -5,6 +5,7 @@ import com.mojang.blaze3d.vertex.PoseStack;
 import lombok.Getter;
 import mezz.jei.api.gui.builder.IRecipeLayoutBuilder;
 import mezz.jei.api.gui.builder.IRecipeSlotBuilder;
+import mezz.jei.api.gui.builder.ITooltipBuilder;
 import mezz.jei.api.gui.drawable.IDrawable;
 import mezz.jei.api.gui.ingredient.IRecipeSlotsView;
 import mezz.jei.api.gui.widgets.IRecipeExtrasBuilder;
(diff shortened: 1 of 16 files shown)


src/main/java/slimeknights/tconstruct/plugin/jei/ToolBuildingCategory.java
```diff
@@ -147,10 +148,10 @@ public void draw(ToolBuildingRecipe recipe, IRecipeSlotsView recipeSlotsView, Gu
   }
 
   @Override
-  public List<Component> getTooltipStrings(ToolBuildingRecipe recipe, IRecipeSlotsView recipeSlotsView, double mouseX, double mouseY) {
-    return recipe.requiresAnvil() && GuiUtil.isHovered((int) mouseX, (int) mouseY, 76, 44, ITEM_SIZE, ITEM_SIZE) ?
-      List.of(TConstruct.makeTranslation("jei", "tinkering.tool_building.anvil")) :
-      List.of();
+  public void getTooltip(ITooltipBuilder tooltip, ToolBuildingRecipe recipe, IRecipeSlotsView recipeSlotsView, double mouseX, double mouseY) {
+    if (recipe.requiresAnvil() && GuiUtil.isHovered((int)mouseX, (int)mouseY, 76, 44, ITEM_SIZE, ITEM_SIZE)) {
+      tooltip.add(TConstruct.makeTranslation("jei", "tinkering.tool_building.anvil"));
```
KnightMiner, line 153: Huh, wonder why this isn't a constant. Must have had it separate for hot swapping at some point.

I can fix it after merge since its the only issue I see (and its not caused by your PR)

KnightMiner:
This PR is also reminding me of a few of the weird hacks I had to use to get around an API bug/misunderstanding. Will have to check those some time to see how many are still relevant.

Is the recipe focuses using the ingredient subtype interpreter context a bug worth reporting to JEI so it gets addressed? Or is that considered intentional behavior?

mezz:
Thanks! I was doing the update very mechanically as a pure refactor without logic changes, and wasn't paying attention to how you've implemented things. I'll take a closer look at the issue you pointed out, now that this is done.

mezz:
> Is the recipe focuses using the ingredient subtype interpreter context a bug worth reporting to JEI so it gets addressed? Or is that considered intentional behavior?

This seems to be something I could improve in JEI. It should find an exact match if possible, and fall back on matching the `UidContext.RECIPE`

mezz:
Implemented here: https://github.com/mezz/JustEnoughItems/commit/a5059b22f6f7b284c1a1ea57f4900512f86c22fe

PR #5716, opened 2026-08-29
Use RecipeIngredientRole.CATALYST for catalysts

mezz:
Very minor, this marks ingredients that aren't consumed as "catalysts" which can be useful for some mods that use JEI to calculate recipe trees and things like that.

### DIFF
--- a/src/main/java/slimeknights/tconstruct/plugin/jei/AlloyRecipeCategory.java
+++ b/src/main/java/slimeknights/tconstruct/plugin/jei/AlloyRecipeCategory.java
@@ -103,7 +103,7 @@ public class AlloyRecipeCategory implements IRecipeCategory<AlloyRecipe> {
   /**
    * Draws a variable number of fluids
    * @param builder      Builder
-   * @param role         Role of the set of fluids in the recipe
+   * @param role         Role of the fluids in the recipe
    * @param x            X start
    * @param y            Y start
    * @param totalWidth   Total width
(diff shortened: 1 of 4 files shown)


src/main/java/slimeknights/tconstruct/plugin/jei/AlloyRecipeCategory.java
```diff
@@ -115,7 +115,7 @@ public void draw(AlloyRecipe recipe, IRecipeSlotsView slots, GuiGraphics graphic
    * @param <T> Object type
    * @return Max amount based on fluids
    */
-  public static <T> int drawVariableFluids(IRecipeLayoutBuilder builder, RecipeIngredientRole role, int x, int y, int totalWidth, int height, List<T> fluids, int minAmount, Function<T,List<FluidStack>> mapper, Function<T,IRecipeSlotTooltipCallback> tooltip) {
+  public static <T> int drawVariableFluids(IRecipeLayoutBuilder builder, Function<T,RecipeIngredientRole> role, int x, int y, int totalWidth, int height, List<T> fluids, int minAmount, Function<T,List<FluidStack>> mapper, Function<T,IRecipeSlotTooltipCallback> tooltip) {
```
KnightMiner, line 137: With how late we are in 1.20, its possible that tinkers addons are calling non-API methods, so if changing a signature I always like to add a deprecated overload with the old signature.

Exception is when its not really feasible for an addon to call it, but I think as a public static method there is a chance this is called.

KnightMiner:
One weird case that may not be possible to handle: in part builder recipes, whether the pattern is consumed is just determined by a tag. This means a single recipe can have some of the patterns consumed and some of them not. I assume that JEI cannot easily have a single slot animate between catalyst and input.

It might be worth checking if all the inputs match the tag to mark it as a catalyst, though we honestly don't have any recipes that work that way at this time (everything is either all consumed or a mix).

mezz:
Thanks for the review, I addressed both points. 
For the part builder, I just marked it as catalyst only if they are all catalysts, otherwise treat it as input. I think that's good enough for this situation.

PR #5717, opened 2026-08-29
Use createRecipeExtras to simplify drawing recipes

mezz:
This one has logic changes, it moves some drawing to the `createRecipeExtras` system.

Doing this removes the need for some complicated caching (`LoadingCache<Integer,IDrawableAnimated> cachedArrows;`) and simplifies cases where strings need to be drawn centered somewhere.
The widgets system also lets us set an area for drawing and tooltips together, which removes a lot of redundant position checks and makes it harder to mess up the hover areas.

One thing that came up here, the small `+` icon used for the `AbstractMeltingCategory` had a 16x16 hover area (too big to the right and down). I preserved that logic here in case it's intentional, otherwise it's kind of hard to hover over.


KnightMiner:
> One thing that came up here, the small + icon used for the AbstractMeltingCategory had a 16x16 hover area (too big to the right and down). I preserved that logic here in case it's intentional, otherwise it's kind of hard to hover over.

Must have been an oversight, but it is a fair point that 6x6 is a pretty small. Does this PR center the tooltip or is it still biased towards the lower left?

mezz:
I didn't change it, so it's still biased towards the lower right. I'll center it, that should be better

KnightMiner:
Looking at it in game, it seems the current area overlaps the tank area a bit, and goes out of bounds at the bottom. So the best solution is probably making the icon a little bigger and shifting its position over. Perhaps make it an 8x8 icon with a 12x12 area around it for the tooltip, think that will be big enough?

Feel free to just relocate and adjust the bounds for the icon or update the texture. If updating the texture, there should be about 2 pixels of clearance above the current icon to expand into.

image

Edit: I just pushed the last of the stuff I was actively working on so that should avoid any more merge conflicts for today at least.

mezz:
I increased the asset size to 8x8 and gave it 2px padding on all sides, then moved it over a bit. I think this is better now.
also resolved merge conflicts, it was minor

src/main/java/slimeknights/tconstruct/plugin/jei/modifiers/ModifierRecipeCategory.java
```diff
@@ -70,80 +84,52 @@ public ModifierRecipeCategory(IGuiHelper helper) {
...
-    SlotCount slots = recipe.getSlots();
-    if (slots == null && GuiUtil.isHovered(checkX, checkY, 102, 58, 24, 16)) {
-      tooltip.addAll(SlotIngredientRenderer.INPUT.getTooltip(null, TooltipFlag.NORMAL));
+    if (recipe.getSlots() == null) {
+      List<Component> slotlessTooltip = SlotIngredientRenderer.INPUT.getTooltip(null, TooltipFlag.NORMAL);
+      builder.addWidget(new RecipeTooltipWidget(SLOTLESS, 102, 58, slotlessTooltip));
+    }
+    Component levelText = getLevelText(recipe);
+    if (levelText != null) {
+      builder.addText(levelText, 83, 9)
```
KnightMiner: I feel like this has slightly less area than it used to, noticing the level range doesn't get to use the full space when it fit before. Can always adjust the translation to make it shorter but would be nice to have a bit more space.

Image

KnightMiner: I do like the way it looks for some of the text that was oversized though, like 

Image

Level range is just such a common one it should not be truncated.

src/main/java/slimeknights/tconstruct/plugin/jei/util/RecipeTooltipWidget.java
```diff
@@ -0,0 +1,60 @@
...
+import mezz.jei.api.gui.widgets.IRecipeWidget;
+import net.minecraft.client.gui.GuiGraphics;
+import net.minecraft.client.gui.navigation.ScreenPosition;
+import net.minecraft.network.chat.Component;
+
+import javax.annotation.Nullable;
+import java.util.List;
+
+/** Adds a tooltip to a rectangular area of a recipe, optionally drawing content in the same location. */
+public class RecipeTooltipWidget implements IRecipeWidget {
```
KnightMiner, line 14: Not a change required for this PR, but I feel this might be a useful utility to have in JEI directly as icons with tooltips comes up in a lot of recipes.

src/main/java/slimeknights/tconstruct/plugin/jei/MoldingRecipeCategory.java
```diff
@@ -51,16 +52,6 @@ public void draw(MoldingRecipe recipe, IRecipeSlotsView slots, GuiGraphics graph
     // if no mold, we "pickup" the item, so draw no table
     if (!recipe.getPattern().isEmpty()) {
       block.draw(graphics, 51, 40);
```
KnightMiner, line 54: curious, any reason why some drawables are left in this method while others are moved to create extras?

I assume anything unconditional makes sense here, but the block for the pattern is conditional.

src/main/java/slimeknights/tconstruct/plugin/jei/modifiers/ModifierRecipeCategory.java
KnightMiner: image

Suppose that is pretty close to the edges, so perhaps either shorter translation or more space is the solution here. Just changing the text to `Levels: ` would be sufficient I think.

src/main/java/slimeknights/tconstruct/plugin/jei/util/RecipeTooltipWidget.java
mezz: Agreed yeah. I am using this as an exercise to get ideas for how to improve the widget system. I'm adding things to 26.2 while I work on this, and then I'll backport.

src/main/java/slimeknights/tconstruct/plugin/jei/MoldingRecipeCategory.java
mezz: I thought about it like "background" vs "foreground" elements but I should probably reconsider how I do this. 
It may be better to move everything to extras, I'll try to offer clearer support in JEI.

src/main/java/slimeknights/tconstruct/plugin/jei/modifiers/ModifierRecipeCategory.java
mezz: Ah, it looks like the old text was technically overflowing outside of the recipe bounds.
I gave it 2 more pixels of width to the left and now it fits fine.

KnightMiner: image

opted to shrink the text anyways as that being right against the edge felt a little tight. But may be useful for one of the non-level labels like embellishments

src/main/java/slimeknights/tconstruct/plugin/jei/MoldingRecipeCategory.java
KnightMiner: I think anything that is constantly drawn feels cleaner to me in just draw. Extras does have the advantage of the builder syntax for common elements like arrows or shapeless icons though.

I plan to do a bit more cleanup with some of the elements come 1.21, just figured I should minimize resource pack breaking for now with the backgrounds by keeping the drawn elements at the same location.

PR #5718, opened 2026-08-30
Update to use new JEI widget and tooltip methods

mezz:
I added some new widget and tooltip methods to support the cases in Tinker's better.
The resulting recipe display is the same as before, just with a bit less code and complexity.


src/main/java/slimeknights/tconstruct/plugin/jei/AlloyRecipeCategory.java
```diff
@@ -112,7 +115,7 @@ public static <T> int drawVariableFluids(IRecipeLayoutBuilder builder, RecipeIng
    * @param <T> Object type
    * @return Max amount based on fluids
    */
-  public static <T> int drawVariableFluids(IRecipeLayoutBuilder builder, Function<T,RecipeIngredientRole> role, int x, int y, int totalWidth, int height, List<T> fluids, int minAmount, Function<T,List<FluidStack>> mapper, Function<T,IRecipeSlotTooltipCallback> tooltip) {
```
mezz, line 115: We just added this method in the last PR, so I felt that it's ok to remove it instead of deprecating.
The older version of this method is still kept for compatibility.

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
```diff
@@ -56,14 +52,14 @@ protected AbstractCastingCategory(IGuiHelper guiHelper, RecipeType<IDisplayableC
 
   @Override
   public void createRecipeExtras(IRecipeExtrasBuilder builder, IDisplayableCastingRecipe recipe, IFocusGroup focuses) {
+    builder.addDrawableWidget(block).setPosition(38, 35);
     int coolingTime = recipe.getCoolingTime();
-    builder.addDrawable(guiHelper.drawableBuilder(BACKGROUND_LOC, 117, 32, 24, 17)
-                                 .buildAnimated(Math.max(1, coolingTime), StartDirection.LEFT, false), 58, 18);
+    builder.addAnimatedRecipeArrowWidget(Math.max(1, coolingTime)).setPosition(58, 18);
```
KnightMiner: The casting arrow was invented (black instead of white) as white arrows usually indicate heating. The black arrow was a stylistic choice as it was cooling instead.

src/main/java/slimeknights/tconstruct/plugin/jei/entity/EntityMeltingRecipeCategory.java
```diff
@@ -83,14 +83,14 @@ public void setRecipe(IRecipeLayoutBuilder builder, EntityMeltingRecipe recipe,
     // output
     builder.addOutputSlot(115, 11)
            .setFluidRenderer(FluidValues.INGOT * 2, false, 16, 32)
-           .addTooltipCallback(new FluidTooltip(recipe.getDamage())) // object is cheap, no need to cache
+           .addRichTooltipCallback(new FluidTooltip(recipe.getDamage())) // object is cheap, no need to cache
```
KnightMiner: Minor, but this comment is probably not relevant anymore as JEI is caching those objects for us now internally.

Could be updated to something of the form of not worth deduplicating multiple recipes with the same damage, but honestly, not sure thats necessary to point out.

src/main/java/slimeknights/tconstruct/plugin/jei/melting/AbstractMeltingCategory.java
```diff
@@ -62,12 +58,11 @@ public AbstractMeltingCategory(IGuiHelper helper, RecipeType<MeltingRecipe> reci
 
   @Override
   public void createRecipeExtras(IRecipeExtrasBuilder builder, MeltingRecipe recipe, IFocusGroup focuses) {
-    IDrawable arrow = guiHelper.drawableBuilder(BACKGROUND_LOC, 150, 41, 24, 17)
-                                 .buildAnimated(recipe.getTime() * 5, StartDirection.LEFT, false);
-    builder.addWidget(new RecipeTooltipWidget(
-      arrow, 56, 18, Component.translatable(KEY_COOLING_TIME, recipe.getTime() / 4)));
+    builder.addAnimatedRecipeArrowWidget(recipe.getTime() * 5)
```
KnightMiner, line 62: Just checking, this draws the background arrow too, correct? Not a problem if it does but may be worth a comment to remind me for whenever I fully move away from drawing the background (given melting is essentially 2 standard slots and 3 fluid tanks, I could just put all three shapes on one texture)

src/main/java/slimeknights/tconstruct/plugin/jei/modifiers/ModifierRecipeCategory.java
```diff
@@ -111,18 +111,19 @@ private static Component getLevelText(IDisplayModifierRecipe recipe) {
...
     if (recipe.isIncremental()) {
-      builder.addWidget(new RecipeTooltipWidget(incremental, 83, 59, TEXT_INCREMENTAL));
+      builder.addDrawableWidget(incremental).setPosition(83, 59).setTooltip(TEXT_INCREMENTAL);
     }
     if (recipe.getSlots() == null) {
-      List<Component> slotlessTooltip = SlotIngredientRenderer.INPUT.getTooltip(null, TooltipFlag.NORMAL);
-      builder.addWidget(new RecipeTooltipWidget(SLOTLESS, 102, 58, slotlessTooltip));
+      builder.addDrawableWidget(SLOTLESS)
+        .setPosition(102, 58)
+        .setTooltip(tooltip -> SlotIngredientRenderer.INPUT.getTooltip(tooltip, null, Minecraft.getInstance().player, TooltipFlag.NORMAL));
```
KnightMiner: This one does seem a little odd as my understanding is it does not cache the value, instead fetching it each time its hovered, though I guess the difference is pretty small.

I do find it slightly concerning to call that method with a null ingredient as its not annotated nullable on the JEI docs, though I suppose we at least know its safe in this method.

Honestly, could just simplify the whole thing by making `SlotlessIngredientRenderer#TEXT_FREE` be public or have a public getter; not sure we have any other cases that need a null in that ingredient tooltip renderer.

src/main/java/slimeknights/tconstruct/plugin/jei/util/FluidTooltipCallback.java
```diff
@@ -42,6 +47,30 @@ default void onTooltip(IRecipeSlotView recipeSlotView, List<Component> tooltip)
...
+    while (listIterator.hasNext()) {
+      FormattedText line = listIterator.next().left().orElse(null);
+      if (line instanceof Component component && component.getContents() instanceof TranslatableContents translatable && AMOUNT_KEY.equals(translatable.getKey())) {
+        listIterator.remove();
+        FluidStack fluid = recipeSlotView.getDisplayedIngredient(ForgeTypes.FLUID_STACK).orElse(FluidStack.EMPTY);
+        List<Component> newTooltip = new ArrayList<>();
+        onFluidTooltip(fluid, recipeSlotView, newTooltip);
+        for (Component newLine : newTooltip) {
+          listIterator.add(Either.left(newLine));
+        }
```
KnightMiner: Lets move these lines to a new helper `default void onFluidTooltip(FluidStack, IRecipeSlotView, ITooltipBuilder)`. That will allow us to implement the new approach in parallel to the old one if we wish, which may make migration a bit easier.

src/main/java/slimeknights/tconstruct/plugin/jei/AlloyRecipeCategory.java
```diff
@@ -112,7 +115,7 @@ public static <T> int drawVariableFluids(IRecipeLayoutBuilder builder, RecipeIng
...
+  public static <T> int drawVariableFluidsWithRichTooltip(IRecipeLayoutBuilder builder, Function<T,RecipeIngredientRole> role, int x, int y, int totalWidth, int height, List<T> fluids, int minAmount, Function<T,List<FluidStack>> mapper, Function<T,IRecipeSlotRichTooltipCallback> tooltip) {
```
KnightMiner: I would like to have a name that won't need to be renamed again come the API cleanup for 1.21. Maybe `drawMultipleFluids`?

Perhaps we just relocate this method to a new class where it is more likely to be expected instead of under a specific category (its just here as alloys was the original that used it). `FluidTooltipCallback` seems a little too different since this is about `setRecipe`, so maybe a new class under `jei.util`? Could make it something generic like `CategoryUtil`?

src/main/java/slimeknights/tconstruct/plugin/jei/casting/AbstractCastingCategory.java
mezz: ah sorry that was careless of me, fixed

src/main/java/slimeknights/tconstruct/plugin/jei/entity/EntityMeltingRecipeCategory.java
mezz: makes sense, removed it

src/main/java/slimeknights/tconstruct/plugin/jei/melting/AbstractMeltingCategory.java
mezz: This does draw the background too. I added a comment because it might be easy to forget later

src/main/java/slimeknights/tconstruct/plugin/jei/modifiers/ModifierRecipeCategory.java
mezz: I did make sure the `SlotIngredientRenderer #getTooltip` override accepts `@Nullable SlotCount slots`, so that's not a problem,
but exposing `SlotIngredientRenderer#TEXT_FREE` makes this simpler and make more sense imo

src/main/java/slimeknights/tconstruct/plugin/jei/AlloyRecipeCategory.java
mezz: `CategoryUtil` seems fine, it can collect some similar helpers later.
I'll call it `drawMultipleFluids`

src/main/java/slimeknights/tconstruct/plugin/jei/util/FluidTooltipCallback.java
mezz: makes sense, done

Issue #2579, opened 2016-11-20
Seared glass and window not valid for Furnace


xenoflot:
Hi,

According to the manual, the Seared Furnace "can be made using seared blocks of any kind". I've found however that the Seared Glass and Seared Window blocks are not valid.

Cheers

tconstruct-1.10.2-2.5.6b
mantle-1.10.2-1.1.1
forge 2122

KnightMiner:
Seared block refers to all variants of seared stone, bricks, etc. Seared glass/windows does not work intentionally, which is why we specifically say they work on the other multiblocks.

If you can suggest a less ambiguous way to word it without listing every variant, we might be able to prevent the confusion.

xenoflot:
I'd suggest "can be made using opaque seared blocks". Opaque is a reasonably common word and even more common amongst Minecraft players.

PR #5719, opened 2026-09-02
(1.12) (book) (translation) Clarify that seared furnace uses only opaque blocks

Tokarak:
Fixes: #2579 (for version 1.12 only)

All translations now clarify that only "opaque" seared blocks may be used, as suggested in #2579. 

Background info: I was very frustrated that my furnace wasn't working in SkyFactory 4. The game does nothing to suggest that seared glass cannot be used, and seared glass is discoverable through JEI. Googling for common issues only shows things which are addressed in the book, or stupid ideas ("the tank must be in the corner"; "try building in another chunk").
Notes:

- Note that the ko_kr translation also had a grammatical error fixed. The previous version is in English (Google Translate):
  > You can craft scorched bricks and a scorched furnace controller at a scorched furnace.
- The Russian had a letter lowered to lowercase, for consistency with the rest of the book when "seared blocks" are translated.

Disclaimer: All translations from English were done using an LLM (with review from a second LLM, and verified with google translate for the Korean), which I hope is fine for a small patch.


KnightMiner:
We don't release new updates for 1.12 at this time. If that feature gets ported to 1.20, the book will be clarified.

You might have noticed the linked issue was closed nearly 10 years ago.

PR #5720, opened 2026-09-05
Add banner textures for traveler's shields

onethirdpirsquaredh:
Adds banner textures for traveler's shields. Doesn't add it to the `tconstruct:modifiable/banner` tag or add it to the traveler's shield modifier map, as I'm pretty sure those would require datagen or something and I don't know how to do that.

Examples:

image
image

A lot of detail is definitely lost from the original banner, but I think that's to be expected since the canvas is 16x16 and a circle.
Banners in the first image (besides the ominous banner) were taken from https://minecraft.tools/en/banner-galery.php?order=top&page=1


KnightMiner:
Those look pretty solid from the examples. I'll take a closer look at this when I get a chance later.

If you are interested, Tinkers Things could also use a set for the laminar shield in that mod. That would be a PR to that repo.

